PR #9, opened 2019-01-02
MsGraphicsPkg: Fix incorrect parameters in DEBUG macro.

tsunghowu:
When compiling the code with GCC 7, found couple of coding errors on using DEBUG macro.


spbrogan (approved):
Thanks for these changes.  We have noticed this coding practice doesn't work across toolchains and have been slowly addressing this.  We appreciate the help and we will soon be enabling CI with other toolchains to catch these types of issues.

PR #12, opened 2019-01-04
MsGraphicsPkg: FIX: Compilation warning with GCC regarding missing braces and some build erros.

tsunghowu:
This fixes a warning from GCC that complains about missing braces (caused by
multi-dimensional array is treated as a linear arrary and unicode marco). And fix
errors about initialized an 'extern' variable and build path errors in UNIX.

Test: With very little modifications in OvmfPkgX64.dsc, Drivers in MsGraphicsPkg can
be built successfully with OvmfPkg. In Qemu, SetupBrowserDxe could use MsGraphicsPkg/DisplayEngineDxe (with GopOverrideDxe, MsUiThemeProtocolDxe, SimpleWindowManger, OnScreenKeyboardDxe) to render the UI correctly.


tsunghowu:
Build failed with VS compiler. Drop this.

PR #14, opened 2019-02-01
MsGraphicsPkg/PcBdsPkg: Fix build errors with GCC.

tsunghowu:
Most of errors are about missing EFIAPI, one for byte packing, one for unused
variable(returned status) when using PcdSet32S.


spbrogan (approved):
Thanks

PR #21, opened 2019-03-22
Add bars for Platform Mode 2 & 3 and remove STATE_UNDEFINED color bar

joschock:
Provide two additional Platform color bars (Indigo and Brown) and remove STATE_UNDEFINED so we can repurpose its color for one of the platform color bars.


corthon:
Looks good so far. Something we're experimenting with is "document as you go". Can you please add a note to the Main Changes section of the README.rst in the root of the repo with a line about this change (adding the extra color bar)? Thanks!

joschock:
Readme updated.

corthon (approved):

PR #22, opened 2019-04-05
[Feature] Added SharedCrypto (drop-in replacement for BaseCryptLib)

matthewfcarlson:
This uses a precompiled EFI's to speed up compile times when using OpenSSL and decrease space used. This approach can be used by other crypto Libraries other than OpenSSL as long as they conform to the API of BaseCryptLib.


.gitignore
```diff
@@ -5,10 +5,15 @@ tags/
 *.pyc
 *.bak
 BuildConfig.conf
+/Conf
+/.vscode
+SharedCrypto_extdep/
```
spbrogan: this shouldn't be needed as wildcard should cover

.gitignore
```diff
@@ -5,10 +5,15 @@ tags/
...
 
 # Ignore SelfDescribingEnvironment artifacts.
 *_extdep/
 
 # Ignore cloned dependencies
 /Common
 /MU_BASECORE
+.NugetOutput/
```
spbrogan: these also look like local dev relics

SharedCryptoPkg/Driver/CommonSharedCryptoFull.c
```diff
@@ -0,0 +1,161 @@
+
+/** @file
+  This module is consumed by both DXE and SMM as well as PEI
```
spbrogan: typo
Also lets put together a more descriptive description of this module.  What is "full".  What is in the c file

SharedCryptoPkg/Driver/CommonSharedCryptoMu.c
```diff
@@ -0,0 +1,160 @@
+
+/** @file
+  This module is consumed by both DXE and SMM as well as PEI
```
spbrogan: same comment here.  what does the "mu" version mean?

SharedCryptoPkg/Driver/CommonSharedCryptoShaOnly.c
```diff
@@ -0,0 +1,162 @@
+
+/** @file
+  This module is consumed by the reduced version of PEI
```
spbrogan: the file name is more decriptive but still a good decription would be helpful and would be great to explain why this version is important or used.

SharedCryptoPkg/Driver/Dxe/SharedCryptoDxe.c
```diff
@@ -0,0 +1,60 @@
+/** @file
+  This module installs Crypto protocols used by Project Mu Firmware
+
+  Specifically, it installs the DXE protocols listed in the SharedCryptoPkg.dec
```
spbrogan: not true in this version

SharedCryptoPkg/Driver/Smm/SharedCryptoSmm.inf
```diff
@@ -0,0 +1,71 @@
...
+  BaseCryptLib
+
+[Protocols]
+  gSharedCryptoSmmProtocolGuid ## produces
+
+[Depex]
+  TRUE
+
+[BuildOptions]
+  MSFT:*_*_*_CC_FLAGS  = /FAcs /X /GS
```
spbrogan: should dump this

SharedCryptoPkg/Driver/Smm/SharedCryptoSmmMu.inf
```diff
@@ -0,0 +1,71 @@
...
+  BaseCryptLib
+
+[Protocols]
+  gSharedCryptoSmmProtocolGuid ## produces
+
+[Depex]
+  TRUE
+
+[BuildOptions]
+ MSFT:*_*_*_CC_FLAGS  = /FAcs /X /GS /D DEBUG_TARGET /D EFI_DEBUG /D DEBUG_MODE=1
```
spbrogan: same.  dump this...should leverage tools_def

SharedCryptoPkg/Include/Private/Shared/SharedCrypto.h
```diff
@@ -0,0 +1,1203 @@
...
+#define __SHARED_CRYPTO_H__
+
+#define SHARED_CRYPTO_VERSION 1
+
+typedef struct _SHARED_CRYPTO_FUNCTIONS SHARED_CRYPTO_FUNCTIONS;
+
+
+//A function that returns the version number of the packaged crypto efi
+typedef
+UINTN
```
spbrogan: for a multiphase object it is best to define the size to a specific size

SharedCryptoPkg/Include/Private/Shared/SharedCrypto.h
```diff
@@ -0,0 +1,1203 @@
...
+(EFIAPI *CBD_GET_VERSION) (
```
spbrogan: Might be good to remove CBD since it doesn't make sense anymore.

SharedCryptoPkg/Library/CryptLibSharedDriver/DxeCryptLibSharedDriver.c
```diff
@@ -0,0 +1,76 @@
...
+    }
+    else {
+      Version = pCryptoProtocol->SharedCrypto_GetLowestSupportedVersion();
+      if (Version != SHARED_CRYPTO_VERSION)
+      {
+        ProtocolNotFound(EFI_PROTOCOL_ERROR);
+      }
+    }
+  }
+  DEBUG((DEBUG_ERROR, "[CBL_DXE] Found Support Protocol at %x.\n", pCryptoProtocol));
```
spbrogan: not an error

SharedCryptoPkg/Library/CryptLibSharedDriver/DxeCryptLibSharedDriver.c
```diff
@@ -0,0 +1,76 @@
...
+  EFI_STATUS Status
+)
+{
+  DEBUG((DEBUG_ERROR, "[CBL_DXE] Failed to locate Support Protocol. Status = %r\n", Status));
+  ASSERT_EFI_ERROR(Status);
+}
+
+VOID ProtocolFunctionNotFound (CHAR8* function_name)
+{
+  DEBUG((DEBUG_ERROR, "[CBL_DXE] This function was not found: %a\n",function_name));
```
spbrogan: what is CBL

SharedCryptoPkg/Library/CryptLibSharedDriver/DxeCryptLibSharedDriver.c
```diff
@@ -0,0 +1,76 @@
...
+
+**/
+#include <Uefi.h>
+#include <Library/BaseLib.h>
+#include <Library/DebugLib.h>
+#include <Library/UefiBootServicesTableLib.h>
+#include <SharedCryptoHelpers.h>
+#include <Protocol/SharedCryptoProtocol.h>
+
+extern EFI_GUID gSharedCryptoProtocolGuid;
```
spbrogan: this extern shouldn't be needed because it should be externed in the protocol .h

SharedCryptoPkg/Library/CryptLibSharedDriver/DxeCryptLibSharedDriver.inf
```diff
@@ -0,0 +1,60 @@
...
+##
+
+
+[Defines]
+  INF_VERSION                    = 0x00010017
+  BASE_NAME                      = DxeSharedCryptoLib
+  FILE_GUID                      = B38CBDA6-8017-4111-8232-9E8328DE82F6
+  VERSION_STRING                 = 1.0
+  MODULE_TYPE                    = DXE_DRIVER
+  LIBRARY_CLASS                  = SharedCryptoLib
```
spbrogan: need to add supported module types so that this doesn't get linked into the wrong module types

SharedCryptoPkg/Library/CryptLibSharedDriver/PeiCryptLibSharedDriver.c
```diff
@@ -0,0 +1,83 @@
...
+**/
+#include <Uefi.h>
+#include <Library/BaseLib.h>
+#include <Library/DebugLib.h>
+#include <Library/PeiServicesLib.h>
+#include <Library/BaseCryptLib.h>
+#include <Ppi/SharedCryptoPpi.h>
+#include <SharedCryptoHelpers.h>
+
+extern EFI_GUID gSharedCryptoPpiGuid;
```
spbrogan: not needed.  should be in ppi

spbrogan (changes requested):

.gitignore
matthewfcarlson: When publishing a driver, we aggregate the files needed into a new folder called .NugetOutput, maybe this folder should be inside the build folder

.gitignore
matthewfcarlson: Done

SharedCryptoPkg/Driver/CommonSharedCryptoFull.c
matthewfcarlson: Sorry, where's the typo?

.gitignore
matthewfcarlson: Okay it now gets put in the build directory

corthon:
Build failing? Looks like we forgot to exclude an incompatible architecture somewhere.

corthon:
> Build failing? Looks like we forgot to exclude an incompatible architecture somewhere.

Actually, I think it's the removal of /GS, so now we may have extraneous libs.

matthewfcarlson:
I think it's the fact that we're missing BaseSecurityLib and RngLib for X64 and X86

SharedCryptoPkg/Driver/CommonSharedCryptoShaOnly.c
```diff
@@ -0,0 +1,170 @@
...
```
spbrogan: this is like the previous two except obviously more limited in support.  Is there any reason this only supports PEI?

SharedCryptoPkg/Driver/Pei/SharedCryptoPeiShaOnly.inf
```diff
@@ -0,0 +1,66 @@
...
+  PeimEntryPoint
+  BaseCryptLib
+  DebugLib
+  PeiServicesLib
+
+[Ppis]
+  gSharedCryptoPpiGuid ## produces
+
+[Depex]
+  gEfiPeiMemoryDiscoveredPpiGuid
```
spbrogan: why is this needed?  Does this only work post memory init?

SharedCryptoPkg/Driver/Smm/SharedCryptoSmm.inf
```diff
@@ -0,0 +1,70 @@
...
+  BASE_NAME                      = SharedCryptoSmm
+  FILE_GUID                      = 391B853F-F488-479B-A3D6-870766C7A38F
+  MODULE_TYPE                    = DXE_SMM_DRIVER
+  VERSION_STRING                 = 1.0
+  ENTRY_POINT                    = SharedCryptoSmmEntry
+
+#
+# The following information is for reference only and not required by the build tools.
+#
+#  VALID_ARCHITECTURES           = IA32 X64 AARCH64
```
spbrogan: aarch64 isn't valid for DXE_SMM_DRIVER type.

SharedCryptoPkg/Library/CryptLibSharedDriver/PeiCryptLibSharedDriver.inf
```diff
@@ -0,0 +1,62 @@
...
+##
+
+
+[Defines]
+  INF_VERSION                    = 0x00010017
+  BASE_NAME                      = PeiSharedCryptoLib
+  FILE_GUID                      = 3E8B50C6-F68C-4212-B903-94A10FE02399
+  VERSION_STRING                 = 1.0
+  MODULE_TYPE                    = PEIM
+  LIBRARY_CLASS                  = SharedCryptoLib
```
spbrogan: this should be BaseCryptLib | PEIM.  It should be <library class> | <module types supported>

SharedCryptoPkg/Library/CryptLibSharedDriver/SmmCryptLibSharedDriver.c
```diff
@@ -0,0 +1,71 @@
...
+    Status = gSmst->SmmLocateProtocol(&gSharedCryptoSmmProtocolGuid, NULL, (VOID **)&pCryptoProtocol);
+    if (EFI_ERROR(Status))
+    {
+      ProtocolNotFound(Status);
+    }
+    else {
+      Version = pCryptoProtocol->SharedCrypto_GetLowestSupportedVersion();
+      if (Version != SHARED_CRYPTO_VERSION)
+      {
+        ProtocolNotFound(EFI_PROTOCOL_ERROR);
```
spbrogan, line 52: pCryptoProtocol needs to be set NULL otherwise 2nd call will try to use protocol.

SharedCryptoPkg/Library/CryptLibSharedDriver/SmmCryptLibSharedDriver.c
```diff
@@ -0,0 +1,71 @@
...
+      }
+    }
+  }
+  return pCryptoProtocol;
```
spbrogan, line 56: unsafe when asserts are turned off.

SharedCryptoPkg/Package/SharedCryptoPkgDxe.inf
```diff
@@ -0,0 +1,61 @@
...
+#
+
+# these binaries are resolved by the Nuget Package
+# if they are not found, check there
+
+[Binaries]
+  DXE_DEPEX|SharedCryptoPkg/Package/Mu-SharedCrypto_extdep/X64/SharedCryptoDxe.depex|*
+
+[Binaries.X64]
+  PE32|SharedCryptoPkg/Package/Mu-SharedCrypto_extdep/X64/SharedCryptoDxe.efi|*
```
spbrogan: is the sharedcryptopkg/package needed.  Usually it is relative to inf

SharedCryptoPkg/Driver/CommonSharedCryptoShaOnly.c
matthewfcarlson: Does that fix things?

SharedCryptoPkg/Driver/Pei/SharedCryptoPeiShaOnly.inf
matthewfcarlson: I actually don't have a good answer for this. I seem to remember that this was before we had a great depex solution (capturing the depex file from the efi build). I'll try removing this on Monday and making sure we still boot.

SharedCryptoPkg/Library/CryptLibSharedDriver/SmmCryptLibSharedDriver.c
matthewfcarlson: Good catch

SharedCryptoPkg/Library/CryptLibSharedDriver/DxeCryptLibSharedDriver.inf
matthewfcarlson: How's this?

SharedCryptoPkg/Driver/Dxe/SharedCryptoDxe.inf
```diff
@@ -0,0 +1,67 @@
...
+  ENTRY_POINT                    = SharedCryptoDxeEntry
+
+#
+# The following information is for reference only and not required by the build tools.
+#
+#  VALID_ARCHITECTURES           = IA32 X64 AARCH64
+#
+
+[Sources]
+  ../CommonSharedCryptoFull.c
```
mdkinney: INF files should never use '..' in the [Sources] section.  This can be resolved by moving the INF files to the root module directory and put shared sources in a directory below the INF file.

This comment applies to all INF files in the PR that use '..'

SharedCryptoPkg/Include/Private/Shared/SharedCrypto.h
```diff
@@ -0,0 +1,1203 @@
+
```
mdkinney: I see a single structure is used for the PPI, DXE Protoco, and SMM Protocol.  Instead of a shared folder, move the shared content to the Protocol directory and use a type name for the DXE Protocol.  Then from the PPI and SMM Protocol include files, include the Protocol and typedef the PPI and SMM Protocol from the DXE Protocol.

I like the use of Private to not make the PPI, Protocol, and SMM Protocol public.

SharedCryptoPkg/Include/Private/Shared/SharedCrypto.h
```diff
@@ -0,0 +1,1203 @@
...
+#define SHARED_CRYPTO_VERSION 1
+
+typedef struct _SHARED_CRYPTO_FUNCTIONS SHARED_CRYPTO_FUNCTIONS;
+
+
+//A function that returns the version number of the packaged crypto efi
+typedef
+UINTN
+(EFIAPI *SHAREDCRYPTO_GET_VERSION) (
```
mdkinney: Please add comment blocks for each API.  Should be the same as BaseCryptLib.

SharedCryptoPkg/Include/Private/Shared/SharedCrypto.h
```diff
@@ -0,0 +1,1203 @@
...
+  );
+
+
+typedef
+BOOLEAN
+(EFIAPI *SHAREDCRYPTO_ARC4_Reset) (
+  IN OUT  VOID  *Arc4Context
+  );
+
+#pragma pack(push, 1)
```
mdkinney: Should not be any reason to use #pragma pack for a PPI, Protocol, SMM Protocol.

SharedCryptoPkg/Library/CryptLibSharedDriver/Common/CommonSharedCryptoLib.c
```diff
@@ -0,0 +1,3884 @@
...
+  @return  The size, in bytes, of the context buffer required for MD4 hash operations.
+  @retval  0   This interface is not supported.
+
+**/
+UINTN
+EFIAPI
+Md4GetContextSize (
+    VOID)
+{
+  SHARED_CRYPTO_FUNCTIONS *prot = GetProtocol();
```
mdkinney, line 58: Instead of looking up the protocol on every API, it would be better if the library has a CONSTRUCTOR that looks up the protocol and stores it in a module global and the lib instance has a [Depex] on the protocol GUID.  This will work for DXE and SMM instances.  If the PEI instance is in an XIP module, then the module global will not work, so you may need to use this style of the PEI use case.

SharedCryptoPkg/Library/CryptLibSharedDriver/DxeCryptLibSharedDriver.c
```diff
@@ -0,0 +1,74 @@
...
+{
+  DEBUG((DEBUG_ERROR, "[SharedCryptoLibrary_DXE] Failed to locate Support Protocol. Status = %r\n", Status));
+  ASSERT_EFI_ERROR(Status);
+  pCryptoProtocol = NULL;
+}
+
+VOID ProtocolFunctionNotFound (CHAR8* function_name)
+{
+  DEBUG((DEBUG_ERROR, "[SharedCryptoLibrary_DXE] This function was not found: %a\n",function_name));
+  ASSERT_EFI_ERROR(EFI_PROTOCOL_UNREACHABLE);
```
mdkinney: This error code is defined as an ICMP error.  I recommend changing to something like EFI_UNSUPPORTED

SharedCryptoPkg/Library/CryptLibSharedDriver/DxeRuntimeCryptLibSharedDriver.inf
```diff
@@ -0,0 +1,60 @@
...
+##
+
+
+[Defines]
+  INF_VERSION                    = 0x00010017
+  BASE_NAME                      = SharedCryptoLibDxeRuntime
+  FILE_GUID                      = BE6B0F1E-E315-4223-9A0E-BE0BF35A210B
+  VERSION_STRING                 = 1.0
+  MODULE_TYPE                    = DXE_RUNTIME_DRIVER
+  LIBRARY_CLASS                  = SharedCryptoLib
```
mdkinney: Where is the lib class defined?

SharedCryptoPkg/Package/SharedCryptoPkgDxe.inf
```diff
@@ -0,0 +1,61 @@
...
+#
+
+# these binaries are resolved by the Nuget Package
+# if they are not found, check there
+
+[Binaries]
+  DXE_DEPEX|Mu-SharedCrypto_extdep/X64/SharedCryptoDxe.depex|*
+
+[Binaries.X64]
+  PE32|Mu-SharedCrypto_extdep/X64/SharedCryptoDxe.efi|*
```
mdkinney: Where is this path?

It may be better to merge the source content of this PR into the CryptoPkg, and to build releases of the CryptoPkg into a binary release of the CryptoPkg that contains the PEIM, DXE Driver, and DXE SMM Driver.

SharedCryptoPkg/SharedCryptoPkg.dec
```diff
@@ -0,0 +1,51 @@
...
+[Defines]
+  DEC_SPECIFICATION              = 0x00010005
+  PACKAGE_NAME                   = SharedCryptoPkg
+  PACKAGE_UNI_FILE               = SharedCryptoPkg.uni
+  PACKAGE_GUID                   = 0E9D6868-7D81-4A21-9785-BA1021F3F63B
+  PACKAGE_VERSION                = 0.10
+
+
+[Includes.common.Private]
+  Include/Private
```
mdkinney: Do not put Private below Include.  When a package has both public and private interfaces, the public interfaces are typically below Include.  This means consumers of the package can still access the Private dirs with <Private/Protocol/Protocol.h>.  If the Private path is at the top level such as PrivateInclude, then consunmers can not get to the include files.

SharedCryptoPkg/Package/SharedCryptoPkgDxe.inf
matthewfcarlson: The extdep resolved into SharedCryptoPkg/Package. You can see the dependency in SharedCryptoPkg\Package\SharedCrypto_ext_dep.json. I like that idea of moving this into CryptoPkg- makes it simpler as the header is in the same place as the wrapper

SharedCryptoPkg/SharedCryptoPkg.dec
matthewfcarlson: Oh that's a great point, I'll fix this now

SharedCryptoPkg/Driver/Dxe/SharedCryptoDxe.inf
matthewfcarlson: Excellent point. Should be fixed.

SharedCryptoPkg/Include/Private/Shared/SharedCrypto.h
matthewfcarlson: All comments added from BaseCryptLib

SharedCryptoPkg/Include/Private/Shared/SharedCrypto.h
matthewfcarlson: We were having some trouble without this (strange memory errors and page faults in PEI and SMM). It could be possible that this is no longer needed. I'll test without it tomorrow on a device to make sure it can boot without it.

SharedCryptoPkg/Library/CryptLibSharedDriver/Common/CommonSharedCryptoLib.c
matthewfcarlson: That's sort of what we do. For DXE and SMM, we store the found pointer and don't do a LocateProtocol if we have already found it before. There is a DEPEX for these libraries on their respective protocols so DXE modules that rely on BaseCryptLib won't load if SharedCrypto doesn't.

matthewfcarlson: I'll look into using a constructor instead.

SharedCryptoPkg/Library/CryptLibSharedDriver/DxeCryptLibSharedDriver.c
matthewfcarlson: Good idea- fixed.

SharedCryptoPkg/Driver/CommonSharedCryptoFull.c
spbrogan: not seeing it anymore.  must have been a late night. :)

SharedCryptoPkg/Library/CryptLibSharedDriver/DxeCryptLibSharedDriver.inf
```diff
@@ -0,0 +1,60 @@
...
+##
+
+
+[Defines]
+  INF_VERSION                    = 0x00010017
+  BASE_NAME                      = DxeSharedCryptoLib
+  FILE_GUID                      = B38CBDA6-8017-4111-8232-9E8328DE82F6
+  VERSION_STRING                 = 1.0
+  MODULE_TYPE                    = DXE_DRIVER
+  LIBRARY_CLASS                  = BaseCryptLib | DXE_DRIVER
```
spbrogan: how about UEFI apps and UEFI drivers?  
Best link i can find is here: https://edk2-docs.gitbooks.io/edk-ii-inf-specification/content/2_inf_overview/24_[defines]_section.html#24-defines-section look at the Library_class tag. 
and here: https://edk2-docs.gitbooks.io/edk-ii-inf-specification/content/appendix_f_module_types.html

SharedCryptoPkg/Library/CryptLibSharedDriver/PeiCryptLibSharedDriver.c
```diff
@@ -0,0 +1,78 @@
...
+        ProtocolNotFound(EFI_PROTOCOL_ERROR);
+      }
+
+    }
+  }
+
+  return pCryptoProt;
+}
+
+VOID ProtocolNotFound (
```
spbrogan, line 68: same bug here as dxe version if the version doesn't match you still return a pointer rathern than NULL

SharedCryptoPkg/PlatformBuild.py
```diff
@@ -0,0 +1,34 @@
+##
+## Script to Build Kingsburg UEFI firmware
```
spbrogan: clean up comment

SharedCryptoPkg/PlatformBuild.py
```diff
@@ -0,0 +1,34 @@
...
+#
+
+# Smallest 'main' possible. Please don't add unnecessary code.
+if __name__ == '__main__':
+  # If CommonBuildEntry is not found, the mu_environment pip module has not been installed correctly
+  try:
+    from MuEnvironment import CommonBuildEntry
+  except ImportError:
+    print("Running Python version {0} from {1}".format(sys.version, sys.executable))
+    raise RuntimeError("Please run \"python -m pip install --upgrade mu_build\".\nContact MS Core UEFI team if you run into any problems.")
```
spbrogan: can you update to use the requirements.txt file and change to "contact Project Mu team"

SharedCryptoPkg/PlatformBuildWorker.py
```diff
@@ -0,0 +1,272 @@
...
+        self.env.SetValue("LaunchBuildLogProgram", "Notepad",
+                          "default - will fail if already set", True)
+        self.env.SetValue("LaunchLogOnSuccess", "True",
+                          "default - will fail if already set", True)
+        self.env.SetValue("LaunchLogOnError", "True",
+                          "default - will fail if already set", True)
+
+        return 0
+
+    def __SetEsrtGuidVars(self, var_name, guid_str, desc_string):
```
spbrogan: dont need

SharedCryptoPkg/PlatformBuildWorker.py
```diff
@@ -0,0 +1,272 @@
...
+        if(rc != 0):
+            logging.critical("ComputeVersionValue failed")
+            return rc
+
+        #
+        # Validate that the value is correct for this product.
+        #
+        rc = self.ValidateVersionValue()
+        if(rc != 0):
+            logging.critical("ValidateVersionValue failed")
```
spbrogan: this "version" stuff is not valid for this build

SharedCryptoPkg/PlatformBuildWorker.py
```diff
@@ -0,0 +1,272 @@
...
+
+    # ------------------------------------------------------------------
+    #
+    # Method to build capsule and package as a windows driver package
+    #
+    # ------------------------------------------------------------------
+    def BuildCapsule(self):
+        return 0
+
+    def ComputeVersionValue(self):
```
spbrogan: dump this

SharedCryptoPkg/PlatformBuildWorker.py
```diff
@@ -0,0 +1,272 @@
...
+
+        logging.debug("Return value from GitDiffGatedBuild %d" % ret)
+        return True
+
+    # ------------------------------------------------------------------
+    #
+    # Method to build capsule and package as a windows driver package
+    #
+    # ------------------------------------------------------------------
+    def BuildCapsule(self):
```
spbrogan: dump this

SharedCryptoPkg/PlatformBuildWorker.py
```diff
@@ -0,0 +1,272 @@
...
+            dateint = (y << 9) | (m << 5) | d
+
+            # set build id value
+            self.env.SetValue("BLD_*_BUILDID", str(buildint),
+                              "BuildId Computed")
+            self.env.SetValue("BLD_*_BUILDDATE", dateint,
+                              "Computed during Build Process")
+        return 0
+
+    def ValidateVersionValue(self):
```
spbrogan: dump this

SharedCryptoPkg/Readme.md
```diff
@@ -0,0 +1,113 @@
+# Crypto Bin Package
```
spbrogan: update to match naming

spbrogan (changes requested):
Need more changes

SharedCryptoPkg/Driver/Pei/SharedCryptoPeiShaOnly.inf
matthewfcarlson: Alright- this is removed and it's been verified booting.

SharedCryptoPkg/Library/CryptLibSharedDriver/PeiCryptLibSharedDriver.c
matthewfcarlson: Fixed

matthewfcarlson: Fixed

SharedCryptoPkg/PlatformBuild.py
matthewfcarlson: I think you meant this file not requirements.txt

SharedCryptoPkg/PlatformBuildWorker.py
matthewfcarlson: Fixed.

SharedCryptoPkg/PlatformBuildWorker.py
matthewfcarlson: Dumped

SharedCryptoPkg/Library/CryptLibSharedDriver/DxeCryptLibSharedDriver.inf
matthewfcarlson: How's this?

SharedCryptoPkg/Package/SharedCryptoPkgDxe.inf
matthewfcarlson: I think for now we'll have this live in Mu_plus and work on getting this upstreamed to Tianocore in CryptoPkg.

SharedCryptoPkg/Include/Private/Shared/SharedCrypto.h
matthewfcarlson: Alright- I restructured things a little bit that hopefully lines up with what you mentioned in your comment.

SharedCryptoPkg/Library/CryptLibSharedDriver/DxeRuntimeCryptLibSharedDriver.inf
matthewfcarlson: I'm not sure if I understand the question, sorry

matthewfcarlson:
The last thing to do is investigate the tools_def.conf instead of the DSC override.

SharedCryptoPkg/Include/Private/Shared/SharedCrypto.h
matthewfcarlson: Alright removed the pragma pack and it looks like it's still working

mdkinney:
The Private/Include/Protocol directory should have one file for the DXE Protocol and another file for the SMM Protocol.  The SharedProtocol.h can be removed and the DXE Protocol can contain all the contents to define the protocol.  The SMM Protocol file includes the DXE Protocol file and defines the SMM GUID and a typedef for the SMM version of the protocol.  See MdeModulePkg/Include/Protocol/SmmFirmwareVolumeBlock.h for an example to follow.

mdkinney:
The use of the term 'Shared' 'SHARED' should be removed from all types, variables, functions, library, and module names.  The private protocol/ppi/smm protocol are an abstraction for the CryptoPkg/Include/BaseCryptLib.h, so the name of the protocol and the modules should make this relationship obvious.  Perhaps using the names listed below for includes.

The use of 'Base' in the name of the library class is unfortunate.  The term 'Base' in the name of a lib instance implies module compatibility with all module types.

Private/Protocol/BaseCrypt.h
    EDKII_BASE_CRYPT_PROTOCOL_GUID
    EDKII_BASE_CRYPT_PROTOCOL
    gEdkiiBaseCryptProtocolGuid
Private/Protocol/SmmBaseCrypt.h
    EDKII_SMM_BASE_CRYPT_PROTOCOL_GUID
    EDKII_SMM_BASE_CRYPT_PROTOCOL
    gEdkiiSmmBaseCryptProtocolGuid
Private/Ppi/BaseCrypt.h
    EDKII_PEI_BASE_CRYPT_PROTOCOL_GUID
    EDKII_PEI_BASE_CRYPT_PROTOCOL
    gEdkiiPeiBaseCryptProtocolGuid

Library Instances (not using keyword 'Base')
  DxeCryptLibOnProtocol
  SmmCryptLibOnProtocol
  PeiCryptLibOnPpi

Drivers
  CryptDxe
  CryptSmm
  CryptPei

We could define a few feature flag PCDs that select between a NULL value and function pointer value when filling in the structure of crypto services.  This can reduce the number of INFs and C files.

Can we also consider making this a pull request against CryptoPkg in mu_tiano_plus?

matthewfcarlson:
There's been some discussion about moving the PublishDrivers.py, PlatformBuild.py, and the Driver DSC out of Mu_plus. The places that make the most sense for it to move to would be a new repo that would be a submodule of mu_plus (similar to OpenSSL) or in Mu_BASECORE under Basetools since that's where our other nuget publishing tools.

One problem with moving is the eventual goal of moving it upstream to tianocore. Since they don't have anyway to resolve the nuget dependency (as of right now), we should provide the tools for anyone to build their own EFI's to use. If it's in a separate repo, that becomes an extra hurdle for someone to adopt SharedCrypto. Just something we need to be mindful of.

matthewfcarlson:
After some discussion, we've decided to move forward with this PR. Our goal is to land this and deploy it internally for testing. We will stage a PR later that moves this into MU_TIANO_PLUS. At that point, we will restructure the NuGet packaging tools and rename the drivers and libraries to align with BaseCryptLib.

matthewfcarlson:
Add this issue to track the comments Mike K made: https://github.com/Microsoft/mu/issues/59

spbrogan (approved):

corthon (approved):

PR #23, opened 2019-04-25
Replacing GCC_CC_XIP flags for MathLib to allow floating point operations

ashleyknutsen:


MsCorePkg/Library/MathLib/MathLib.inf
```diff
@@ -51,3 +51,6 @@
 [Protocols]
 
 [Guids]
+
+[BuildOptions]
```
spbrogan, line 55: Please add comment why we need this and indicate which flags you are removing

spbrogan (approved):

PR #24, opened 2019-06-05
Changing the status returned by RequestToLock in Foundation.

corthon:
VariablePolicy Disabled is considered a valid condition, but the RequestToLock interface to VariablePolicy would return an error if Disabled. This wouldn't be a problem normally, but many places will ASSERT if RequestToLock returns an error.

This may be re-evaluated in the future, but it should also encourage drivers to move away from VariableLockProtocol and towards VariablePolicyProtocol.


dmitryantipov (approved):

kuqin12 (approved):

corthon:
@dmitryantipov and @kuqin12...
Still look good? I made some changes after you saw it.

kuqin12:
> @dmitryantipov and @kuqin12...
> Still look good? I made some changes after you saw it.

Looks even better ~

corthon:
@spbrogan, this one good, too?

spbrogan:
Yes.  Looks good
________________________________

PR #31, opened 2019-12-23
Avoid extra string manipulation for gEfiStatusCodeDataTypeStringGuid

joschock:
This PR changes the implementation of SerialStatusCodeHandler so that pre-formatted strings (such as ones passed from FSP) don't use AsciiSprint() and are instead passed straight to the writer routine.

### DIFF
--- a/MsCorePkg/Universal/StatusCodeHandler/Serial/Common/SerialStatusCodeHandler.c
+++ b/MsCorePkg/Universal/StatusCodeHandler/Serial/Common/SerialStatusCodeHandler.c
@@ -157,7 +157,7 @@ SerialStatusCode(
   //
   // Call SerialPort Lib function to do print.
   //
-  WriteStatusCode(BufferPtr, CharCount);
+  WriteStatusCode((UINT8 *)BufferPtr, CharCount);
 
 Cleanup:
   return EFI_SUCCESS;
--- a/MsCorePkg/Universal/StatusCodeHandler/Serial/Common/SerialStatusCodeHandler.c
+++ b/MsCorePkg/Universal/StatusCodeHandler/Serial/Common/SerialStatusCodeHandler.c
@@ -61,35 +61,37 @@ SerialStatusCode(
   CHAR8           *Description;
   CHAR8           *Format;
   CHAR8           Buffer[EFI_STATUS_CODE_DATA_MAX_SIZE];
+  CHAR8           *BufferPtr;
   UINT32          ErrorLevel;
   UINT32          LineNumber;
   UINTN           CharCount;
   BASE_LIST       Marker;
 
-  Buffer[0] = '\0';
+  BufferPtr = &Buffer[0];
+  BufferPtr[0] = '\0';
 
-  if (Data != NULL && ReportStatusCodeExtractAssertInfo (CodeType, Value, Data, &Filename, &Description, &LineNumber)) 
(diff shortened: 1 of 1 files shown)


MsCorePkg/Universal/StatusCodeHandler/Serial/Common/SerialStatusCodeHandler.c
```diff
@@ -135,18 +137,15 @@ SerialStatusCode(
     //
     // EFI_STATUS_CODE_STRING_DATA
     //
-    CharCount = AsciiSPrint (
-                  Buffer,
-                  sizeof (Buffer),
-                  "%a\n\r",
-                  ((EFI_STATUS_CODE_STRING_DATA *) Data)->String.Ascii
-                  );
+    BufferPtr = ((EFI_STATUS_CODE_STRING_DATA *) Data)->String.Ascii;
```
corthon, line 140: Will this "in-place" Ascii string have the newlines?
Are the newlines necessary?

MsCorePkg/Universal/StatusCodeHandler/Serial/Common/SerialStatusCodeHandler.c
```diff
@@ -158,7 +157,7 @@ SerialStatusCode(
   //
   // Call SerialPort Lib function to do print.
   //
-  WriteStatusCode((UINT8*)Buffer, CharCount);
+  WriteStatusCode((UINT8 *)BufferPtr, CharCount);
```
corthon, line 160: If BufferPtr is still pointing at Data->String.Ascii, will this print the entire formatted string, or only part of the string?

corthon (changes requested):

corthon:
I'm pretty sure I get what you're doing here, and I'm pretty sure your implementation is correct. Just making sure I understand what's going on.

MsCorePkg/Universal/StatusCodeHandler/Serial/Common/SerialStatusCodeHandler.c
joschock: Yes, it will, and no they are not. This is actually the main reason I'm in here; FSP uses this StatusCode Data Type to print things. And if you add a new line to each thing it print, then:

```
for (i=0; i<something; i++) {
  DEBUG((DEBUG_INFO, "0x%02x ", data[i]));
}
```

looks like this:

```
0xDE
0xAD
0XBE
0xEF
```

instead of

`0xDE 0xAD 0xBE 0xEF `

as expected.

MsCorePkg/Universal/StatusCodeHandler/Serial/Common/SerialStatusCodeHandler.c
joschock: It will print CharCount chars. CharCount is provided by whomever generated the Data->String.Ascii and must be the correct number of chars to print. If someone submits a EFI_STATUS_CODE_STRING_DATA for printing where Data->String.Ascii and Data->DataHeader.Size are inconsistent, that's a bug in the creator of the EFI_STATUS_CODE_STRING_DATA.

MsCorePkg/Universal/StatusCodeHandler/Serial/Common/SerialStatusCodeHandler.c
joschock: Honestly, if we need to minimize changes here for some reason, simply removing the \n\r would solve the main issue.

corthon (approved):

PR #33, opened 2020-02-28
Update SmmPagingProtection tests to add additional cases. 

joschock:
Add a new test suite and several new test cases to SmmPagingProtection


UefiTestingPkg/FunctionalSystemTests/SmmPagingProtectionsTest/Smm/SmmPagingProtectionsTestSmm.c
```diff
@@ -231,13 +257,236 @@ SmmMemoryProtectionsTestInvalidRange (
...
+{
+  DEBUG ((DEBUG_VERBOSE, "%a()\n", __FUNCTION__));
+
+  // Make sure that the required telemetry/handling is
+  // performed to get accurate test results.
+  EnableExceptionTestMode ();
+
+  DEBUG ((DEBUG_INFO, "[%a] - Attempting to write non-authorized MSR IA32_MTRR_PHYSBASE0...\n", __FUNCTION__));
+
+  AsmWriteMsr64(MSR_IA32_MTRR_PHYSBASE0, 0);
```
spbrogan, line 383: i wonder if all the values of "unauthorized*" should be external in header file or maybe even library.  As i can image some differences between silicon partners.

UefiTestingPkg/FunctionalSystemTests/SmmPagingProtectionsTest/App/SmmPagingProtectionsTestApp.inf
```diff
@@ -32,21 +32,26 @@
...
 
 [Protocols]
   gEfiSmmCommunicationProtocolGuid
 
-
 [Guids]
   gEdkiiPiSmmCommunicationRegionTableGuid
+
+[FixedPcd]
+  gUefiTestingPkgTokenSpaceGuid.SkipSmmProtectionFeatures
```
spbrogan, line 57: not sure i like this PCD.  I don't think SmmProtectionFeatures is very clear.  

We try to contain this info in one place.  For example maybe the SMI driver returns an unsupported when the test app requests a feature the platform doesn't support.  In the test this would then take the skipped path.  
Then any config can go into the SMI handler instead of being in both.  Just a thought

UefiTestingPkg/FunctionalSystemTests/SmmPagingProtectionsTest/Smm/SmmPagingProtectionsTestSmm.c
```diff
@@ -12,13 +12,19 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
+#include <Library/IoLib.h>
+#include <Library/MemoryAllocationLib.h>
 #include <Library/SmmServicesTableLib.h>
 #include <Library/SmmMemLib.h>
-#include <Library/DebugLib.h>
-#include <Library/BaseMemoryLib.h>
 
 #include <Protocol/SmmExceptionTestProtocol.h>
 
+#include <Register/Intel/ArchitecturalMsr.h>
```
spbrogan, line 25: This is probably a good indicator that we need a library/abstraction so that our test can apply cross silicon.

PR #37, opened 2020-10-09
Personal/macarl/fix ui rectangle

matthewfcarlson:


MsGraphicsPkg/Include/Library/DisplayDeviceStateLib.h
```diff
@@ -9,22 +9,28 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #ifndef __DEVICE_DISPLAY_STATE_LIB_H__
 #define __DEVICE_DISPLAY_STATE_LIB_H__
 
+#include <Protocol/GraphicsOutput.h>
```
spbrogan: I hate this...i understand why it is desirable and for graphicsOutput it is unfortunate that those types have not been promoted to uefi_types.h or something very common.  But i still think it is better to make caller add this in their c code.  Including from public  header files just creates a web

MsGraphicsPkg/Include/Library/UiRectangleLib.h
```diff
@@ -21,6 +23,7 @@ typedef struct
   UINT32                       Height;
   UINT8*                       FrameBufferBase;
   UINTN                        PixelsPerScanLine;  //in framebuffer
+  UINT8                        PixelSizeInBytes; // the size of the pixel
```
spbrogan: Does this really need to be in the public rectangle object?

MsGraphicsPkg/Library/BaseUiRectangleLib/UiRectangle.c
```diff
@@ -21,12 +21,14 @@ This structure is used by all the other functions to modify and draw the object
...
-This is to support aligned framebuffers
+                           This is to support aligned framebuffers
+@param PixelFormat       - An enum that tells use what format the pixel are in
+@param PixelBitMap       - A pointer to the exact layout of the pixels
 @param Width             - The width of the rectangle
 @param Height            - The height of the rectangle
 @param StyleInfo  - Style info for this (color, sizes, fill types, border, etc)
 
-@ret   A new UI_RECTANGLE structure used for drawing a rectangle
+@ret   A new UI_RECTANGLE structure used for updating and drawing the rectangle
```
spbrogan: what does updating mean?

MsGraphicsPkg/Library/BaseUiRectangleLib/UiRectangle.c
```diff
@@ -59,14 +64,53 @@ IN UI_STYLE_INFO* StyleInfo
...
-  FillDataSize = GetFillDataSize(Width, Height, StyleInfo);
+  DEBUG((DEBUG_INFO, "MATTHEW CARLSON START\n"));
+
+  // This is the default arrangement - 32 bit RGB
+  if (PixelFormat == PixelRedGreenBlueReserved8BitPerColor) {
+    PixelSizeInBytes = 4;
+  }
+
+  // This is the a reverse arrangement of the standard 32 bit BGR (RGB reversed)
+  if (PixelFormat == PixelBlueGreenRedReserved8BitPerColor) {
```
spbrogan: why isn't this else if ()

MsGraphicsPkg/Library/BaseUiRectangleLib/UiRectangle.c
```diff
@@ -59,14 +64,53 @@ IN UI_STYLE_INFO* StyleInfo
...
+  }
+  // This is currently unsupported by this library
+  else if (PixelFormat == PixelBltOnly) {
+    DEBUG((DEBUG_ERROR, "%a:%d - BLT only GOP's are not supported currently\n", __FILE__,__LINE__));
+    ASSERT(PixelFormat != PixelBltOnly);
+  }
+  // If this is bitmask, this mean to check the PixelBitMap and figure out how to handle it
+  else if (PixelFormat == PixelBitMask) {
+    DEBUG((DEBUG_INFO, "Figuring out how to map values from bitmap onto pixel size\n"));
+    ASSERT(PixelFormatBitMap != NULL);
```
spbrogan: need production solution when asserts are off. 
This would lead to NULL pointer dereference

MsGraphicsPkg/Library/BaseUiRectangleLib/UiRectangle.c
```diff
@@ -59,14 +64,53 @@ IN UI_STYLE_INFO* StyleInfo
...
+    UINT32 PixelBitFormat = PixelFormatBitMap->BlueMask | PixelFormatBitMap->GreenMask | PixelFormatBitMap->RedMask | PixelFormatBitMap->ReservedMask;
+    ASSERT(PixelBitFormat != 0); // make sure we have a valid format
+    DEBUG((DEBUG_INFO, "GOP Pixel Format: %x\n", PixelBitFormat));
+    UINT8  BitsInPixel = 1;
```
spbrogan: for (BitsInPixel=0; PixelBitFormat> 0; PixelBitFormat>>=1) { BitsInPixel++; }

Then you don't actually have to check if PixelBitFormat != 0 earlier

MsGraphicsPkg/Library/BaseUiRectangleLib/UiRectangle.c
```diff
@@ -35,12 +37,15 @@ new_UI_RECTANGLE(
 IN POINT *UpperLeft,
 IN UINT8 *FrameBufferBase,
 IN UINTN PixelsPerScanLine,
+IN EFI_GRAPHICS_PIXEL_FORMAT PixelFormat,
+IN EFI_PIXEL_BITMASK* PixelFormatBitMap,
```
spbrogan: is PixelFormatBitMap optional and only required if PixelFormat is PixelBitMask

MsGraphicsPkg/Library/BaseUiRectangleLib/UiRectangle.c
```diff
@@ -59,14 +64,53 @@ IN UI_STYLE_INFO* StyleInfo
...
+  else if (PixelFormat == PixelBitMask) {
+    DEBUG((DEBUG_INFO, "Figuring out how to map values from bitmap onto pixel size\n"));
+    ASSERT(PixelFormatBitMap != NULL);
+    UINT32 PixelBitFormat = PixelFormatBitMap->BlueMask | PixelFormatBitMap->GreenMask | PixelFormatBitMap->RedMask | PixelFormatBitMap->ReservedMask;
+    ASSERT(PixelBitFormat != 0); // make sure we have a valid format
+    DEBUG((DEBUG_INFO, "GOP Pixel Format: %x\n", PixelBitFormat));
+    UINT8  BitsInPixel = 1;
+    while( PixelBitFormat>>=1 ) BitsInPixel++;
+    DEBUG((DEBUG_INFO, "GOP Bits in Pixel: %d\n", BitsInPixel));
+    PixelSizeInBytes = BitsInPixel / 8;
```
spbrogan: what happens if BitsInPexel equals a size not power of 2?  You don't check that here and would cause you to corrupt the buffer.

spbrogan (commented):
I have to think about this some more.  
I think the pixel size/format has leaked into too much code.

spbrogan:
It also might be a good idea to look at some of the open source software GOP drivers to see if they have any better algorithms for this.

matthewfcarlson:
I agree with your comments. I'll address them tomorrow

MsGraphicsPkg/Include/Library/DisplayDeviceStateLib.h
matthewfcarlson: Yeah I agree with you on really despising this

MsGraphicsPkg/Include/Library/UiRectangleLib.h
matthewfcarlson: It's used in a lot of places

MsGraphicsPkg/Library/BaseUiRectangleLib/UiRectangle.c
matthewfcarlson: I just made sure the function comment was the same as in the header file

MsGraphicsPkg/Library/BaseUiRectangleLib/UiRectangle.c
matthewfcarlson: Correct

MsGraphicsPkg/Library/BaseUiRectangleLib/UiRectangle.c
matthewfcarlson: Probably just forgot

MsGraphicsPkg/Library/BaseUiRectangleLib/UiRectangle.c
matthewfcarlson: Agree- this needs to fail gracefully

MsGraphicsPkg/Library/BaseUiRectangleLib/UiRectangle.c
matthewfcarlson: If the bits per pixel is not a power of two, then we have bigger problems. I'm open to solutions though. We should probably round up.

PR #40, opened 2021-03-28
Add support for splitting memory pages when a memory range is not a full page

spbrogan:
There are a few cases where a page might have different memory types if UEFI treats these ranges with the same attributes.


corthon:
This should be retargeted at release/202102, and backported if necessary.

spbrogan:
completed against 202102 #46

PR #42, opened 2021-05-06
Remove gEfiEventPreReadyToBootGuid from Dfci

spbrogan:
This guid is not used by these modules and therefore should not be declared a dependency.  Extra dependencies like this are incorrect and make the code harder to integrate into other code bases.


spbrogan:
Fixes microsoft/mu#111

PR #76, opened 2022-10-24
SUIT List Box element should support cell selection via keyboard even when cell type isn't checkbox.

jeffglaum:
## Description

Keyboard navigation of the SUIT List Box element should support cell selection even when the cell type isn't checkbox (as it is for the FrontPage boot menu options, for example).  This change separates the cell selection from the possible toggling action of a checkbox.

- [ ] Breaking change?
  - Will this change break pre-existing builds or functionality without action being taken? This is not a breaking change.

## How This Was Tested

Tested using FrontPage and a keyboard (both the main frame list box and the boot selection list box).  Also tested with a custom application that uses a list box implementation to show Wi-Fi access points (cBMR application which will be pushed to MU_OEM_SAMPLE shortly).  Now list selections can be made with the space bar and only if the type is checkbox will the toggle event be issued.

## Integration Instructions

NA.


MsGraphicsPkg/Library/SimpleUIToolKit/ListBox.c
```diff
@@ -497,7 +497,7 @@ Draw (
 
     // If the ListBox was created with the checkbox option flag, toggle the checkbox select state.
     //
-    if ((L' ' == pKey->Key.UnicodeChar) && (UIT_LISTBOX_FLAGS_CHECKBOX == (this->m_Flags & UIT_LISTBOX_FLAGS_CHECKBOX))) {
+    if ((L' ' == pKey->Key.UnicodeChar)) {
```
makubacki: You can clean up an extra set of parentheses here.

makubacki (approved):

MsGraphicsPkg/Library/SimpleUIToolKit/ListBox.c
jeffglaum: Sounds good.  Done.

mikeytdisco (approved):

PR #77, opened 2022-10-24
Added new Progress Bar UI element to SUIT.

jeffglaum:
## Description

Added a new progress bar UI element to the SUIT.

- [ ] Breaking change?
  - Will this change break pre-existing builds or functionality without action being taken? This is not a breaking change.

## How This Was Tested

Tested with a UEFI application that uses a progress bar to show download progress (cBMR application which will be pushed to MU_OEM_SAMPLE shortly).  The UI element follows in the template established by earlier UI elements.

## Integration Instructions

NA


jeffglaum:
Hmm.  I used uncrustify to format the sources but for some reason it seems to be failing the CI gate.  I'll look into whether I have the right/same version.

jeffglaum:
> Hmm. I used uncrustify to format the sources but for some reason it seems to be failing the CI gate. I'll look into whether I have the right/same version.

Was using older config file.  Updated and refreshed affected sources.

MsGraphicsPkg/Library/SimpleUIToolKit/SimpleUIToolKit.inf
```diff
@@ -1,61 +1,60 @@
...
+  gMsOSKProtocolGuid
+
+[Guids]
+
+[Pcd]
+  gMsGraphicsPkgTokenSpaceGuid.PcdMsGopOverrideProtocolGuid
+
+[BuildOptions]
+  *_VS2013x86_*_CC_FLAGS       = /Oi /Og /fp:fast /arch:SSE
+#  *_VS2015x86_*_CC_FLAGS       = /Od /FAcs
```
spbrogan: these old compiler flags should get cleaned up while you are here

spbrogan:
would it be possible/easy/reasonable to replace https://github.com/microsoft/mu_basecore/blob/release/202208/MdeModulePkg/Library/DisplayUpdateProgressLibGraphics/DisplayUpdateProgressLibGraphics.c 

Have you reviewed the above?

jeffglaum:
> would it be possible/easy/reasonable to replace https://github.com/microsoft/mu_basecore/blob/release/202208/MdeModulePkg/Library/DisplayUpdateProgressLibGraphics/DisplayUpdateProgressLibGraphics.c
> 
> Have you reviewed the above?



> would it be possible/easy/reasonable to replace https://github.com/microsoft/mu_basecore/blob/release/202208/MdeModulePkg/Library/DisplayUpdateProgressLibGraphics/DisplayUpdateProgressLibGraphics.c
> 
> Have you reviewed the above?

Thanks.  I'm aware of that separate progress bar implementation.  The context in which it's used is slightly different but I'll have a look to see if they can be merged.

MsGraphicsPkg/Library/SimpleUIToolKit/SimpleUIToolKit.inf
jeffglaum: Thanks.  Removed.

MsGraphicsPkg/Library/SimpleUIToolKit/SimpleUIToolKit.inf
```diff
@@ -57,4 +57,3 @@
 
 [BuildOptions]
   *_VS2013x86_*_CC_FLAGS       = /Oi /Og /fp:fast /arch:SSE
```
mikeytdisco: And you want to keep VS2103 why?

jeffglaum: What's the new normalized VS version?  Since these flags are scope to a VS compiler but suitably generic, can I just use MSFT:*_*_*_CC_FLAGS (will that work for any VS version?)?

MsGraphicsPkg/Library/SimpleUIToolKit/SimpleUIToolKit.inf
jeffglaum: Looking into how to generically update the CC flags for all versions as these should be version-generic...

spbrogan: does this work?
``` ini
MSFT:*_*_*_CC_FLAGS = <what ever you want here>
```

jeffglaum: The flags are no longer needed (or in some cases, even valid).  Removed.

mikeytdisco (approved):

jeffglaum:
> would it be possible/easy/reasonable to replace https://github.com/microsoft/mu_basecore/blob/release/202208/MdeModulePkg/Library/DisplayUpdateProgressLibGraphics/DisplayUpdateProgressLibGraphics.c
> 
> Have you reviewed the above?

Looked into this and I don't see an easy way to leverage much of this in common.  The SUIT implementation of the progress bar necessarily is dependent on the SWM for rendering into a window (and SRE surface) whereas the DisplayUpdateProgressLibGraphics implementation doesn't seem able to take that dependency and talks to GOP directly.  The core drawing code could probably be made common but in the end that's just a couple rectangle fills.  Unfortunately the contexts in which they're used different enough that I don't see an easy way to merge them.

I further looked at capabilities/features between the two and with the exception that DisplayUpdateProgressLibGraphics could render the progress bar "over" the boot splashscreen logo (as mask), there don't appear to be any meaningful feature or error handling differences to synchronize between the two.

At this point I don't see further actions to be taken on this one.

jeffglaum:
Moving these changes to the new default branch (release/202208).

PR #80, opened 2022-10-25
Update reset notification list to be notified before NVMe

mikeytdisco:
## Description

Reset notification was handled by the NVMe driver before the AdvLogger could write the log to disk.

- [No] Breaking change?

## How This Was Tested

Tested on multiple systems.

## Integration Instructions

N/A

### DIFF
--- a/AdvLoggerPkg/AdvancedFileLogger/AdvancedFileLogger.c
+++ b/AdvLoggerPkg/AdvancedFileLogger/AdvancedFileLogger.c
@@ -126,7 +126,7 @@ OnResetNotificationProtocolInstalled (
   // Get a pointer to the report status code protocol.
   //
   Status = gBS->LocateProtocol (
-                  &gEfiResetNotificationProtocolGuid,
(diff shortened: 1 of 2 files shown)


makubacki:
Is/was there a change to the NVMe code? If so, could you link that in the description for reference?

mikeytdisco:
There was no change to the NVMe code.  The Reset System driver processes 3 queues.  Both NVMe and AdvLogger were using the second queue processed by ResetSystemRuntimeDxe. AdvLogger was changed to use the first queue processed by ResetSystemRuntimeDxe.

spbrogan:
Not sure about this change.  You are leveraging the "filter" list.
When we developed the new reset code we wrote some documentation about how it should be used.  Do you know where that is?  I don't see it in edk2.

spbrogan (changes requested):
I would like to see the docs and confirm this is acceptable use

apop5 (approved):

spbrogan (approved):
this looks like it should work and is aligned with the requirements of the api headers.

PR #81, opened 2022-10-25
Added a new SUIT progress bar UI element.

jeffglaum:
Description
Added a new progress bar UI element to the SUIT.

Breaking change?
This is not a breaking change.

How This Was Tested
Tested with a UEFI application that uses a progress bar to show download progress (cBMR application which will be pushed to MU_OEM_SAMPLE shortly). The UI element follows in the template established by earlier UI elements.

### DIFF
--- a/MsGraphicsPkg/Include/UIToolKit/ProgressBar.h
+++ b/MsGraphicsPkg/Include/UIToolKit/ProgressBar.h
@@ -1,0 +1,72 @@
+/** @file
+
+  Implements a simple progress bar control for showing incremental progress.
+
+Copyright (C) Microsoft Corporation. All rights reserved.
+SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef _UIT_PROGRESSBAR_H_
+#define _UIT_PROGRESSBAR_H_
+
+//
+typedef struct _ProgressBarDisplayInfo {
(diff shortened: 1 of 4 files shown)


jeffglaum:
Moved PR to the release/202208 (new default) branch.

Earlier request: 

would it be possible/easy/reasonable to replace https://github.com/microsoft/mu_basecore/blob/release/202208/MdeModulePkg/Library/DisplayUpdateProgressLibGraphics/DisplayUpdateProgressLibGraphics.c

Have you reviewed the above?

Looked into this and I don't see an easy way to leverage much of this in common. The SUIT implementation of the progress bar necessarily is dependent on the SWM for rendering into a window (and SRE surface) whereas the DisplayUpdateProgressLibGraphics implementation doesn't seem able to take that dependency and talks to GOP directly. The core drawing code could probably be made common but in the end that's just a couple rectangle fills. Unfortunately the contexts in which they're used different enough that I don't see an easy way to merge them.

I further looked at capabilities/features between the two and with the exception that DisplayUpdateProgressLibGraphics could render the progress bar "over" the boot splashscreen logo (as mask), there don't appear to be any meaningful feature or error handling differences to synchronize between the two.

At this point I don't see further actions to be taken on this one.

jeffglaum:
Further requesting this be back-ported to the release/202202 branch (tag/label required?).  Thanks.

mikeytdisco (approved):

spbrogan (approved):

PR #84, opened 2022-11-01
Update the finding max mode logic to account for vertical resolution as well

kuqin12:
## Description

The exiting logic would only look for the maximal horizontal resolution. The issue is that if the supported modes have other modes that supports more vertical pixels, this logic will truncate the display to a shorter screen with the same number of horizontal pixels.

- [ ] Breaking change?

## How This Was Tested

This change is verified on QEMU virtual platforms.

## Integration Instructions

N/A


os-d (approved):

mikeytdisco:
I'm not sure that works (and I'll add either).  A mode may be supported that supports a monitor that is higher than it is wide.  The original code was for the simple case of 800x600, 1024,768, and 1280,1024.  What happens if there is a 1024x1280 resolution?

kuqin12:
> I'm not sure that works (and I'll add either). A mode may be supported that supports a monitor that is higher than it is wide. The original code was for the simple case of 800x600, 1024,768, and 1280,1024. What happens if there is a 1024x1280 resolution?

Synced offline. This will be handled in the GOP driver.

PR #87, opened 2022-11-02
Add Stack Cookie Interrupt Handler

TaylorBeebe:
## Description

Stack cookie check failure will emit a 0x40 interrupt - add a handler for stack cookie check failure Interrupts to MemoryProtectionExceptionHandlerLib

## Breaking change?
No

## How This Was Tested

Enabling Stack Cookies on Q35

## Integration Instructions

N/A


Flickdm (approved):

MsCorePkg/Library/MemoryProtectionExceptionHandlerLib/MemoryProtectionExceptionHandlerLib.c
```diff
@@ -24,7 +24,8 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/MsWheaEarlyStorageLib.h>
 #include <Library/PeCoffGetEntryPointLib.h>
 
-#define IA32_PF_EC_ID  BIT4
+#define IA32_PF_EC_ID        BIT4
+#define EXCEPT_STACK_COOKIE  0x40
```
makubacki, line 28: It would be kind of nice if a globally visible FixedAtBuild PCD could set the interrupt value used everywhere. Did you consider that?

makubacki (approved):

MsCorePkg/Library/MemoryProtectionExceptionHandlerLib/MemoryProtectionExceptionHandlerLib.c
TaylorBeebe: I didn't consider that, but that seems like a potentially confusing design unless the stack cookie code also uses the fixed at build PCD when generating the interrupt. I suppose we can hook the PCD value into the stack cookie check code if there aren't any restrictions on the execution environment.

PR #96, opened 2022-11-08
Remove Page Audit Report Generator Print

TaylorBeebe:
## Description

The PTE GCD type may be None, so attempts to log it will cause a runtime exception.

## Breaking change?

No

## How This Was Tested

Running the script

## Integration Instructions

N/A


makubacki:
This is interesting in that it explicitly attempts to print if `pte.GcdType` is `None`. There's a lot of similar checks against `None` in the file, did you confirm there's not another case of this elsewhere?

```python
                        if pte.GcdType is None:
                            pte.GcdType = mr.GcdType
                            logging.info("pte.GcdType: %d"% (pte.GcdType))
```

makubacki (approved):

apop5 (approved):

PR #101, opened 2022-11-10
Update DXE Paging Audit App to Include RWX Test

TaylorBeebe:
## Description

Our memory protection policy is now robust enough to ensure that platforms have no read/write/execute pages before ExitBootServices. This update adds a test to the DxePagingAuditApp to check the page table for RWX pages and only exempt them if the region is part of a nonprotected image or special region.

Users can still utilize the app to dump paging data to the EFI partition by calling the application with the '-d' flag. By default, the app will run the RWX test.

## Breaking change?

No

## How This Was Tested

Running the test on Q35

## Integration Instructions

The test will identify RWX regions. Platforms should identify these regions to determine if they must be RWX. If they really must be RWX, the platform can utilize the Memory Protection Special Region Protocol to create a special region.

### DIFF
--- a/UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
+++ b/UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
@@ -137,6 +137,8 @@ NoReadWriteExcecute (
       if (!IgnoreRWXAddress) {
         UT_LOG_ERROR ("Memory Range 0x%llx-0x%llx is Read/Write/Execute\n", Map[Index].LinearAddress, Map[Index].LinearAddress + Map[Index].Length);
         FoundRWXAddress = TRUE;
+      } else {
+        UT_LOG_WARNING ("Memory Range 0x%llx-0x%llx is Read/Write/Execute. This range is excepted from the test.\n", Map[Index].LinearAddress, Map[Index].LinearAddress + Map[Index].Length);
       }
     }
   }
--- a/UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
+++ b/UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
@@ -194,14 +194,16 @@ DxePagingAuditTestAppEntryPoint (
     RunTests = FALSE;
     if (StrnCmp (ShellParams->Argv[1], L"-r", 4) == 0) {
       RunTests = TRUE;
-    } else if (StrnCmp (ShellParams->Argv[1], L"-h", 4) == 0) {
-      DEBUG ((DEBUG_INFO, "-h : Print available flags\n"));
-      DEBUG ((DEBUG_INFO, "-d : Dump the page table files to the EFI partition\n"));
-      DEBUG ((DEBUG_INFO, "-r : Run the application tests\n"));
     } else if (StrnCmp (ShellParams->Argv[1], L"-d", 4) == 0) {
       DumpPagingInfo (NULL, NULL);
     } else {
-      DEBUG ((DEBUG_INFO, "Invalid argument. Use \'-h\' to see a list of valid arguments.\n"));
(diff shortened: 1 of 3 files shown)


UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
```diff
@@ -29,7 +163,124 @@ DxePagingAuditTestAppEntryPoint (
...
+  }
+
+  if (ShellParams->Argc > 1) {
+    RunTests = FALSE;
+    if (StrnCmp (ShellParams->Argv[1], L"-r", 4) == 0) {
+      RunTests = TRUE;
+    } else if (StrnCmp (ShellParams->Argv[1], L"-h", 4) == 0) {
+      DEBUG ((DEBUG_INFO, "-h : Print available flags\n"));
+      DEBUG ((DEBUG_INFO, "-d : Dump the page table files to the EFI partition\n"));
+      DEBUG ((DEBUG_INFO, "-r : Run the application tests\n"));
```
os-d: Should the usage mention that combining parameters is invalid? As is, it just looks at the first parameter.

os-d: nit: We could not explicitly check for `-h` and instead dump it in the else. If the user gives any flag besides `-d` or `-r`, we just dump the help.

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
```diff
@@ -29,7 +163,124 @@ DxePagingAuditTestAppEntryPoint (
...
+
+    Cr4.UintN = AsmReadCr4 ();
+
+    if (Cr4.Bits.LA57 != 0) {
+      PagingMode = Paging5Level;
+    } else {
+      PagingMode = Paging4Level;
+    }
+
+    Status = PageTableParse (AsmReadCr3 (), PagingMode, NULL, &MapCount);
```
os-d, line 240: Perhaps adding some comments here on the register reads, i.e. what you are reading (can be inferred, but nice to know for sure).

os-d (approved):

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
TaylorBeebe: Added

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
TaylorBeebe: Updated for both comments

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
```diff
@@ -29,7 +163,129 @@ DxePagingAuditTestAppEntryPoint (
...
+
+    // Poll CR4 to deterimine the page table depth
+    Cr4.UintN = AsmReadCr4 ();
+
+    if (Cr4.Bits.LA57 != 0) {
+      PagingMode = Paging5Level;
+    } else {
+      PagingMode = Paging4Level;
+    }
+
```
apop5, line 238: CR3 is X86 specific. So this app cannot be used on any other architectures?

apop5 (approved):

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
TaylorBeebe: Correct. Even the common portion of the app uses x86 opinionated logic such as x86 page table definitions, so it will need to be broken into x86 and ARM versions in the future.

PR #105, opened 2022-11-15
DfciPkg/UnitTests/pip-requirements.txt: Remove pytool-library; bump robot framework

makubacki:
## Description

- Removes edk2-pytool-library since it has a very old version and
   the module can be picked up from (1) the main repo
   pip-requirements.txt or (2) the edk2-pytool-extensions dependency.

- Updates robotframework pip module version since it also now very old.

- [ ] Breaking change?
  **No** - Should not be breaking for the reasons mentioned in the description.

## How This Was Tested

Todo

## Integration Instructions

N/A

Signed-off-by: Michael Kubacki <michael.kubacki@microsoft.com>

### DIFF
--- a/DfciPkg/UnitTests/DfciTests/pip-requirements.txt
+++ b/DfciPkg/UnitTests/DfciTests/pip-requirements.txt
@@ -1,2 +1,1 @@
-edk2-pytool-library>=0.10.13
-robotframework>=4.0.1
+robotframework>=6.0.1


makubacki:
@spbrogan - I only found this section in the documentation that is related to this change - [DfciPkg/UnitTests/DfciTests - Setting up the HOST system](https://github.com/microsoft/mu_plus/tree/release/202208/DfciPkg/UnitTests/DfciTests#setting-up-the-host-system).

It only references that the `pip-requirements.txt` file should be installed but does not mention the contents. It does reference Python 3.9.4 but updating that is a bit unrelated to this change and probably fine to leave.

makubacki:
@mikeytdisco, this is something I found when going through the pip modules pulled down in our pipelines.

I don't have an environment set up to run the tests. Have you been running them recently? Could you try including this change the next time you run the robot tests?

apop5 (approved):

makubacki:
Comment to bump as a reminder

makubacki:
Moved to issue in mu_feature_dfci - https://github.com/microsoft/mu_feature_dfci/issues/46

PR #108, opened 2022-11-16
Fix unitialized variable found in clang compiler experiments

cfernald:
Fix a uninitialized variable usage case found when experimenting with new clang compiler configuration. If the first `goto exit` statement is used then a comparison will be done against the uninitialized value of the variable


kuqin12 (approved):

makubacki (approved):
I think it would be good to add a description.

You could call out the `goto` case there which is the main issue I see with it not being initialized.

apop5 (approved):

PR #109, opened 2022-11-17
Add MacEmulationDxe Support

Erich-McMillan:
## Description

Adds generic MacAddressEmulationDxe driver which will set the NIC address to a custom value.

This driver is designed to glue together EDK network stack with an OEM provided platform library which will determine whether emulation is enabled, provide the custom mac address and perform any custom platform behavior such as add info to ACPI.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Tested on a physical ARM device running non-MuBase EDK2 based framework with a Surface Ethernet adapter which supports MacEmulation. 
Modifications necessary to validate on ARM non-MuEdk2
- addition of `EfiNamedEventSignal` for `gSnpNetworkInitializedEventGuid` to `SnpUndi32Initialize`. Introduced to MU at [7dc143d594bfef6d2a2a24b10b96e1258a15a344](https://github.com/microsoft/mu_basecore/commit/7dc143d594bfef6d2a2a24b10b96e1258a15a344)
- Updates to MacAddressEmulationDxe.inf to remove MsCorePkg.dec (added gSnpNetworkInitializedEventGuid to Edk2/NetworkPkg/NetworkPkg.dec`

Two test cases were:

1. Boot to PXE boot (no target, just query for PXE servers), capture network traffic via Wireshark and ensure only custom mac address used for ARP/DHCP
2. Boot to EFI Shell, run `ifconfig -l eth0` to confirm custom mac is present. Capture network traffic via Wireshark and ensure when `ifconfig -r eth0; ping 8.8.8.8` all network traffic ARP/DHCP/ICMP uses only custom mac address. 

Automated test results:
image

## Integration Instructions

- Will propose introducing [7dc143d594bfef6d2a2a24b10b96e1258a15a344](https://github.com/microsoft/mu_basecore/commit/7dc143d594bfef6d2a2a24b10b96e1258a15a344) to Tiano/Edk2 for upstream integration so this driver is compatible with Tiano/Edk2
- Add unit tests to CI for mu_plus.


Erich-McMillan:
@microsoft-github-policy-service agree [company="Microsoft"]

Erich-McMillan:
@microsoft-github-policy-service agree company="Microsoft"

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,239 @@
...
+  IN CONST VOID *SnpContext
+  )
+{
+  MAC_EMULATION_SNP_NOTIFY_CONTEXT* Context;
+  BOOLEAN IsMatch = FALSE;
+
+  if (SnpHandle != NULL && Snp != NULL && SnpContext != NULL) {
+    IsMatch = TRUE;
+    Context = (MAC_EMULATION_SNP_NOTIFY_CONTEXT*) SnpContext;
+  }
```
joschock: need an else with a return here? Otherwise you're going to dereference a potentially null Snp pointer

joschock: I guess maybe you are relying on IsMatch to avoid that?

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,239 @@
...
+  IN OUT MAC_EMULATION_SNP_NOTIFY_CONTEXT *Context
+  )
+{
+  EFI_STATUS Status = EFI_NOT_STARTED;
+  EFI_TPL OldTpl;
+
+  DEBUG ((DEBUG_ERROR, "[%a]: Start\n", __FUNCTION__));
+
+  if (Snp != NULL && Context != NULL) {
+    OldTpl = gBS->RaiseTPL(TPL_HIGH_LEVEL);
```
joschock: maybe a block comment with explanatory note about why the TPL adjust is needed.

MsCorePkg/MacAddressEmulationDxe/Test/EntryPointTest.c
```diff
@@ -0,0 +1,113 @@
+#include "SnpSupportsMacEmuCheckTest.h"
+
+/**
+  Unit test for SnpSupportsMacEmuCheck_ReturnsTrue_WhenInterfaceIsSupported_AndInterfaceMatchesPreviouslyAssignedInterface ()
```
joschock: Stale comment?

joschock: And below

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,239 @@
...
+
+  return IsMatch;
+}
+
+/**
+  
+**/
+EFI_SIMPLE_NETWORK_PROTOCOL*
+FindMatchingSnp (
+  IN SNP_MATCH_FUNCTION MatchFunction,
```
rogurr: How come you are using a function pointer here instead of just linking to the function where needed.  I see it only called once in the driver using a single function pointer, and in the test code it uses NULL, but the function is returning invalid parameter using NULL.

This seems like extra work to understand the code, will it be used later?

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,239 @@
...
+  
+  if (Context == NULL) {
+    DEBUG ((DEBUG_ERROR, "[%a]: Context unexpectedly null.\n", __FUNCTION__));
+    ASSERT (Context != NULL);
+    return;
+  }
+
+  MacContext = (MAC_EMULATION_SNP_NOTIFY_CONTEXT*)Context;
+
+  SnpToConfigureEmu = FindMatchingSnp(SnpSupportsMacEmuCheck, (VOID*) MacContext);
```
rogurr: Can we put a check here to return early if the return pointer is NULL?  SetSnpMacViaContext does correctly handle an input NULL, but it is not listed as an optional parameter, and it makes it a little easier to debug when handling an error return higher in the stack.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,239 @@
...
+  UINTN                             HandleIdx;
+  EFI_HANDLE                        *SnpHandleBuffer;
+  EFI_SIMPLE_NETWORK_PROTOCOL       *SnpInstance;
+
+  SnpInstance = NULL;
+  SnpHandleBuffer = NULL;
+
+  if (MatchFunction != NULL) {
+    Status = gBS->LocateHandleBuffer (ByProtocol, &gEfiSimpleNetworkProtocolGuid, NULL, &HandleCount, &SnpHandleBuffer);
+    ASSERT_EFI_ERROR (Status);
```
rogurr: John's original code had a note here that indicated EFI_NOT_FOUND is a valid return due to the callback being executed when the notify is created.  Is that no longer a possibility with TPL change?

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.h
```diff
@@ -0,0 +1,73 @@
...
+typedef
+BOOLEAN
+(*SNP_MATCH_FUNCTION)(
+  IN CONST EFI_HANDLE SnpHandle,
+  IN CONST EFI_SIMPLE_NETWORK_PROTOCOL *Snp,
+  OPTIONAL IN CONST VOID *SnpContext
+  );
+
+BOOLEAN
+SnpSupportsMacEmuCheck (
```
rogurr: Can we get block comments for each function here, in the .C file, and all files in the Test directory?

Just the standard comments of what it does, parameters, and returns.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,239 @@
...
+#include <Library/DebugLib.h>
+#include <Library/UefiLib.h>
+#include <Library/UefiBootServicesTableLib.h>
+#include <Library/NetLib.h>
+
+#include "Library/MacAddressEmulationPlatformLib.h"
+
+
+BOOLEAN
+SnpSupportsMacEmuCheck (
```
rogurr: This function name is very close to the library function SnpSupportsMacEmulation that is called on line 52.  Maybe we need to change this function to SnpSupportsMacEmulation and update the library to something a little more descriptive?  Maybe PlatformSnpMacEmulationSupport or something with platform in the name so OEMs know it is specifically for them to use?

I do like all of the checks here for type, is changeable, etc.  Can we add comments to the NULL library <OEM TODO> area indicating these checks are being done before the library call and let them know they only need to handle things specific to the platform such as returning FALSE if it is a USB dongle they don't want to support?

MsCorePkg/Include/Library/MacAddressEmulationPlatformLib.h
```diff
@@ -0,0 +1,53 @@
...
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+#ifndef _MAC_ADDRESS_EMULATION_PLATFORM_LIB_H_
+#define _MAC_ADDRESS_EMULATION_PLATFORM_LIB_H_
+
+/**
+  Called by the MacAddressEmulationDxe driver's entry function to determine if the platform's MAC emulation is enabled.
```
spbrogan: a library api should not speculate or define who will call it when.  This api can specify the requirements on the environment when it is called.

MsCorePkg/Library/MacAddressEmulationPlatformLibNull/MacAddressEmulationPlatformLibNull.c
```diff
@@ -0,0 +1,112 @@
+/** @file
+
+  NULL implementation of the platform library that provides hooks for the MAC Address Emulation feature.
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+#include <Uefi.h>
+#include <Library/DebugLib.h>
```
spbrogan: not using debuglib

spbrogan: You should probably include your library class here:
``` c
#include <Library/MacAddressEmulationPlatformLib.h>
```

MsCorePkg/Library/MacAddressEmulationPlatformLibNull/MacAddressEmulationPlatformLibNull.inf
```diff
@@ -0,0 +1,33 @@
...
+  MODULE_TYPE                    = DXE_DRIVER
+  VERSION_STRING                 = 1.0
+  LIBRARY_CLASS                  = MacAddressEmulationPlatformLib
+
+[Sources]
+  MacAddressEmulationPlatformLibNull.c
+
+[Packages]
+  MdePkg/MdePkg.dec
+  MdeModulePkg/MdeModulePkg.dec
```
spbrogan: i don't think you actually need MdeModulePkg.

MsCorePkg/Library/MacAddressEmulationPlatformLibNull/MacAddressEmulationPlatformLibNull.inf
```diff
@@ -0,0 +1,33 @@
...
+# NULL implementation of the platform library to provide hooks for the MAC Address Emulation feature.
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+[Defines]
+  INF_VERSION                    = 1.27
+  BASE_NAME                      = MacAddressEmulationPlatformLib
```
spbrogan: I would suggest to add Null here so if symbols are ever used it is easier to decode usage

MsCorePkg/Library/MacAddressEmulationPlatformLibNull/MacAddressEmulationPlatformLibNull.inf
```diff
@@ -0,0 +1,33 @@
...
+  MacAddressEmulationPlatformLibNull.c
+
+[Packages]
+  MdePkg/MdePkg.dec
+  MdeModulePkg/MdeModulePkg.dec
+  MsCorePkg/MsCorePkg.dec
+
+[LibraryClasses]
+  BaseLib
+  DebugLib
```
spbrogan: not using debuglib or baselib in the module

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,239 @@
...
+
+#include <Library/BaseLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/MemoryAllocationLib.h>
+#include <Library/DebugLib.h>
+#include <Library/UefiLib.h>
+#include <Library/UefiBootServicesTableLib.h>
+#include <Library/NetLib.h>
+
+#include "Library/MacAddressEmulationPlatformLib.h"
```
spbrogan: need to use the include path version <> instead of "".  That is not a local file.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.h
```diff
@@ -0,0 +1,73 @@
...
+#define MAC_TO_CHAR8_ARGS(MacAddress) \
+  ((UINT8 *)&MacAddress)[0], \
+  ((UINT8 *)&MacAddress)[1], \
+  ((UINT8 *)&MacAddress)[2], \
+  ((UINT8 *)&MacAddress)[3], \
+  ((UINT8 *)&MacAddress)[4], \
+  ((UINT8 *)&MacAddress)[5]
+
+typedef struct {
+  VOID            *Registration;
```
spbrogan: can this be typed

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.h
```diff
@@ -0,0 +1,73 @@
+/** @file
+
+  DXE Driver header for handling MAC Address Emulation.
+
+  Copyright (C) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#pragma once
```
spbrogan: uefi code generally uses other pattern for protection.  Keep consistent.  Maybe someday we will change all of it.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.h
```diff
@@ -0,0 +1,73 @@
...
+  EFI_MAC_ADDRESS EmulationAddress;
+  EFI_MAC_ADDRESS PermanentAddress;
+} MAC_EMULATION_SNP_NOTIFY_CONTEXT;
+
+typedef
+BOOLEAN
+(*SNP_MATCH_FUNCTION)(
+  IN CONST EFI_HANDLE SnpHandle,
+  IN CONST EFI_SIMPLE_NETWORK_PROTOCOL *Snp,
+  OPTIONAL IN CONST VOID *SnpContext
```
spbrogan: can this be typed to the above context?

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.h
```diff
@@ -0,0 +1,73 @@
...
+  );
+
+VOID
+EFIAPI
+SimpleNetworkProtocolNotify (
+  IN EFI_EVENT  Event,
+  IN VOID       *Context
+  );
+
+EFI_STATUS
```
spbrogan: generally, don't need entry point function in header file

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,239 @@
...
+SnpSupportsMacEmuCheck (
+  IN CONST EFI_HANDLE SnpHandle,
+  IN CONST EFI_SIMPLE_NETWORK_PROTOCOL *Snp,
+  IN CONST VOID *SnpContext
+  )
+{
+  MAC_EMULATION_SNP_NOTIFY_CONTEXT* Context;
+  BOOLEAN IsMatch = FALSE;
+
+  if (SnpHandle != NULL && Snp != NULL && SnpContext != NULL) {
```
spbrogan: I would suggest using a return here.  
Yes have multiple return points from a function can be a bad design but this design leaves a lot of room for mistakes and is much harder to read.   For basic parameter checks I would suggest using a return.

MsCorePkg/MsCorePkg.dec
```diff
@@ -66,6 +66,8 @@
   #
   UpdateFacsHardwareSignatureLib|Include/Library/UpdateFacsHardwareSignatureLib.h
 
+  MacAddressEmulationPlatformLib|Include/Library/MacAddressEmulationPlatformLib.h
```
rogurr, line 71: Each of the above .H files has a comment stating info about the file.  Can we follow suit here and add comments?

MsCorePkg/MacAddressEmulationDxe/Test/EntryPointTest.c
```diff
@@ -0,0 +1,113 @@
+#include "SnpSupportsMacEmuCheckTest.h"
```
rogurr: We probably need headers for info and copyright on all files being added

rogurr: I don't see in this file where the function defined in SnpSupportsMacEmuCheckTest.h is used, but it won't compile without it since it needs the included MacAddressEmulationDxeHostTestCommon.h file.  Can we just include that file directly?

MsCorePkg/MacAddressEmulationDxe/Test/FindMatchingSnpTest.c
```diff
@@ -0,0 +1,217 @@
+#include "SnpSupportsMacEmuCheckTest.h"
+
+EFI_BOOT_SERVICES mBootServices;
```
rogurr: This is defined both here and in the SimpleNetworkProtocolNotifyTest.c files, does it give a warning when linking?  The 2 files use different pointers in the structure so if a 3rd file links to the wrong one we could dereference a NULL pointer.

MsCorePkg/MacAddressEmulationDxe/Test/FindMatchingSnpTest.c
```diff
@@ -0,0 +1,217 @@
...
+  return UNIT_TEST_PASSED;
+}
+
+UNIT_TEST_STATUS
+EFIAPI 
+FindMatchingSnpTestSetup (
+  IN UNIT_TEST_CONTEXT  Context
+  )
+{
+  gBS = &mBootServices;
```
rogurr: Only the following 2 function pointers are initialized in the mBootServices structure.  If code being tested in the driver is changed to use other pointers, I assume we would try to de-reference a NULL.  Should we copy the mBootServices structure to our local copy first before re-assigning gBS and overriding the 2 functions?

MsCorePkg/MacAddressEmulationDxe/Test/MacAddressEmulationDxeHostTestCommon.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include <stdio.h>
+#include <string.h>
+#include <stdarg.h>
+#include <stddef.h>
+#include <setjmp.h>
+#include <cmocka.h>
```
rogurr: I noticed a lot of places the code is using standard C items like these .H files, size_t, int, etc.  Is there a reason we can't use the standard UEFI H files and definitions?  Will this code be compiled outside UEFI?

MsCorePkg/MacAddressEmulationDxe/Test/SimpleNetworkProtocolNotifyTest.c
```diff
@@ -0,0 +1,192 @@
...
+EFIAPI 
+SimpleNetworkProtocolNotifyTestSetup (
+  IN UNIT_TEST_CONTEXT  Context
+  )
+{
+    gBS = &mBootServices;
+    gBS->LocateHandleBuffer = LocateHandleBuffer;
+    gBS->HandleProtocol = HandleProtocol;
+    gBS->RaiseTPL = RaiseTPL;
+    gBS->RestoreTPL = RestoreTPL;
```
rogurr: I see where you put these override functions in the MacAddressEmulationDxeHostTestMain.c file and declared the prototype for HandleProtocol and LocateHandleBuffer in MacAddressEmulationDxeHostTestCommon.h.  But I don't see the prototypes anywhere for the TPL functions.

I'm sure the code compiles, is there maybe a file or something missing when pushed to this sub-module?  Maybe the Common .H file is out of date?

MsCorePkg/MacAddressEmulationDxe/Test/SimpleNetworkProtocolNotifyTest.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include "MacAddressEmulationDxeHostTestCommon.h"
+
+VOID
+RegisterSimpleNetworkProtocolNotifyTests (
+    UNIT_TEST_SUITE_HANDLE SuiteHandle
```
rogurr: Can you maybe run a beautify or walk the files for coding standards?  We are trying to watch for even little things like this using a 4 space indent instead of 2.

MsCorePkg/MacAddressEmulationDxe/Test/MacAddressEmulationDxeHostTestCommon.h
rogurr: I see in the TestMain.c file below there is a standard POSIX C main() function, so I assume you are compiling 2 ways.  Is there a way we can use just UEFI definitions in that situation so engineers digging into this only have to work with 1 coding standard?

MsCorePkg/MacAddressEmulationDxe/Test/MacAddressEmulationDxeHostTestMain.c
```diff
@@ -0,0 +1,166 @@
...
+
+#include "SnpSupportsMacEmuCheckTest.h"
+#include "FindMatchingSnpTest.h"
+#include "SimpleNetworkProtocolNotifyTest.h"
+#include "EntryPointTest.h"
+
+#define UNIT_TEST_NAME     "Mac Address Emulation Dxe Host Test"
+#define UNIT_TEST_VERSION  "0.1"
+
+EFI_RUNTIME_SERVICES  mMockRuntime;
```
rogurr: Is this maybe a holdover from a copy/paste to start coding?  I don't see anywhere that links to this and it doesn't appear you are pulling in the UnitTest MockUefiRuntimeServicesTableLib.

rogurr (changes requested):

MsCorePkg/Library/MacAddressEmulationPlatformLibNull/MacAddressEmulationPlatformLibNull.inf
Erich-McMillan: Also appear to not need MsCorePkg, removing both.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
Erich-McMillan: Changed the platform check to `PlatformMacEmulationSnpCheck` to align with the other function `PlatformMacEmulationEnable`.

Added documentation to indicate caller must ensure the following are true prior to programming the mac address:
1. Snp is initialized
2. Mac address is changable
3. Snp type is ethernet

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
Erich-McMillan: Yeah relying on IsMatch, but as @spbrogan pointed out this design is confusing. I changed the input validation to use a return, open to changing the other if statements to return as well if y'all prefer.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
Erich-McMillan: Originally intended to make unit testing this function easier since you can't mock functions within the same .c file, didn't actually need to use it. If the IsMatch function becomes more complex then using a function pointer makes stubbing the behavior much easier.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
Erich-McMillan: I usually prefer to let the lower level functions handle these cases otherwise you end up with several layers of NULL checks which can make the code slower (depending on whether compiler can optimize them away) and harder to read.

If I add a debug warn/verbose when input null to `SetSnpMacViaContext` would that address your concern?

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.h
Erich-McMillan: Sure wish VFR compiler understood `#pragma once` usually that's the argument "for" using `#ifdef`.

Maybe we'll get around to fixing the VFR compiler someday.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.h
Erich-McMillan: I believe `EfiNamedEventListen` returns `OUT VOID* Registration`, also good catch seems this wasn't being assigned (or used). Perhaps better to remove entirely?

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.h
Erich-McMillan: Prefer not to comment on test functions since those usually take the form of

```
/**
  Unit test for ValidateBlobWithXdrCertificates () API with all failed validation.

  @param[in]  Context    [Optional] An optional parameter that enables:
                         1) test-case reuse with varied parameters and
                         2) test-case re-entry for Target tests that need a
                         reboot.  This parameter is a VOID* and it is the
                         responsibility of the test author to ensure that the
                         contents are well understood by all test cases that may
                         consume it.

  @retval  UNIT_TEST_PASSED             The Unit test has completed and the test
                                        case was successful.
  @retval  UNIT_TEST_ERROR_TEST_FAILED  A test case assertion has failed.
**/
```

which will be the same for every single test case.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.h
Erich-McMillan: Generally yes, but for unit testing I've added it to ensure the entry point is callable from the test files.

MsCorePkg/MacAddressEmulationDxe/Test/EntryPointTest.c
Erich-McMillan: Yeah seems a copy paste error, thanks.

MsCorePkg/MacAddressEmulationDxe/Test/FindMatchingSnpTest.c
Erich-McMillan: I'm not following you here.

In the unit test we would want the program to crash if we dereference a null pointer since that means the test isn't fully complete yet.

Not sure what point would be to copy mBootService to a local copy here since it's just a statically assigned memory range.

MsCorePkg/MacAddressEmulationDxe/Test/MacAddressEmulationDxeHostTestCommon.h
Erich-McMillan: Yes, these are host based unit tests designed to compile only in Linux/Windows.

I'm a bit new to UEFI unit test framework, but at HP we had a custom unit test framework and at some level you do need to call the OS functions for the actual implementation.

I did copy these headers from an existing test, possible they aren't needed since the `UnitTestLib` should include them and stub the UEFI functionality internally.

MsCorePkg/MacAddressEmulationDxe/Test/MacAddressEmulationDxeHostTestMain.c
Erich-McMillan: Yes a holdover, will double check if it compiles then remove if necessary.

MsCorePkg/MacAddressEmulationDxe/Test/SimpleNetworkProtocolNotifyTest.c
Erich-McMillan: Good catch yes need to add them to the header.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
Erich-McMillan: Will add that logic back in, still applicable.

MsCorePkg/MacAddressEmulationDxe/Test/MacAddressEmulationDxeHostTestCommon.h
Erich-McMillan: Looks like those headers are required for the end-test file to include in order to compile, will leave them in.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,276 @@
...
+  DEBUG ((DEBUG_VERBOSE, "[%a]: Start\n", __FUNCTION__));
+
+  if (MatchFunction != NULL) {
+    Status = gBS->LocateHandleBuffer (ByProtocol, &gEfiSimpleNetworkProtocolGuid, NULL, &HandleCount, &SnpHandleBuffer);
+    if (EFI_ERROR (Status) && (Status != EFI_NOT_FOUND)) {
+      ASSERT_EFI_ERROR (Status);
+    }
+
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "[%a]: Unexpected error from LocateHandleBuffer. Status=%r\n", __FUNCTION__, Status));
```
joschock, line 104: consider moving this message before the assert; that way the debug message fires before the assert easing debug. from a nitpicky wording standpoint, EFI_NOT_FOUND isn't really unexpected here, if the system has no SNPs.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,276 @@
...
+{
+  EFI_STATUS  Status = EFI_NOT_STARTED;
+  EFI_TPL     OldTpl;
+
+  DEBUG ((DEBUG_VERBOSE, "[%a]: Start\n", __FUNCTION__));
+
+  if ((Snp != NULL) && (Context != NULL)) {
+    // Raising to highest TPL is necessary to avoid any network stack callbacks
+    // such as responding to network packets while the mac address is being
+    // configured in cases where the network stack has already been started.
```
joschock: Pedantry: this comment is not strictly true, since the TPL is lowered to CALLBACK for the stationaddress call. Packet handling _could_ occur during Snp->StationAddress() if the network stack has any packet handling callbacks at NOTIFY or higher.

The main reason TPL shenanigans are needed here is because StationAddress attempts to RaiseTpl to CALLBACK:
```
File: MU_BASECORE\NetworkPkg\SnpDxe\Station_address.c
215:   OldTpl = gBS->RaiseTPL (TPL_CALLBACK);
```
Which basically means that Snp->StationAddress () can only be called at CALLBACK or lower. However, this event callback function needs to run at TPL_NOTIFY, because the rest of the network stack may be running at TPL_CALLBACK. If this notify function also runs at TPL_CALLBACK, it won't be able to interrupt the main network stack to set the MAC, leaving the MAC effectively unset until after the TPL drops back to APP, which might be way after the network stack has tried to do all its operations at TPL_CALLBACK with the wrong mac. 

TLDR; the TPL stuff here is to ensure that the MAC is changed in a timely fashion relative to other network operations, it is not to ensure no interruption or packet handling during MAC change.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,276 @@
...
+    DEBUG ((DEBUG_WARN, "[%a]: SNP interface type is not Ethernet.\n", __FUNCTION__));
+    IsMatch = FALSE;
+  }
+
+  if (IsMatch && !Snp->Mode->MacAddressChangeable) {
+    DEBUG ((DEBUG_WARN, "[%a]: SNP interface does not support MAC address programming", __FUNCTION__));
+    IsMatch = FALSE;
+  }
+
+  if (IsMatch && !PlatformMacEmulationSnpCheck (SnpHandle)) {
```
joschock: Reflecting on TPL comment I made, I realized that PlatformMacEmulationSnpCheck() is gonna get called at TPL_NOTIFY during the SNP callback; might be good to add that to the function description so that implementors are aware of potential constraint there, since some protocols/operations may not be available at that level.

MsCorePkg/MacAddressEmulationDxe/Readme.md
```diff
@@ -0,0 +1,24 @@
...
+
+This driver provides UEFI support for MAC Address Emulation. There are two main
+
+## Features in this driver
+
+1. Support for updating the station address (i.e. the MAC address) for UEFI SNP
+instances that are used for PXE boot.  This ensures that the MAC address is
+emulated in preboot.
+
+2. Support for the configuration story for MAC emulation.  This includes a
```
joschock: Consider removing this feature: I don't think the full settings manager or provisioning support is in view in this implementation  - it's abstracted behind the platform library, correct?

MsCorePkg/MacAddressEmulationDxe/Readme.md
```diff
@@ -0,0 +1,24 @@
...
+code that verifies the proper provisioning of the MAC address to use for
+emulation.
+
+## Usage
+
+To use this feature, platforms must do the following:
+
+1. Provide an implementation of MacAddressEmulationPlatformLib.
+
+2. Update the platform DSC/FDF to build and include the driver:
```
joschock: the `:` seems to imply some missing example code?

MsCorePkg/MacAddressEmulationDxe/Readme.md
```diff
@@ -0,0 +1,24 @@
+# MacAddressEmulationDxe
```
joschock, line 1: Does this readme need a copyright?

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
rogurr: If it's not being used and there are currently no plans to use, can we remove it?  I'm all for keeping things as simple as possible especially since some programmers have trouble with pointers.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
rogurr: My request came from seeing people just copy/paste code when modifying later.  If this code is sending down a known error condition such as a NULL pointer, I was thinking of the lazy programmer later that might not be as thorough in checking parameters.  If this point in the code already knows it is an error, is there a benefit to sending the error condition to the next function?

MsCorePkg/MacAddressEmulationDxe/Test/FindMatchingSnpTest.c
rogurr: You answered my question, we would want it to crash if it tried to use the NULL.  Thanks.

joschock (approved):

MsCorePkg/MacAddressEmulationDxe/Readme.md
Erich-McMillan: I felt like it should, but there is a markdown format check that prevents anything but a title line for the first line of a markdown file.

Other readme.md do not have a copyright header in this repo either, probably for this reason @makubacki does the markdown formatter support copyrights in any way?

MsCorePkg/MacAddressEmulationDxe/Readme.md
Erich-McMillan: That's true, will update.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
Erich-McMillan: Will move to before assert, not sure I understand your comment about EFI_NOT_FOUND not being *un*expected; seems the code already handles EFI_NOT_FOUND gracefully?

joschock: Meaning if EFI_NOT_FOUND occurs, you won't assert, but you'll print "[FindMatchingSnp] Unexpected error from LocateHandleBuffer. Status=EFI_NOT_FOUND" at error level, which might be a bit of a red herring if EFI_NOT_FOUND is legit.

Erich-McMillan: True, but if I move the debug into the `if` then it will only happen for unexpected errors.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
Erich-McMillan: But then the two functions are coupled pretty tightly no? Additionally, this is a big reason to have unit tests perform input validation, as long as we're doing unit tests that check null inputs we should be fine.

rogurr (approved):
Thank you Erich.

makubacki:
@Erich-McMillan, I just added the latest PR template to your PR message. Can you please check the appropriate boxes?

MsCorePkg/MacAddressEmulationDxe/Readme.md
makubacki: No need for a copyright header in this file.

Erich-McMillan:
> @makubacki , I just added the latest PR template to your PR message. Can you please check the appropriate boxes?

Not sure if this "impacts functionality", but I've updated the relevant checks boxes in template.

makubacki:
> > @makubacki , I just added the latest PR template to your PR message. Can you please check the appropriate boxes?
> 
> Not sure if this "impacts functionality", but I've updated the relevant checks boxes in template.

At least right now, impacts functionality means a functional impact on firmware code (as opposed to something like a refactor). We can consider clarifying that in the instructions in the future.

Erich-McMillan:
> > > @makubacki , I just added the latest PR template to your PR message. Can you please check the appropriate boxes?
> > 
> > 
> > Not sure if this "impacts functionality", but I've updated the relevant checks boxes in template.
> 
> At least right now, impacts functionality means a functional impact on firmware code (as opposed to something like a refactor). We can consider clarifying that in the instructions in the future.

Yes I think some clarification would be good in the future.

MsCorePkg/Include/Library/MacAddressEmulationPlatformLib.h
```diff
@@ -0,0 +1,60 @@
...
+/**
+  Reports whether MacEmulation is enabled, and returns the address to emulate.
+
+  @param[out]  Address - Pointer to a buffer to receive the emulated MAC address to use for this platform.
+
+  @retval  EFI_UNSUPPORTED - Feature is not enabled/supported by the platform.
+  @retval  EFI_STATUS      - Status of other calls in the function.
+
+**/
+EFI_STATUS
```
spbrogan, line 22: all public functions should use EFIAPI to make the calling convention well defined and interoperable on all compiler tool chains.

MsCorePkg/Library/MacAddressEmulationPlatformLibNull/MacAddressEmulationPlatformLibNull.c
```diff
@@ -0,0 +1,107 @@
...
+  @retval TRUE  - Device represented by SnpHandle supports MAC emulation.
+  @retval FALSE - Device represented by SnpHandle does not support MAC emulation.
+**/
+BOOLEAN
+PlatformMacEmulationSnpCheck (
+  IN  EFI_HANDLE  SnpHandle
+  )
+{
+  /*
+        < OEM TODO >
```
spbrogan: I think maybe platform implementor would be better name.  I think we have tried to move away from OEM.  Example in Tianocore can be seen here: https://github.com/tianocore/edk2/blob/master/UefiCpuPkg/PiSmmCpuDxeSmm/MpService.c#L367

spbrogan (approved):

MsCorePkg/Include/Library/MacAddressEmulationPlatformLib.h
```diff
@@ -0,0 +1,60 @@
...
+IsMacEmulationEnabled (
```
makubacki: It seems like this function should return a `BOOLEAN` instead of a status. Or, be renamed something like `GetMacEmulationAddress()`.

MsCorePkg/Include/Library/MacAddressEmulationPlatformLib.h
```diff
@@ -0,0 +1,60 @@
...
+  IN  EFI_HANDLE  SnpHandle
+  );
+
+/**
+  Execute any platform or network controller specific code required to implement MAC-address emulation outside of a
+  call to the UEFI API EFI_SIMPLE_NETWORK_PROTOCOL::StationAddress().
+
+  @param[in]  Address - MAC Address to use for emulation
+
+  @retval  EFI_STATUS
```
makubacki, line 54: Need to add the expected status codes here

MsCorePkg/Library/MacAddressEmulationPlatformLibNull/MacAddressEmulationPlatformLibNull.inf
```diff
@@ -0,0 +1,30 @@
...
+
+[Packages]
+  MdePkg/MdePkg.dec
+  MsCorePkg/MsCorePkg.dec
+
+[LibraryClasses]
+
+[Protocols]
+
+[Guids]
```
makubacki: You can remove these sections if they're not used.

MsCorePkg/MacAddressEmulationDxe/Readme.md
```diff
@@ -0,0 +1,26 @@
+# MacAddressEmulationDxe
+
+This driver provides UEFI support for MAC Address Emulation. There are two main
```
makubacki: Incomplete sentence.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.inf
```diff
@@ -0,0 +1,50 @@
...
+#  DXE Driver for handling MAC Address Emulation.
+#
+#  Copyright (C) Microsoft Corporation.
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+[Defines]
+  INF_VERSION                    = 1.27
+  BASE_NAME                      = MacAddressEmulation
```
makubacki:
```suggestion
  BASE_NAME                      = MacAddressEmulationDxe
```

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,278 @@
...
+#include <Library/BaseMemoryLib.h>
+#include <Library/MemoryAllocationLib.h>
+#include <Library/DebugLib.h>
+#include <Library/UefiLib.h>
+#include <Library/UefiBootServicesTableLib.h>
+#include <Library/NetLib.h>
+
+#include <Library/MacAddressEmulationPlatformLib.h>
+
+/**
```
makubacki, line 21: Can you clean up the documentation format to match the normal conventions? Such as drop `@brief`.

Not a big deal for existing code but there's no reason to mix styles in new code being added.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,278 @@
...
+  @remark  typically called at TPL_NOTIFY
+**/
+BOOLEAN
+SnpSupportsMacEmuCheck (
+  IN CONST EFI_HANDLE                        SnpHandle,
+  IN CONST EFI_SIMPLE_NETWORK_PROTOCOL       *Snp,
+  IN CONST MAC_EMULATION_SNP_NOTIFY_CONTEXT  *Context
+  )
+{
+  BOOLEAN  IsMatch = FALSE;
```
makubacki: This is a function where I think a `goto` helps show the logic a little cleaner.

Feel free to ignore. Mainly a suggestion to reduce comparisons and help put a little more focus on the logic flow.

```
  if (input validation checks fail)
    return FALSE;

  if (SNP check fails)
    goto CheckFailure;

  return TRUE;

CheckFailure:
  return FALSE;
```

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,278 @@
...
+  if (EFI_ERROR (Status)) {
+    if (Status != EFI_UNSUPPORTED) {
+      DEBUG ((DEBUG_ERROR, "[%a]: Failed to determine MAC Emulated Address support. Status = %r\n", __FUNCTION__, Status));
+    }
+
+    return Status;
+  }
+
+  // Allocate and initialize context
+  Context = AllocateZeroPool (sizeof (MAC_EMULATION_SNP_NOTIFY_CONTEXT));
```
makubacki: Consider using the variable name for the size. If somone updates the type in the future, they won't have to update the type here as well.

```suggestion
  Context = AllocateZeroPool (sizeof (*Context));
```

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,278 @@
...
+  if (Context == NULL) {
+    DEBUG ((DEBUG_ERROR, "[%a]: cannot allocate notify context\n", __FUNCTION__));
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  Context->Assigned = FALSE;
+  CopyMem (&Context->EmulationAddress, &Address, sizeof (EFI_MAC_ADDRESS));
```
makubacki: Similarly here.

```suggestion
  CopyMem (&Context->EmulationAddress, &Address, sizeof (Context->EmulationAddress));
```

MsCorePkg/UnitTests/MsCorePkgHostTest.dsc
```diff
@@ -0,0 +1,48 @@
+## @file
+# Host Test DSC for the MsCore Package
+#
+# Copyright (c) Microsoft Corporation
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+################################################################################
+[Defines]
+  PLATFORM_NAME                  = MsCorePkg
```
makubacki:
```suggestion
  PLATFORM_NAME                  = MsCorePkgHostTest
```

MsCorePkg/UnitTests/MsCorePkgHostTest.dsc
```diff
@@ -0,0 +1,48 @@
...
+  PLATFORM_GUID                  = 55442A6C-2455-414D-9842-23002F5B20C7
+  PLATFORM_VERSION               = 0.1
+  DSC_SPECIFICATION              = 0x00010005
+  OUTPUT_DIRECTORY               = Build/MsCorePkg
```
makubacki: I don't think we have a firm convention here, but I think it would be nice to separate the builds from this DSC.

```suggestion
  OUTPUT_DIRECTORY               = Build/MsCorePkg/HostTest
```

makubacki (approved):
Looks good. Left some minor suggestions.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
Erich-McMillan: Okay with using multiple returns here, goto would be better if there were clean up that needed to be performed regardless of failure, but no cleanup needed in this case.

Would multiple returns be better?

makubacki: I mentioned the goto because it sets the infrastructure in place for clean up logic with roughly the same lines of code now. But, multiple returns would be perfectly fine as well.

MsCorePkg/MacAddressEmulationDxe/MacAddressEmulationDxe.c
```diff
@@ -0,0 +1,278 @@
...
+    DEBUG ((DEBUG_WARN, "[%a]: SNP interface does not support MAC address programming", __FUNCTION__));
+    return FALSE;
+  }
+
+  if (!PlatformMacEmulationSnpCheck (SnpHandle)) {
+    DEBUG ((DEBUG_WARN, "[%a]: Platform library reports not to support this SNP", __FUNCTION__));
+    return FALSE;
+  }
+
+  if (IsMatch && (Context->Assigned == TRUE)) {
```
makubacki: Looks like you can drop `IsMatch` entirely now.

Erich-McMillan: Yeah forgot to delete that. Was letting the CI do the unit tests for me (which would have caught this error) 😁

PR #116, opened 2022-11-30
Add PcdShellFvGuid to allow platforms to provide FV GUID to search shell image in

srisrid:
## Description

Add PcdShellFvGuid to allow platforms to provide FV GUID to search shell image in

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

When a FV GUID is provided, a default boot option for shell with the FV device path is created. as long as the device path exists, system is able to boot to internal UEFI shell.

If the PCD is not updated by platform and is at default value (zero guid), then all FV's are searched and shell is located to create a boot option device path.

## Integration Instructions

N/A


mikeytdisco (approved):

joschock (approved):

makubacki (approved):

PcBdsPkg/PcBdsPkg.dec
```diff
@@ -125,6 +125,9 @@
   # @Prompt FFS Name of Shell Application
   gPcBdsPkgTokenSpaceGuid.PcdShellFile|{ 0xB7, 0xD6, 0x7A, 0xC5, 0x15, 0x05, 0xA8, 0x40, 0x9D, 0x21, 0x55, 0x16, 0x52, 0x85, 0x4E, 0x37 }|VOID*|0x40000129
 
+  ## GUID of the FV to locate the shell in.
```
makubacki, line 128: I think it would be helpful to mention that this is optional here. Prevents having to look up the PCD logic in the code to figure out how it is used.

srisrid: Updated.

srisrid:
<Testing in progress>

srisrid:
@microsoft-github-policy-service agree [company="microsoft"]

srisrid:
@microsoft-github-policy-service agree company="microsoft"

PR #120, opened 2022-12-09
Add Arm64 Translation Table Parsing to Windows Paging Audit Scripts

TaylorBeebe:
## Description

Updates the python scripts and adds an html template to enable the parsing and presentation of Armv8 translation tables.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Generating Arm64 paging audits using ARM-based Surface platforms.

## Integration Instructions

N/A


UefiTestingPkg/AuditTests/PagingAudit/Windows/PagingReportGenerator.py
```diff
@@ -228,6 +232,7 @@ def main():
     parser.add_argument('-p', "--PlatformName", dest="PlatformName", help="Name of Platform.  Will show up on report", default="Test Platform")
     parser.add_argument('-t', "--type", choices=['SMM', 'DXE'], dest="Type", help="SMM or DXE Paging Report", required=True)
     parser.add_argument("--PlatformVersion", dest="PlatformVersion", help="Version of Platform.  Will show up report", default="1.0.0")
+    parser.add_argument('-a', "--architecture", choices=['Arm64', 'x86'], dest="Architecture", help="The architecture of the pages. Can be ARM64 or x86", required=True)
```
kuqin12: nit: should we really require this? maybe just default to x86?

kuqin12 (approved):

UefiTestingPkg/AuditTests/PagingAudit/Windows/PagingReportGenerator.py
TaylorBeebe: The report generation won't fail if you try to parse AArch64 pages with x86 page definitions, but the report will be garbage. I think forcing an architecture specification via command line is good for avoiding confusion.

os-d (approved):

PR #121, opened 2022-12-12
Add Arm64 Translation Table Data Collection via DxePagingAudit

TaylorBeebe:
## Description

Update the DxePagingAuditTestApp and DxePagingAuditDriver to enable the collection of Arm64 translation tables.

The interfaces remain the same. However, this update changes the behavior of DxePagingAuditTestApp to write the paging output to the volume containing the app itself instead of the EFI partition (if possible) which is much more convenient because the files don't need to be copied out of the partition for parsing and an OS image isn't necessary to audit the page table on QEMU.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

The x86 side was tested via the page table parsing scripts to ensure the output is comparable. The Arm64 translation table collection was tested on ARM Surface platforms and our QEMU-based ARM platform. The collected data was tested using the updates to the parsing scripts in a separate PR.

## Integration Instructions

N/A

### DIFF
--- a/UefiTestingPkg/AuditTests/PagingAudit/UEFI/DxePagingAuditTestApp.inf
+++ b/UefiTestingPkg/AuditTests/PagingAudit/UEFI/DxePagingAuditTestApp.inf
@@ -18,13 +18,20 @@
 
 [Sources]
   Dxe/App/DxePagingAuditTestApp.c
+  Dxe/App/DxePagingAuditTestApp.h
   PagingAuditCommon.c
   PagingAuditCommon.h
 
-
 [Sources.X64]
   X64/PagingAuditProcessor.c
+  Dxe/App/X64/DxePagingAuditTests.c
 
+[Sources.AARCH64]
+  AArch64/PagingAuditProcessor.c
+  Dxe/App/AArch64/DxePagingAuditTests.c
+  
+[Packages.AARCH64]
+  ArmPkg/ArmPkg.dec
 
 [Packages]
   MdePkg/MdePkg.dec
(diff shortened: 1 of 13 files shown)


UefiTestingPkg/AuditTests/PagingAudit/UEFI/Arm64/PagingAuditProcessor.c
```diff
@@ -0,0 +1,240 @@
...
+
+  @return the maximum support physical address bits supported.
+**/
+UINT8
+EFIAPI
+CalculateMaximumSupportAddressBits (
+  VOID
+  )
+{
+  return 36;
```
os-d: Why is this 36?

UefiTestingPkg/AuditTests/PagingAudit/UEFI/DxePagingAuditTestApp.inf
```diff
@@ -18,13 +18,20 @@
...
 [Sources]
   Dxe/App/DxePagingAuditTestApp.c
+  Dxe/App/DxePagingAuditTestApp.h
   PagingAuditCommon.c
   PagingAuditCommon.h
 
-
 [Sources.X64]
   X64/PagingAuditProcessor.c
+  Dxe/App/x86/DxePagingAuditTestsx86.c
```
os-d: nit: naming, is it more consistent to use x64? Same I guess with aarch64.

os-d (approved):

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Arm64/PagingAuditProcessor.c
TaylorBeebe: You may be confusing total width with address width. The width of the address is 36 bits because the first 12 bits are unnecessary, so the address mask is 0xFFFFFFFFF000 giving 256 terabytes of addressable memory.

TaylorBeebe: I should clarify I'm just referring to the translation table when saying that. Writing to address 0x1000 is different from writing to address 0x1001, but both use the same translation table entry.

os-d: Gotcha, thanks!

apop5 (approved):

UefiTestingPkg/AuditTests/PagingAudit/UEFI/DxePagingAuditTestApp.inf
TaylorBeebe: Renamed

PR #122, opened 2022-12-12
Adding customizable GUID for BERT extra variables support

kuqin12:
## Description

Current implement will allow extra variables with names specified under
`PcdBertEntriesVariableNames` and namespace GUID being
`gEfiHardwareErrorVariableGuid` to be collected into BERT table for OS
consumption.

However, variables under `gEfiHardwareErrorVariableGuid` has certain
restrictions that does not allow random variable names to be used and
could potentially clash with `HwErrRec####` occupied by other firmware
entities.

This change will add the flexibility for the platform to specify their own
variable GUID to be collected during boot for BERT table usage.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on QEMU virtual systems and proprietary physical devices.

## Integration Instructions

For platforms that need to support specific CPER formatted variables
coalition into the BERT table, one should override the PCD value of
`gMsWheaPkgTokenSpaceGuid.PcdBertEntriesVariableGuid` of their choices.
Otherwise, `gEfiHardwareErrorVariableGuid` will be the default value and
certain variable naming schema might restrict the specific use case.

### DIFF
--- a/MsWheaPkg/HwErrBert/HwErrBert.inf
+++ b/MsWheaPkg/HwErrBert/HwErrBert.inf
@@ -57,6 +57,7 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdAcpiDefaultCreatorId
   gEfiMdeModulePkgTokenSpaceGuid.PcdAcpiDefaultCreatorRevision
   gMsWheaPkgTokenSpaceGuid.PcdBertEntriesVariableNames
+  gMsWheaPkgTokenSpaceGuid.PcdBertEntriesVariableGuid
 
 [Guids]
   gEfiHardwareErrorVariableGuid       ## CONSUMES
--- a/MsWheaPkg/HwErrBert/HwErrorBert.c
+++ b/MsWheaPkg/HwErrBert/HwErrorBert.c
@@ -28,6 +28,7 @@ STATIC EFI_EVENT  mExitBootServicesEvent = NULL;
 STATIC EFI_EVENT  mReadyToBootEvent      = NULL;
 UINT16            mVarNameListCount      = 0;
 CHAR16            *mVarNameList          = NULL;
+EFI_GUID          *mVarGuidList          = NULL;
 
 /**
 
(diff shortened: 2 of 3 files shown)


MsWheaPkg/HwErrBert/HwErrBert.inf
```diff
@@ -1,70 +1,71 @@
...
+
+[Guids]
+  gEfiHardwareErrorVariableGuid       ## CONSUMES
+  gMsWheaReportServiceGuid            ## SOMETIMES_CONSUMES
+  gEfiFirmwareErrorSectionGuid        ## SOMETIMES_CONSUMES
+  gEfiEventNotificationTypeBootGuid   ## SOMETIMES_CONSUMES
+  gEfiEventReadyToBootGuid            ## CONSUMES             ## Event
+  gEfiEventExitBootServicesGuid       ## CONSUMES
+
+[Depex]
```
os-d: While is this whole file changed? Did line endings get updated?

os-d (approved):

MsWheaPkg/HwErrBert/HwErrBert.inf
kuqin12: I think this is done by the uncrustify tool. The line ending seems to be the change.

os-d: Just as long as it is taking the line endings we want :)

kuqin12: Yeah, I will try to remove that and run the build once more. Want to avoid that overhaul if possible.

kuqin12: Nice catch, that change was not needed and seems to be in the wrong direction. Fixed in the latest update.

apop5 (approved):

makubacki (approved):

PR #123, opened 2022-12-13
Introducing MP management driver and unit test app

kuqin12:
## Description

This change added a new driver that can power on, power off, suspend and resume
designated secondary application processors. The suspend operation supports C1,
C2 and C3 for all secondary cores.

The BSP suspend operation is also supported, but the core will be woken up from
timer interrupt.

On ARM specific implementation (which is the only implementation for now), the driver
is built on top of PSCI based MP service driver from MU_TIANO_ARM repo. The C2
and C3 will go through PSCI commands. The wakeup interrupt for AP is through SGI
from BSP.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

This change is verified using the shell-based unit test on FVP and QEMU SBSA platforms
(with needed TFA changes).

## Integration Instructions

For ARM based systems, one needs to add `ArmPkg/Drivers/ArmPsciMpServicesDxe/ArmPsciMpServicesDxe.inf`
in addition to `UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.inf` into platform
dsc and fdf files.


UefiTestingPkg/FunctionalSystemTests/MpManagement/App/MpManagementTestApp.inf
```diff
@@ -0,0 +1,44 @@
...
+  BASE_NAME                      = MpManagementTestApp
+  FILE_GUID                      = B6ED4248-1EEE-4E70-BCB2-E167E42B1230
+  MODULE_TYPE                    = UEFI_APPLICATION
+  VERSION_STRING                 = 1.0
+  ENTRY_POINT                    = MpManagementTestApp
+
+#
+# The following information is for reference only and not required by the build tools.
+#
+#  VALID_ARCHITECTURES           = IA32 X64 AARCH64
```
apop5, line 19: If this is currently only for ARM, shouldn't IA32 X64 be removed?

UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.inf
```diff
@@ -0,0 +1,82 @@
...
+  BASE_NAME                      = MpManagement
+  FILE_GUID                      = 945DEFA8-8F4A-4049-BCFC-906D2B1D9455
+  MODULE_TYPE                    = DXE_DRIVER
+  VERSION_STRING                 = 0.1
+  ENTRY_POINT                    = MpManagementEntryPoint
+
+#
+# The following information is for reference only and not required by the build tools.
+#
+#  VALID_ARCHITECTURES           = IA32 X64 AARCH64
```
apop5: If this is currently only supporting arm, should the IA32 X64 be removed for now?

UefiTestingPkg/Include/Protocol/MpManagement.h
```diff
@@ -0,0 +1,182 @@
...
+  AP_POWER_C3,
+  AP_POWER_NUM
+} AP_POWER_STATE;
+
+///
+/// Forward declaration for the MP_MANAGEMENT_PROTOCOL.
+///
+typedef struct _MP_MANAGEMENT_PROTOCOL MP_MANAGEMENT_PROTOCOL;
+
+typedef
```
apop5: function header looks like its missing.

apop5 (approved):

UefiTestingPkg/FunctionalSystemTests/MpManagement/App/MpManagementTestApp.inf
kuqin12: The test application is general as it only depends on the `MP_MANAGEMENT_PROTOCOL`. But the implementation for `MP_MANAGEMENT_PROTOCOL` is only for ARM right now. I will remove the X86 marks for the protocol DXE driver. But I think this app's `VALID_ARCHITECTURES` could stay as it is. Please let me know if you think otherwise.

UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.inf
kuqin12: Agreed, will remove. thanks!

UefiTestingPkg/Include/Protocol/MpManagement.h
kuqin12: This function should be removed. Good catch. Thanks.

kuqin12: Removed in the latest update.

UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.inf
kuqin12: Removed in the latest updated.

UefiTestingPkg/FunctionalSystemTests/MpManagement/App/MpManagementTestApp.inf
apop5: That should be fine

os-d (approved):

PR #125, opened 2022-12-15
Fixing a build break inside MP management driver

kuqin12:
## Description

There was a build break introduced during function description changes,
the input parameter and the consumed variable has a name mismatch.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change is tested on QEMU and FVP virtual platforms.

## Integration Instructions

N/A.


kuqin12:
Not sure why this is not being caught by the pipeline build.

apop5 (approved):

spbrogan (approved):

makubacki (approved):

PR #130, opened 2022-12-20
Changed FreeRamForSEC to not use passed Address since is can be Memory Address

apop5:
When FreeRamForSEC is called, the passed address can be a memory address and not CAR base.

This PR removes the assert checking if the passed address matches PcdAdvancedLoggerCarBase, and modifies the MTRR teardown logic to search for the PcdAdvancedLoggerCarBase instead of the passed address. 

## Description

While transitioning to memory, the platform under test was passing a memory address.

In InitializeDebugAgent(), Memory is allocated and the log is transitioned, but the logbuffer was being passed to FreeRamForSEC after it was transitioned to memory, causing the assert to trip. 

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Ran on emulated platform and verified that logs were still available. 

## Integration Instructions
N/A

### DIFF
--- a/AdvLoggerPkg/Library/DebugAgent/Sec/ia32/RamForSEC.c
+++ b/AdvLoggerPkg/Library/DebugAgent/Sec/ia32/RamForSEC.c
@@ -248,7 +248,6 @@ FreeRamForSEC (
   UINT64                           MtrrValidBitsMask;
   UINT32                           VariableMtrrCount;
 
-  ASSERT (Address == FixedPcdGet64 (PcdAdvancedLoggerCarBase));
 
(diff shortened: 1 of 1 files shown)


mikeytdisco (approved):

makubacki:
@apop5 what changes are you planning to make before publishing the PR?

apop5:
> @apop5 what changes are you planning to make before publishing the PR?

It looks like this is not necessary.

It looks like there is a PEIM that is overwriting the start of the temporary ram base.  I think I can fix this through modifications of PCDs.

PR #134, opened 2022-12-27
Fix C3 resume error on certain platforms

kuqin12:
# Preface

## Description

The original implementation is missing a step of populating the temporary stack pointer
during the C3 resumption process. This would impact the operations involving stack
movements (local variables, calls and returns), until the long jump occurs.

This change is adapted from the assembly `ApEntryPoint` routine from
`ArmPsciMpServicesDxe.inf` to mitigate such null-stack window during C3 resumption.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

The change was tested on Q35 and FVP virtual platforms as well as proprietary physical
platforms.

## Integration Instructions

N/A


UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.inf
```diff
@@ -75,8 +76,9 @@
   gEfiLoadedImageProtocolGuid   ## CONSUMES
   gMpManagementProtocolGuid     ## PRODUCES
 
-[Protocols]
-  gEfiTimerArchProtocolGuid     ## CONSUMES
+[Protocols.AARCH64]
+  gEfiTimerArchProtocolGuid           ## CONSUMES
+  gEfiWatchdogTimerArchProtocolGuid   ## CONSUMES
```
os-d, line 81: Does only the AARCH64 build require these?

os-d (approved):

makubacki (approved):

UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.inf
kuqin12: For now, yes. The requirement on x86 is unknown for now. We can try to coalesce them when we get there.

PR #141, opened 2023-01-09
Adding new test tool:  Driver Dependency Logging Toolset

rogurr:
## Description

This test module is intended to be added to the Core DXE dispatcher to provide an evaluation of each DXE driver loaded during boot for dependency expression declaration and protocol usage.  Notification will be provided for any driver using a protocol not covered by the depex and any depex GUID not used in a locate protocol call.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ X ] Includes documentation?

## How This Was Tested

Library data logging was tested in QEMU X64 and on an AARCH64 development platform.  Added the test library to the CoreDxe driver and verified the log data was correct for a sample of pertinent drivers.  The parsing of data portion was tested by using the output from both the boot logs plus injecting specific failures to test error handling.

## Integration Instructions

Add DriverDependencyLogging.inf to the DxeMain.inf in the platform .DSC file as a driver specific NULL library.
    MdeModulePkg/Core/Dxe/DxeMain.inf {
        <LibraryClasses>
            NULL | UefiTestingPkg/Library/DriverDependencyLogging/DriverDependencyLogging.inf
    }
See the DriverDependencyLogging.md file for details.


UefiTestingPkg/Library/DriverDependencyLogging/DispatchDataParser.py
```diff
@@ -0,0 +1,631 @@
...
+      dec = DecParser()
+      dec.ParseFile(os.path.join(Root, FileName))
+      for e in (dec.Protocols + dec.PPIs + dec.Guids):
+        Guid = str(e.guid).upper()
+        GuidName = e.name
+
+        # DataList is sorted, find the index of where the new entry should reside
+        Begin = 0
+        End = len(DataList)
+        while Begin != End:
```
spbrogan: I would suggest using pythons list search functions rather than implement your own binary search.  I would guess these 40 lines could be done in a line or two of python.  :)

rogurr: I'm definitely not an expert on Python so I could have missed a better way to do this.

I chose this search/edit loop because it has to scan a list within a list examining a single entry within the dictionary object instead of the entire object in both lists.    Most of the examples I found in places like stack overflow or GeeksForGeeks were using qualifiers such as a function or lambda key to enhance the python functions, but I also found this quick bisect loop as a suggestion on Stack Overflow to keep things simple.

So, I wrote it both ways to compare and it seemed much easier to follow this way, but I'd be glad to go back to dig if you have suggestions on where to learn better search algorithms.

UefiTestingPkg/Library/DriverDependencyLogging/DriverDependencyLogging.c
```diff
@@ -0,0 +1,423 @@
...
+#include <Library/BaseLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/MemoryAllocationLib.h>
+#include <Library/UefiLib.h>
+#include "DriverDependencyLogging.h"
+
+//
+// WARNING - This library is tightly coupled to the MdeModulePkg Core DXE driver
+//
+#include "../Core/Dxe/DxeMain.h"
```
spbrogan: Ron,

We need to talk about this.  
Relative includes especially back to DxeMain are not something we should do.   This opens up huge potential for problems.

spbrogan (changes requested):

UefiTestingPkg/Library/DriverDependencyLogging/DriverDependencyLogging.c
rogurr: I agree this is not ideal, but since the code is heavily tied to the DxeMain and is intended to extend the functionality of DxeMain only, this was the best option.

Modifying the DxeMain code to expose a private structure is off the table for an option.  And making a copy of the structure definition locally can cause worse problems since the structure has a LIST_ENTRY embedded and if offsets change, we will have all kinds of random errors when typcasting from the linked list pointer to the data pointer.

Your question does bring up a good point though, maybe it is not clear how tightly this is coupled to the DxeMain driver.  I think I do need to maybe name it something like DxeMainDepexLoggingLib that shows it is a MU_PLUS item that is extending the DxeMain driver.

PR #222, opened 2023-04-14
[REBASE && FF] MS Whea Early Store Changes

TaylorBeebe:
Update to change the checksum calculation routine and the tests cases added to the MsWheaEarlyUnitTestApp.

First commit fixes Issue: https://github.com/microsoft/mu_plus/issues/228

### DIFF
--- a/MsWheaPkg/MsWheaReport/MsWheaEarlyStorageMgr.c
+++ b/MsWheaPkg/MsWheaReport/MsWheaEarlyStorageMgr.c
@@ -219,8 +219,7 @@ header accordingly.
 STATIC
 VOID
 MsWheaESContentChangeChecksumHelper (
-  UINT16  *Buffer,
-  UINTN   Length
+  UINTN  Length
   )
 {
   UINT16                        Sum16;
@@ -229,17 +228,9 @@ MsWheaESContentChangeChecksumHelper (
   DEBUG ((DEBUG_INFO, "Calculate sum content helper...\n"));
 
   MsWheaESReadHeader (&Header);
-  Sum16 = (UINT16)(0x10000 - Header.Checksum);
-
-  // Update length field of header, update checksum accordingly
-  Sum16               = (Sum16 - (UINT16)(Header.ActiveRange & MAX_UINT16) - (UINT16)((Header.ActiveRange >> 16) & MAX_UINT16));
   Header.ActiveRange += (UINT32)Length;
-  Sum16               = (Sum16 + (UINT16)(Header.ActiveRange & MAX_UINT16) + (UINT16)((Header.ActiveRange >> 16) & MAX_UINT16));
-
-  // Calculate updated chunk only
-  Sum16           = Sum16 + CalculateSum16 (Buffer, Length);
-  Header.Checksum = (UINT16)(0x10000 - Sum16);
-
+  MsWheaESCalculateChecksum16 (&Header, &Sum16);
+  Header.Checksum = Sum16;
   MsWheaESWriteHeader (&Header);
 }
 
@@ -401,7 +392,7 @@ MsWheaESV0InfoStore (
     goto Cleanup;
   }
 
-  MsWheaESContentChangeChecksumHelper ((UINT16 *)&WheaV0, sizeof (MS_WHEA_EARLY_STORAGE_ENTRY_V0));
+  MsWheaESContentChangeChecksumHelper (sizeof (MS_WHEA_EARLY_STORAGE_ENTRY_V0));
 
 Cleanup:
   return Status;
--- a/MsWheaPkg/Test/UnitTests/MsWheaEarlyStorageUnitTestApp/MsWheaEarlyUnitTestApp.c
+++ b/MsWheaPkg/Test/UnitTests/MsWheaEarlyStorageUnitTestApp/MsWheaEarlyUnitTestApp.c
@@ -85,8 +85,7 @@ MsWheaESWriteHeader (
 
 VOID
 MsWheaESContentChangeChecksumHelper (
-  UINT16  *Buffer,
-  UINTN   Length
+  UINTN  Length
   );
 
 VOID
@@ -331,7 +330,7 @@ MsWheaESContentUpdateTest (
   UINT8  Data;
 
   MsWheaESWriteData (TestDataArray, sizeof (TestDataArray), 0);
-  MsWheaESContentChangeChecksumHelper (TestDataArray, sizeof (TestDataArray));
+  MsWheaESContentChangeChecksumHelper (sizeof (TestDataArray));
 
   MsWheaESReadHeader (&UnitTestHeader);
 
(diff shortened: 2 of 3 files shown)


MsWheaPkg/MsWheaReport/MsWheaEarlyStorageMgr.c
```diff
@@ -229,17 +228,9 @@ MsWheaESContentChangeChecksumHelper (
...
-  Sum16               = (Sum16 - (UINT16)(Header.ActiveRange & MAX_UINT16) - (UINT16)((Header.ActiveRange >> 16) & MAX_UINT16));
   Header.ActiveRange += (UINT32)Length;
-  Sum16               = (Sum16 + (UINT16)(Header.ActiveRange & MAX_UINT16) + (UINT16)((Header.ActiveRange >> 16) & MAX_UINT16));
-
-  // Calculate updated chunk only
-  Sum16           = Sum16 + CalculateSum16 (Buffer, Length);
-  Header.Checksum = (UINT16)(0x10000 - Sum16);
-
+  MsWheaESCalculateChecksum16 (&Header, &Sum16);
+  Header.Checksum = Sum16;
```
kuqin12, line 233: nit: Do you need to 0 this field?

kuqin12 (approved):

MsWheaPkg/MsWheaReport/MsWheaEarlyStorageMgr.c
TaylorBeebe: The MsWheaESCalculateChecksum16() zeros the checksum before doing the calculation. Is that what you mean?

kuqin12: yep, thanks!

MsWheaPkg/MsWheaReport/MsWheaEarlyStorageMgr.c
makubacki, line 1: @TaylorBeebe - Can you please add a label to categorize this change (e.g. "bug", "enhancement", etc.). My understanding is that it is an enhancement to improve robustness.

makubacki: It would also be nice to have an issue linked related to the motivation for the change in MsWheaEarlyStorageMgr.c:

```
The checksum calculation for the MS Whea Early Storage could be invalid
if the size of the early storage was not set properly. To make the
calculation more robust, the checksum calculation is updated to read
the actual bytes of the early store instead of relying on an input
buffer which may not have been properly written to the early store.
```

makubacki: Also, please expand on "the size of the early storage was not set properly". If someone is debugging an issue related to this, they can see your change and make an immediate connection to something they did wrong.

MsWheaPkg/Test/UnitTests/MsWheaEarlyStorageUnitTestApp/MsWheaEarlyUnitTestApp.c
```diff
@@ -601,15 +601,41 @@ MsWheaEarlyUnitTestAppEntryPoint (
...
-    Misc,
-    "Free ES slot find",
-    "MsWhea.Miscellaneous.MsWheaESFindSlotTest",
-    MsWheaESFindSlotTest,
-    MsWheaESVerify,
-    MsWheaESCleanUp,
-    NULL
-    );
+  // Only run the following tests if the early storage capacity is large enough
+  if (MsWheaEarlyStorageGetMaxSize () >= (sizeof (MS_WHEA_EARLY_STORAGE_HEADER) + (sizeof (MS_WHEA_EARLY_STORAGE_ENTRY_V0)))) {
```
makubacki: nit: Is there a reason you surrounded one `sizeof()` invocation with `()` but not the other? It looks a little strange unbalanced.

makubacki: Are you going by the actual value of `PcdMsWheaReportEarlyStorageCapacity` as assigned in the package DSC building the host-based unit test?

If so, I think that should be mocked and positive and negative cases tested.

TaylorBeebe: What do you mean by "mocked and positive and negative cases tested?"

makubacki: Write mock instances of `MsWheaEarlyStorageGetMaxSize()`.

- **Mock 1**: Return a size less than: `(sizeof (MS_WHEA_EARLY_STORAGE_HEADER) + (sizeof (MS_WHEA_EARLY_STORAGE_ENTRY_V0))`
- **Mock 2**: Return a size equal to: `(sizeof (MS_WHEA_EARLY_STORAGE_HEADER) + (sizeof (MS_WHEA_EARLY_STORAGE_ENTRY_V0))`
- **Mock 3**: Return a size greater than: `(sizeof (MS_WHEA_EARLY_STORAGE_HEADER) + (sizeof (MS_WHEA_EARLY_STORAGE_ENTRY_V0))`

Verify functions dependent on `MsWheaEarlyStorageGetMaxSize()` behave as expected given the value returned from `MsWheaEarlyStorageGetMaxSize()`.

TaylorBeebe: I think it's best to just log an error if the early storage region is not large enough to store at least one ES entry. The test which requires two entry slots is still blocked behind a check against the ES capacity but will not log an error if it is too small.

TaylorBeebe: Just a copy-paste artifact

MsWheaPkg/MsWheaReport/MsWheaEarlyStorageMgr.c
TaylorBeebe: I've added a commit to this patch series which checks the sanity of the ES capacity PCD and will file an issue for it

MsWheaPkg/Test/UnitTests/MsWheaEarlyStorageUnitTestApp/MsWheaEarlyUnitTestApp.c
makubacki: Sounds reasonable

makubacki (approved):

MsWheaPkg/MsWheaReport/MsWheaEarlyStorageMgr.c
TaylorBeebe: Added Issue: https://github.com/microsoft/mu_plus/issues/228

TaylorBeebe: Also, "the size of the early storage was not set properly" was updated to be more specific

PR #231, opened 2023-04-28
[REBASE && FF] MemoryProtectionTestApp Updates

TaylorBeebe:
1. Add the ability to use the Memory Attribute Protocol in MemoryProtectionTestApp
2. Update the MemoryProtectionTestApp.c file to add comments and function headers, and remove unused functions and logic.

### DIFF
--- a/UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
+++ b/UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
@@ -52,15 +52,15 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #define DUMMY_FUNCTION_FOR_CODE_SELF_TEST_GENERIC_SIZE  512
 #define ALIGN_ADDRESS(Address)  (((Address) / EFI_PAGE_SIZE) * EFI_PAGE_SIZE)
 
-VOID                                     *mPiSmmCommonCommBufferAddress = NULL;
 UINTN                                    mPiSmmCommonCommBufferSize;
-EFI_CPU_ARCH_PROTOCOL                    *mCpu = NULL;
 MM_MEMORY_PROTECTION_SETTINGS            mMmMps;
 DXE_MEMORY_PROTECTION_SETTINGS           mDxeMps;
-MEMORY_PROTECTION_NONSTOP_MODE_PROTOCOL  *mNonstopModeProtocol      = NULL;
-MEMORY_PROTECTION_DEBUG_PROTOCOL         *mMemoryProtectionProtocol = NULL;
-EFI_MEMORY_ATTRIBUTE_PROTOCOL            *mMemoryAttributeProtocol  = NULL;
-CPU_MP_DEBUG_PROTOCOL                    *mCpuMpDebugProtocol       = NULL;
+VOID                                     *mPiSmmCommonCommBufferAddress = NULL;
+EFI_CPU_ARCH_PROTOCOL                    *mCpu                          = NULL;
+MEMORY_PROTECTION_NONSTOP_MODE_PROTOCOL  *mNonstopModeProtocol          = NULL;
+MEMORY_PROTECTION_DEBUG_PROTOCOL         *mMemoryProtectionProtocol     = NULL;
+EFI_MEMORY_ATTRIBUTE_PROTOCOL            *mMemoryAttributeProtocol      = NULL;
+CPU_MP_DEBUG_PROTOCOL                    *mCpuMpDebugProtocol           = NULL;
 
 /// ================================================================================================
 /// ================================================================================================
@@ -297,11 +297,11 @@ InterruptHandler (
   IN EFI_SYSTEM_CONTEXT  SystemContext
   )
 {
-  // We need to avoid using runtime services to reset the system due because doing so will raise the TPL level
-  // when it is already on TPL_HIGH. HwResetSystemLib is used here instead to perform a bare-metal reset and sidestep
-  // this issue.
+  // Avoid using runtime services to reset the system because doing so will raise the TPL level
+  // when it is already on TPL_HIGH. HwResetSystemLib is used here instead to perform a bare-metal reset
+  // and sidestep this issue.
   ResetWarm ();
-} // InterruptHandler()
+}
 
 /**
   This helper function returns EFI_SUCCESS if the Nonstop protocol is installed.
@@ -315,20 +315,11 @@ GetNonstopProtocol (
   VOID
   )
 {
-  if (mNonstopModeProtocol == NULL) {
-    if (EFI_ERROR (
-          gBS->LocateProtocol (
-                 &gMemoryProtectionNonstopModeProtocolGuid,
-                 NULL,
-                 (VOID **)&mNonstopModeProtocol
-                 )
-          ))
-    {
-      return EFI_NOT_FOUND;
-    }
+  if (mNonstopModeProtocol != NULL) {
+    return EFI_SUCCESS;
   }
 
-  return EFI_SUCCESS;
+  return gBS->LocateProtocol (&gMemoryProtectionNonstopModeProtocolGuid, NULL, (VOID **)&mNonstopModeProtocol);
 }
 
 /**
@@ -345,18 +336,7 @@ CheckMemoryProtectionExceptionHandlerInstallation (
 {
   VOID  *DummyProtocol = NULL;
 
-  if (EFI_ERROR (
-        gBS->LocateProtocol (
-               &gMemoryProtectionExceptionHandlerGuid,
-               NULL,
-               (VOID **)&DummyProtocol
-               )
-        ))
-  {
-    return EFI_NOT_FOUND;
-  }
-
-  return EFI_SUCCESS;
+  return gBS->LocateProtocol (&gMemoryProtectionExceptionHandlerGuid, NULL, (VOID **)&DummyProtocol);
 }
 
 /**
@@ -406,9 +386,7 @@ SmmMemoryProtectionsDxeToSmmCommunicate (
     return EFI_ABORTED;
   }
 
-  //
-  // First, let's zero the comm buffer. Couldn't hurt.
-  //
+  // Zero the comm buffer
   CommHeader     = (EFI_SMM_COMMUNICATE_HEADER *)mPiSmmCommonCommBufferAddress;
   CommBufferSize = sizeof (MEMORY_PROTECTION_TEST_COMM_BUFFER) + OFFSET_OF (EFI_SMM_COMMUNICATE_HEADER, Data);
   if (CommBufferSize > mPiSmmCommonCommBufferSize) {
@@ -418,29 +396,22 @@ SmmMemoryProtectionsDxeToSmmCommunicate (
 
   ZeroMem (CommHeader, CommBufferSize);
 
-  //
-  // Update some parameters.
-  //
-  // SMM Communication Parameters
+  // Update the SMM communication parameters
   CopyGuid (&CommHeader->HeaderGuid, &gMemoryProtectionTestSmiHandlerGuid);
   CommHeader->MessageLength = sizeof (MEMORY_PROTECTION_TEST_COMM_BUFFER);
 
-  // Parameters Specific to this Implementation
+  // Update parameters Specific to this implementation
   VerificationCommBuffer           = (MEMORY_PROTECTION_TEST_COMM_BUFFER *)CommHeader->Data;
   VerificationCommBuffer->Function = RequestedFunction;
   VerificationCommBuffer->Status   = EFI_NOT_FOUND;
   CopyMem (&VerificationCommBuffer->Context, Context, sizeof (MEMORY_PROTECTION_TEST_CONTEXT));
 
-  //
-  // Locate the protocol, if not done yet.
-  //
+  // Locate the protocol if necessary
   if (!SmmCommunication) {
     Status = gBS->LocateProtocol (&gEfiSmmCommunicationProtocolGuid, NULL, (VOID **)&SmmCommunication);
   }
 
-  //
-  // Signal SMM.
-  //
+  // Signal MM
   if (!EFI_ERROR (Status)) {
     Status = SmmCommunication->Communicate (
                                  SmmCommunication,
@@ -451,8 +422,11 @@ SmmMemoryProtectionsDxeToSmmCommunicate (
   }
 
   return VerificationCommBuffer->Status;
-} // SmmMemoryProtectionsDxeToSmmCommunicate()
+}
 
+/**
+  Locates and stores address of comm buffer
+**/
 STATIC
 VOID
 LocateSmmCommonCommBuffer (
@@ -470,8 +444,7 @@ LocateSmmCommonCommBuffer (
       return;
     }
 
-    // We only need a region large enough to hold a MEMORY_PROTECTION_TEST_COMM_BUFFER,
-    // so this shouldn't be too hard.
+    // Only need a region large enough to hold a MEMORY_PROTECTION_TEST_COMM_BUFFER
     BufferSize       = 0;
     SmmCommMemRegion = (EFI_MEMORY_DESCRIPTOR *)(PiSmmCommunicationRegionTable + 1);
     for (Index = 0; Index < PiSmmCommunicationRegionTable->NumberOfEntries; Index++) {
@@ -492,12 +465,21 @@ LocateSmmCommonCommBuffer (
     mPiSmmCommonCommBufferAddress = (VOID *)(UINTN)SmmCommMemRegion->PhysicalStart;
     mPiSmmCommonCommBufferSize    = BufferSize;
   }
-} // LocateSmmCommonCommBuffer()
+}
 
+/**
+  The recursion loop for testing stack overflow protection. This function will
+  recurse until it overflows the stack at which point it's expected that a switch
+  stack is used and an interrupt is generated.
+
+  @param[in]  Count   The current recursion depth.
+
+  @retval             The sum of Count and the return value of the next recursive call.
+**/
 STATIC
 UINT64
 Recursion (
-  UINT64  Count
+  IN UINT64  Count
   )
 {
   UINT64            Sum            = 0;
@@ -514,10 +496,18 @@ Recursion (
   return Sum + Count;
 }
 
+/**
+  A recursive stack overflow function which at every recursion level checks if the interrupt handler
+  has signaled that it ran and cleared the faulting region at which point we unwind the recursion.
+
+  @param[in]  Count   The current recursion depth.
+
+  @retval             The sum of Count and the return value of the next recursive call.
+**/
 STATIC
 UINT64
 RecursionDynamic (
-  UINT64  Count
+  IN UINT64  Count
   )
 {
   UINT64  Sum = 0;
@@ -531,6 +521,15 @@ RecursionDynamic (
   return Sum + Count;
 }
 
+/**
+  Tests the pool guards by allocating a pool and then writing to the guard page. If the testing
+  method is via reset, this function is expected to fault and reset the system. If the testing
+  method is via exception clearing, it's expected that this function will return and the value
+  of ExPersistGetIgnoreNextPageFault() will be FALSE.
+
+  @param[in]  ptr               The pointer to the pool.
+  @param[in]  AllocationSize    The size of the pool.
+**/
 VOID
 PoolTest (
   IN UINT64  *ptr,
@@ -541,58 +540,64 @@ PoolTest (
 
   DEBUG ((DEBUG_INFO, "%a - Allocated pool at 0x%p\n", __FUNCTION__, ptr));
 
-  //
-  // Check if guard page is going to be at the head or tail.
-  //
+  // Check if guard page is going to be at the head or tail
   if (mDxeMps.HeapGuardPolicy.Fields.Direction == HEAP_GUARD_ALIGNED_TO_TAIL) {
-    //
-    // Get to the beginning of the page the pool tail is on.
-    //
+    // Get to the beginning of the page containing the pool
     ptrLoc = (UINT64 *)(((UINTN)ptr) + (UINTN)AllocationSize);
     ptrLoc = ALIGN_POINTER (ptrLoc, 0x1000);
 
-    //
-    // The guard page will be on the next page.
-    //
+    // The guard page will be on the next page
     ptr = (UINT64 *)(((UINTN)ptr) + 0x1000);
   } else {
-    //
-    // Get to the beginning of the page the pool head is on.
-    //
+    // Get to the beginning of the page containing the pool
     ptrLoc = ALIGN_POINTER (ptr, 0x1000);
 
-    //
     // The guard page will be immediately preceding the page the pool starts on.
-    //
     ptrLoc = (UINT64 *)(((UINTN)ptrLoc) - 0x1);
   }
 
   DEBUG ((DEBUG_ERROR, "%a - Writing to 0x%p\n", __FUNCTION__, ptrLoc));
   *ptrLoc = 1;
-} // PoolTest()
+}
 
+/**
+  Test the head guard page of the input pointer by writing to the page immediately preceding it. The input
+  pointer is expected to be page aligned so subtracting 1 from will get the page immediately preceding it.
+
+  @param[in]  ptr   The pointer to the page expected to have a guard page immediately preceding it.
+**/
 VOID
 HeadPageTest (
   IN UINT64  *ptr
   )
 {
-  // Hit the head guard page
+  // Write to the head guard page
   ptr = (UINT64 *)(((UINTN)ptr) - 0x1);
   DEBUG ((DEBUG_ERROR, "%a - Writing to 0x%p\n", __FUNCTION__, ptr));
   *ptr = 1;
-} // HeadPageTest()
+}
 
+/**
+  Test the tail guard page of the input pointer by writing to the page immediately following it. The input
+  pointer can be anywhere within the page with a guard page immediately following it.
+
+  @param[in]  ptr   The pointer to the page expected to have a guard page immediately following it.
+**/
 VOID
 TailPageTest (
   IN UINT64  *ptr
   )
 {
-  // Hit the tail guard page
+  // Write to the tail guard page
   ptr = (UINT64 *)(((UINTN)ptr) + 0x1000);
   DEBUG ((DEBUG_ERROR, "%a - Writing to 0x%p\n", __FUNCTION__, ptr));
   *ptr = 1;
-} // TailPageTest()
+}
 
+/**
+  This dummy function definition is used to test no-execute protection on allocated buffers
+  and the stack.
+**/
 typedef
 VOID
 (*DUMMY_VOID_FUNCTION_FOR_DATA_TEST)(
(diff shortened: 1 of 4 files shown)


TaylorBeebe:
Will resolve https://github.com/microsoft/mu_tiano_platforms/issues/514

makubacki:
Thanks for resolving issue 514. I'll come back to this early next week.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -41,10 +43,14 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include "../MemoryProtectionTestCommon.h"
 #include "UefiHardwareNxProtectionStub.h"
 
-#define UNIT_TEST_APP_NAME     "Heap Guard Test"
-#define UNIT_TEST_APP_VERSION  "0.5"
+#define UNIT_TEST_APP_NAME                 "Memory Protection Test"
+#define UNIT_TEST_APP_VERSION              "2.0"
+#define UNIT_TEST_WARM_RESET_STRING        L"--WarmReset"
```
makubacki: Unless there's a specific reason to mention the reset type, at the app parameter level I'd simply call this reset and let parameter documentation describe that a warm reset is currently performed.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -618,7 +644,10 @@ ImageProtectionPreReq (
   IN UNIT_TEST_CONTEXT  Context
   )
 {
-  if (mDxeMps.ImageProtectionPolicy.Fields.ProtectImageFromFv || mDxeMps.ImageProtectionPolicy.Fields.ProtectImageFromUnknown) {
+  if (((mDxeMps.ImageProtectionPolicy.Fields.ProtectImageFromFv != 0) || (mDxeMps.ImageProtectionPolicy.Fields.ProtectImageFromUnknown != 0)) &&
+      ((mMemoryAttributeProtocol != NULL) || !EFI_ERROR (PopulateMemoryAttributeProtocol ())) &&
```
makubacki: Can this be simplified by dropping the global variable checks and let them be checked in the functions - `mMemoryAttributeProtocol` in `PopulateMemoryAttributeProtocol` and `mMemoryProtectionProtocol` in `PopulateMemoryProtectionDebugProtocol`?

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -861,14 +890,28 @@ UefiPageGuard (
...
   EFI_PHYSICAL_ADDRESS            ptr;
   EFI_STATUS                      Status;
+  UINT64                          Attributes;
 
   DEBUG ((DEBUG_INFO, "%a - Testing Type: %a\n", __FUNCTION__, MEMORY_TYPES[MemoryProtectionContext.TargetMemoryType]));
 
-  if (MemoryProtectionContext.DynamicActive) {
-    if (mNonstopModeProtocol == NULL) {
-      UT_ASSERT_NOT_EFI_ERROR (GetNonstopProtocol ());
+  if (MemoryProtectionContext.TestingMethod == MemoryProtectionTestMemoryAttributeProtocol) {
```
makubacki, line 985: This function is getting long. Can you split up some of the conditional blocks into separate functions that get called from here?

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -897,59 +940,57 @@ UefiPageGuard (
     }
 
     UT_ASSERT_NOT_EFI_ERROR (mNonstopModeProtocol->ResetPageAttributes ());
+  } else {
```
makubacki: Since `MemoryProtectionContext.TestingMethod` is an enum which is an integer, it would help more strictly enforce correctness of the value to explicitly check for `MemoryProtectionTestWarmReset` here and then assert if any other value is present.

TaylorBeebe: UT_LOG_ERROR ("Invalid testing method specified: %d\n", MemoryProtectionContext.TestingMethod);
return UNIT_TEST_ERROR_TEST_FAILED;

added if the test method is unexpected. Explicit check for MemoryProtectionTestReset added

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: Changed from WarmReset to Reset

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: Sure, removed

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: I'd prefer not to split things up unless this is a sticking point. I split up the entry point function because that was a few hundred lines. This one is only 100 which seems pretty normal to me.

makubacki: We decided upon a 60 line max ([contributing.md](https://github.com/microsoft/mu_plus/blob/release/202208/CONTRIBUTING.md#implementation-limits)) but that is almost never enforced.

It's not so much the total length (at this point in time) but the fact that the function is primarily composed of exclusive blocks of functionality. It would be easier to follow and extend if broken into separate functions that perform separate functional repsonsibilities. But, it's more of a nice to have so feel free to pass.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -897,59 +940,60 @@ UefiPageGuard (
...
+    if (MemoryProtectionContext.TestProgress < 2) {
+      //
+      // Context.TestProgress indicates progress within this specific test.
+      // 0 - Just started.
+      // 1 - Completed head guard test.
+      // 2 - Completed tail guard test.
+      //
+      // We need to indicate we are working on the next part of the test and save our progress.
+      //
+      MemoryProtectionContext.TestProgress++;
```
makubacki, line 1073: Just curious, why is the progress update move here? Is it expected that if the gBS->AllocatePages() call fails, MemoryProtectionContext.TestProgress would be 1?

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -964,17 +1008,15 @@ UefiPoolGuard (
...
   DEBUG ((DEBUG_INFO, "%a - Testing Type: %a\n", __FUNCTION__, MEMORY_TYPES[MemoryProtectionContext.TargetMemoryType]));
 
-  if (MemoryProtectionContext.DynamicActive) {
-    if (mNonstopModeProtocol == NULL) {
-      UT_ASSERT_NOT_EFI_ERROR (GetNonstopProtocol ());
-    }
+  if (MemoryProtectionContext.TestingMethod == MemoryProtectionTestMemoryAttributeProtocol) {
+    UT_ASSERT_NOT_NULL (mMemoryAttributeProtocol);
 
     for (Index = 0; Index < NUM_POOL_SIZES; Index++) {
```
makubacki: Was there a reason to define `NUM_POOL_SIZES` instead of using `ARRAY_SIZE()`? Array size, would help keep array length and the max index consistent.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -996,51 +1038,88 @@ UefiPoolGuard (
...
+      SetBootNextDevice ();
+
+      //
+      // Memory type rHardwareNxProtections to the bitmask for the PcdHeapGuardPoolType,
+      // we need to RShift 1 to get it to reflect the correct EFI_MEMORY_TYPE.
+      //
+      Status = gBS->AllocatePool ((EFI_MEMORY_TYPE)MemoryProtectionContext.TargetMemoryType, AllocationSize, (VOID **)&ptr);
+
+      if (EFI_ERROR (Status)) {
+        UT_LOG_WARNING ("Memory allocation failed for type %a of size %x - %r\n", MEMORY_TYPES[MemoryProtectionContext.TargetMemoryType], AllocationSize, Status);
```
makubacki: It seems to be pre-existing throughout the file but it's nice to prefix hex values with `0x` when printing.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
makubacki: Looking through the file more, you'd have to do this in most functions exercising a memory protection. There's not a need for that level of change to do this.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: The structure of this looks the same as it did back in 2018. I think it works as written

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: NUM_POOL_SIZES has been used in the test since 2018 when Bret originally wrote it but I agree using array size also works

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: I can turn this into a rebase PR and change that print along with the ARRAY_SIZE change above if you'd like

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -996,51 +1038,88 @@ UefiPoolGuard (
...
-    // We need to indicate we are working on the next part of the test and save our progress.
-    //
-    AllocationSize = mPoolSizeTable[MemoryProtectionContext.TestProgress];
-    MemoryProtectionContext.TestProgress++;
-    SaveFrameworkState (&MemoryProtectionContext, sizeof (MEMORY_PROTECTION_TEST_CONTEXT));
-    SetBootNextDevice ();
+      //
+      // Memory type rHardwareNxProtections to the bitmask for the PcdHeapGuardPoolType,
+      // we need to RShift 1 to get it to reflect the correct EFI_MEMORY_TYPE.
+      //
```
makubacki: I'm not quite making the connection between the relevance of this comment and the code. Is it relevant to the code here?

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
makubacki: I saw that too; I'm not requesting you to change it. For some of these comments, your feedback based on your experience in the code is helpful to determine if we should change it at some point.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
makubacki: If it's not too much trouble. I think that would be nice to have since we want to particularly expand usage of this test.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -1052,46 +1131,90 @@ UefiCpuStackGuard (
...
+      if (CompareGuid (&gEfiHobMemoryAllocStackGuid, &MemoryHob->AllocDescriptor.Name)) {
+        StackBase = MemoryHob->AllocDescriptor.MemoryBaseAddress;
+        UT_ASSERT_EQUAL (StackBase & EFI_PAGE_MASK, 0);
+        break;
+      }
+
+      Hob.Raw = GET_NEXT_HOB (Hob);
     }
 
+    UT_ASSERT_NOT_EQUAL (StackBase, 0);
```
makubacki, line 1347: `UT_ASSERT_EQUAL (StackBase & EFI_PAGE_MASK, 0);` is inside the HOB search loop. Should this line (`UT_ASSERT_NOT_EQUAL (StackBase, 0);`) precede that? Then, you'd exit on a non-zero, page-aligned stack into the memory attributes check.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -1833,6 +1966,88 @@ AddSmmPageTest (
   }
 } // AddSmmPageTest()
 
+STATIC
+EFI_STATUS
+DetermineTestMethod (
+  IN  MEMORY_PROTECTION_TESTING_METHOD  PreferredTestingMethod,
+  OUT MEMORY_PROTECTION_TESTING_METHOD  *TestingMethod
+  )
```
makubacki, line 2433: Function description block would be nice to have.

makubacki: The `MemoryProtectionTestMax` fall through logic for determining `DeterminedTestingMethod` would particularly be helpful to summarize.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -1833,6 +1966,88 @@ AddSmmPageTest (
...
+  }
+
+  if ((PreferredTestingMethod != MemoryProtectionTestMax) && (PreferredTestingMethod != DeterminedTestingMethod)) {
+    DEBUG ((DEBUG_ERROR, "Could not use desired testing method.\n"));
+    return EFI_INVALID_PARAMETER;
+  }
+
+  switch (DeterminedTestingMethod) {
+    case MemoryProtectionTestReset:
+      DEBUG ((DEBUG_INFO, "Testing with warm reset after each protection violation.\n"));
```
makubacki:
```suggestion
      DEBUG ((DEBUG_INFO, "Testing with a reset after each protection violation.\n"));
```

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Readme.md
```diff
@@ -7,15 +7,46 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
+on EfiBootServicesData, the test will allocate a buffer of that type and check the
+attributes of that buffer either by attempting to execute code within it or by fetching
+the attributes using the memory attribute protocol. The test has a few methods of running.
 
-It consists of:
+## Testing using Warm Resets
 
-- An SMM driver
-- A Shell-based test app
+This method is the slowest, reaching upwards of 45 minutes with a srict protection policy.
```
makubacki:
```suggestion
This method is the slowest, reaching upwards of 45 minutes with a strict protection policy.
```

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Readme.md
```diff
@@ -7,15 +7,46 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 ## About This Test
```
makubacki, line 8: For the times given in this section, can you briefly describe the environment you're referencing? For example, "Estimated run times in this section are based on a QEMU DEBUG build".

makubacki (approved):

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
makubacki: I see that now. The diff was not doing a great job lining things up. If the reset logic hasn't changed, then I won't look any further in this PR. I'll mark this resolved.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: I initialize StackBase to zero and use this line to ensure that the HOB search routine found the stack.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: I'll tack on an extra commit to this, convert it to a REBASE && FF, and add function headers to the whole test

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: Removed NUM_POOL_SIZES

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: Added 0x

TaylorBeebe: This print was removed in the restructure

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: That comment was outdated - removed

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: Added comments to the function in the second commit of this REBASE && FF

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: Fixed

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Readme.md
TaylorBeebe: Fixed

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Readme.md
TaylorBeebe: Added a note that the benchmark is Q35 debug build

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -271,11 +297,11 @@ InterruptHandler (
   IN EFI_SYSTEM_CONTEXT  SystemContext
   )
 {
-  // We need to avoid using runtime services to reset the system due because doing so will raise the TPL level
-  // when it is already on TPL_HIGH. HwResetSystemLib is used here instead to perform a bare-metal reset and sidestep
-  // this issue.
+  // Avoid using runtime services to reset the system due because doing so will raise the TPL level
```
makubacki:
```suggestion
  // Avoid using runtime services to reset the system because doing so will raise the TPL level

```

makubacki:
Thanks for adding documentation to the file, I think it greatly improves readability.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: Fixed in commit amend

apop5 (approved):

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Readme.md
```diff
@@ -7,15 +7,47 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
+the attributes using the memory attribute protocol. The test has a few methods of running.
 
-It consists of:
+## Testing using Resets
 
-- An SMM driver
-- A Shell-based test app
+This method is the slowest, reaching upwards of 45 minutes with a strict protection policy
+(benchmarked on a Q35 DEBUG build). The test will intentionally cause faults for each
+protection policy and use a custom installed fault handler for IA32_PAGE_FAULT exceptions
```
os-d, line 21: I'm assuming `IA32_PAGE_FAULT` is not what the fault handler is on ARM64?

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
```diff
@@ -852,330 +941,666 @@ SmmNullPointerPreReq (
 /// ================================================================================================
 /// ================================================================================================
 
+/**
+  This test case checks that page guards are present for the target memory type within Context.
+
+  The test can be run in 3 ways:
+  1. Using the Memory Attribute Protocol: This version of the test will allocate a page of the target
+     memory type and check that the page preceding and succeeding the allocated page have the EFI_MEMORY_RP
```
os-d, line 949: Ard said that Arm's CpuDxe doesn't have a way to set `EFI_MEMORY_RP`, is this something we are able to set on the ARM64 side?

os-d (approved):

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.c
TaylorBeebe: The access flag is EFI_MEMORY_RP on the ARM side

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Readme.md
TaylorBeebe: The original reset version of the test was only written with x86 in mind, so we don't install a synchronous exception handler for ARM platforms. That would be a future work item, though I'll be very surprised if anyone ever runs the reset version of the test again.

spbrogan: @TaylorBeebe - your comment confuses me.  You say that there is no handler for "ARM platforms" but your comment here says that it can be run on ARM64 platforms.  Does the default synchronous handler just reset the system and thus it works?   For x86 architectures don't we set a flag to indicate we want reboot instead of deadloop?

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Readme.md
```diff
@@ -7,15 +7,47 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
+fault and allow the test to continue. Project Mu currently only supports this method of
+running the test on x86 based platforms.
+
+## Testing by using the Memory Attribute Protocol
+
+This method takes less than 5 seconds to run. Instead of intentionally causing faults,
+this method checks the attributes of the region which should fault against the expected
+protection attributes. For example, instead of intentionally causing a stack overflow to
+check the stack guard, the test will find the stack base and check that the attributes
+are EFI_MEMORY_RP using the Memory Attribute Protocol. This method can be run on either
```
spbrogan: should this comment be updated to indicate that the page beyond the allocated stack space is marked RP.  I don't think you mean that the stack is marked RP?

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Readme.md
TaylorBeebe: It can be run on ARM64 platforms using the memory attribute protocol. Currently, the test doesn't install a synchronous handler for running the reset method on ARM platforms. It's possible, though.

I don't know of any flag on x86 architectures that indicates we should do a reboot instead of deadloop. We've accomplished the reboot by installing our own interrupt handler within this test, and also the memory protection exception handler used on Mu based platforms does a reboot instead of deadloop.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Readme.md
TaylorBeebe: The stack base published in the HOB is the guard page and the next page is the top of the stack. Do you think the way it's phrased is unclear?

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Readme.md
makubacki: @TaylorBeebe if I understand correctly, since this is the reset section, perhaps you should replace _"This method can be run on either X64 or ARM64 platforms."_ with _"This method can be run as implemented on X64. Currently, the test doesn't install a synchronous handler for initiating a reset on ARM platforms."_

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Readme.md
makubacki: I suggest being explicit:

For example, instead of intentionally causing a stack overflow to
check the stack guard, the test will find the stack base and check that the attributes
are EFI_MEMORY_RP using the Memory Attribute Protocol **(since the page at the stack base is a guard page marked EFI_MEMORY_RP)**.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Readme.md
TaylorBeebe: My mistake, the final sentence of this section shouldn't list support for ARM platforms. I'll replace it with Michael's suggestion.

PR #241, opened 2023-05-19
MsCorePkg: Add LoadOptionVariablePolicyDxe

makubacki:
## Description

This driver sets UEFI variable policy for Boot Manager load options that
may not be supported by a platform.

This includes:

1. `DriverOrder` / `Driver####`
2. `PlatformRecovery####`
3. `SysPrepOrder` / `SysPrep####`

---

Any platform that uses this driver essentially "locks out" these
variables from being created and supported by BDS on that platform.

Therefore, the following capability bit must not be set in `L"BootSupport"`
to comply with the UEFI Specification
(`EFI_BOOT_OPTION_SUPPORT_VARIABLE_NAME`): BIT4 - `EFI_BOOT_OPTION_SUPPORT_SYSPREP`.

---

Since `PlatformRecovery####` variables can be set during boot, this
driver checks if such variables were set prior to the driver called
being executed and deletes the variables and then sets the variable
policy.

This is to ensure consistent behavior where the variable cannot be set
regardless of whether the `SetVariable()` invocation occurred prior or
post this driver code executing during boot.

- [x] Impacts functionality?
- [x] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Verified variables are locked with [BootAuditTestApp](https://github.com/microsoft/mu_plus/blob/release/202208/UefiTestingPkg/AuditTests/BootAuditTest/UEFI/BootAuditTestApp.inf)
- Verified variables cannot be written in the OS
- Verified `PlatformRecovery####` variables are deleted if present

## Integration Instructions

Include this driver (`LoadOptionVariablePolicyDxe`) in a platform flash file (FDF)
preferably soon after the UEFI variable driver loads since this is dependent
on the variable read/write architectural protocols and variable policy protocols
both of which should be produced shortly beforehand allowing efficient dispatch
and rapid locking of the variables.

Run `BootAuditTestApp` to confirm variables are locked.

### DIFF
--- a/MsCorePkg/LoadOptionVariablePolicyDxe/LoadOptionVariablePolicyDxe.c
+++ b/MsCorePkg/LoadOptionVariablePolicyDxe/LoadOptionVariablePolicyDxe.c
@@ -1,0 +1,379 @@
+/** @file
+  Sets variable policy for Boot Manager load options that are not required to be supported per the UEFI Specification.
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <PiDxe.h>
+#include <Guid/GlobalVariable.h>
+#include <Library/BaseLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/DebugLib.h>
+#include <Library/MemoryAllocationLib.h>
+#include <Library/UefiBootServicesTableLib.h>
+#include <Library/UefiRuntimeServicesTableLib.h>
+#include <Library/VariablePolicyHelperLib.h>
+
+#define PLATFORM_RECOVERY_VARIABLE_NAME  L"PlatformRecovery"
+
+typedef struct {
+  CHAR16    *VariableName;
+  UINT32    VariableAttributes;
+} BM_LOAD_OPTION_VAR_INFO;
+
+//
+// Boot Manager load options that have variable policy applied by this driver.
+//
+GLOBAL_REMOVE_IF_UNREFERENCED BM_LOAD_OPTION_VAR_INFO  mBmLoadOptionInfo[] = {
+  // Fixed-Name Order Variables
+  {
+    EFI_DRIVER_ORDER_VARIABLE_NAME,
+    (EFI_VARIABLE_RUNTIME_ACCESS | EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_NON_VOLATILE)
+  },
+  {
+    EFI_SYS_PREP_ORDER_VARIABLE_NAME,
+    (EFI_VARIABLE_RUNTIME_ACCESS | EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_NON_VOLATILE)
+  },
+
+  // Wildcard Option Variables
+  {
+    L"Driver####",
+    (EFI_VARIABLE_RUNTIME_ACCESS | EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_NON_VOLATILE)
+  },
+  {
+    L"PlatformRecovery####",
+    (EFI_VARIABLE_RUNTIME_ACCESS | EFI_VARIABLE_BOOTSERVICE_ACCESS)
+  },
+  {
+    L"SysPrep####",
+    (EFI_VARIABLE_RUNTIME_ACCESS | EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_NON_VOLATILE)
+  }
+};
+
+/**
+  Gets the next UEFI variable name.
+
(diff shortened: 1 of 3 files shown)


os-d:
You mention that the compatibility bit needs to be unset in `L"BootSuppport"`, I'm guessing the right name is `L"BootSupport"`?

MsCorePkg/LoadOptionVariablePolicyDxe/LoadOptionVariablePolicyDxe.c
```diff
@@ -0,0 +1,379 @@
...
+
+    //
+    // Allocate a buffer to temporarily hold variable names. To reduce memory
+    // allocations, the default buffer size is 256 characters. The buffer can
+    // be reallocated if expansion is necessary (should be very rare).
+    //
+    *VariableNameBufferSize = sizeof (CHAR16) * 256;
+    *VariableName           = AllocateZeroPool (*VariableNameBufferSize);
+    if (*VariableName == NULL) {
+      return EFI_OUT_OF_RESOURCES;
```
os-d, line 110: Should `*VariableNameBufferSize` be set to 0 in this case in the out of resources case below? Reading the rest of the PR, this is only intended for use here, so doesn't really matter. STATIC keyword?

makubacki:
> You mention that the compatibility bit needs to be unset in `L"BootSuppport"`, I'm guessing the right name is `L"BootSupport"`?

Yeah, I updated it.

MsCorePkg/LoadOptionVariablePolicyDxe/LoadOptionVariablePolicyDxe.inf
```diff
@@ -0,0 +1,74 @@
...
+# """
+# If a platform permits the installation of Load Option Variables, (Boot####, or Driver####, or SysPrep####), the
+# platform must support and recognize all defined values for Attributes within the variable and report these
+# capabilities in BootOptionSupport. If a platform supports installation of Load Option Variables of type Driver####,
+# all installed Driver#### variables must be processed and the indicated driver loaded and initialized during every
+# boot. And all installed SysPrep#### options must be processed prior to processing Boot#### options.
+# """
+#
+# Any platform that uses this driver essentially "locks out" these variables from being created and supported
+# by BDS on that platform. Therefore, the following capability bit must not be set in L"BootSuppport" to comply with
```
os-d:
```suggestion
# by BDS on that platform. Therefore, the following capability bit must not be set in L"BootSupport" to comply with

```

MsCorePkg/LoadOptionVariablePolicyDxe/LoadOptionVariablePolicyDxe.inf
```diff
@@ -0,0 +1,74 @@
...
+# the driver called being executed and deletes the variables and then sets the variable policy. This is to ensure
+# consistent behavior where the variable cannot be set regardless of whether the SetVariable() invocation occurred
+# prior or post this driver code executing during boot.
+#
+#  Copyright (c) Microsoft Corporation.
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+[Defines]
+  INF_VERSION         = 0x00010005
```
os-d: very small nit: newer inf version?

os-d (approved):

MsCorePkg/LoadOptionVariablePolicyDxe/LoadOptionVariablePolicyDxe.inf
makubacki: Updated

MsCorePkg/LoadOptionVariablePolicyDxe/LoadOptionVariablePolicyDxe.inf
makubacki: Updated

apop5 (approved):

MsCorePkg/LoadOptionVariablePolicyDxe/LoadOptionVariablePolicyDxe.c
makubacki: Yes, I think `*VariableNameBufferSize` can be set to `0`.

Can you clarify what you meant by `STATIC` here?

makubacki: I'm going to go ahead and merge since you approved. We can follow up on the static comment later.

os-d: Yeah, that's totally fine. I just meant that the function could have the STATIC keyword on it, to make it clear it is only for consumption in this file, but it doesn't really matter.

PR #243, opened 2023-05-22
Continue power management on disabled APs

kuqin12:
# Preface

## Description

Current implementation will stop the execution on the APs that are disabled. This behavior will prevent the test suites from full validation, as some cores might be disabled by design.

This change will add a check prior to core power management operations and proceed with the other cores if the AP of interest is disabled.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on ARM FVP virtual platform with 16 cores. A shell based unit test is used for validation.

## Integration Instructions

N/A

### DIFF
--- a/UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.c
+++ b/UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.c
@@ -80,8 +80,8 @@ IsProcessorEnabled (
   UINTN                         CpuIndex
   )
 {
-  EFI_PROCESSOR_INFORMATION CpuInfo;
-  EFI_STATUS                Status;
+  EFI_PROCESSOR_INFORMATION  CpuInfo;
+  EFI_STATUS                 Status;
 
   if (Mp == NULL) {
     DEBUG ((DEBUG_ERROR, "%a Input protocol is NULL\n", __FUNCTION__));
--- a/UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.c
+++ b/UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.c
@@ -357,7 +357,6 @@ MpMgmtApOn (
     }
 
     if (!IsProcessorEnabled (mMpServices, Index)) {
-      DEBUG ((DEBUG_INFO, "%a Processor (%d) disabled, skipping processing\n", __FUNCTION__, Index));
       continue;
     }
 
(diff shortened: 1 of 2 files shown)


UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.c
```diff
@@ -64,6 +64,39 @@ GetProcessorInformation (
...
+BOOLEAN
+IsProcessorEnabled (
+  IN  EFI_MP_SERVICES_PROTOCOL  *Mp,
+  UINTN                         CpuIndex
+  )
+{
+  EFI_PROCESSOR_INFORMATION CpuInfo;
+  EFI_STATUS                Status;
+
+  if (Mp == NULL) {
```
apop5, line 86: Mp isn't used in this function, is the check on it being NULL just for completeness?

apop5:
This seems like a general problem with the MP services for arm. 
Is this scenario likely to happen in standard edk2 as well? or is this only exposed because of the testing that we've put into place?

apop5 (approved):

UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.c
kuqin12: Nice catch, it should have been the call to Mp instead of `mMpServices`.

kuqin12:
> This seems like a general problem with the MP services for arm. Is this scenario likely to happen in standard edk2 as well? or is this only exposed because of the testing that we've put into place?

It is mainly due to the tests. MP services for ARM will skip the disabled ones, if we use calls like "StartUpAllAps". In the test, we would want to enable a specific range, i.e. core 8-15, and the current MpManagement would do it in a loop, which will bail if the MP service returned error. And disabled cores will receive a "EFI_INVALID_PARAMETER" call. This change is somewhat to ignore that error by checking the status first.

os-d (approved):
Should this target release/202302 and be backported to release/202208 if necessary?

os-d (approved):
Should this target release/202302 and be backported to release/202208 if necessary?

kuqin12:
> Should this target release/202302 and be backported to release/202208 if necessary?

Theoretically, yes :) But this change was made and tested on 202208, so I thought it would make more sense to start the PR for 202208 and then cherry-pick it forward. But I will do it the other way around to be consistent with the general flow.

cfernald (approved):

UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/MpManagement.c
kuqin12: Fixed.

PR #259, opened 2023-06-08
MsCorePkg/PlatformBootManagerLib: Remove boot mode condition for con var update [Rebase & FF]

makubacki:
## Description

Some platforms may not reach BDS on a boot mode expected by the
current implementation to update the ConIn variable with a new
console input device.

For example, support for a new device might be added in a firmware
update, the update is performed, and the system is reset one or
more times before reaching PlatformBootManagerBeforeConsole().

This change always evaluates a potential update to ConIn regardless
of boot mode to reduce ConIn update complexity.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified the build and boot with the change. Since only the boot mode check
is removed, the underlying behavior within the boot mode condition remains
the same.

## Integration Instructions

Review the change to when console may be updated relative to boot mode in
change and the implementation of `EfiBootManagerUpdateConsoleVariable()`
being used by the platform to determine if you would like to make any
changes to that function. No required changes are expected.

### DIFF
--- a/MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
+++ b/MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
@@ -236,34 +236,24 @@ PlatformBootManagerBeforeConsole (
     FreePool (ConsoleOut);
   }
 
-  //
-  // Fill ConIn/ConOut in Full Configuration boot mode
-  //
-  DEBUG ((DEBUG_INFO, "%a - %x\n", __FUNCTION__, mBootMode));
-  if ((mBootMode == BOOT_WITH_FULL_CONFIGURATION) ||
-      (mBootMode == BOOT_WITH_DEFAULT_SETTINGS) ||
-      (mBootMode == BOOT_WITH_FULL_CONFIGURATION_PLUS_DIAGNOSTICS) ||
-      (mBootMode == BOOT_IN_RECOVERY_MODE))
-  {
-    if (PlatformConsoles != NULL) {
-      while ((*PlatformConsoles).DevicePath != NULL) {
-        //
-        // Update the console variable with the connect type
-        //
-        if (((*PlatformConsoles).ConnectType & CONSOLE_IN) == CONSOLE_IN) {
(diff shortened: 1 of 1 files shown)


joschock (approved):

makubacki:
Note: Targets release/202208 due to the platforms being tested against. Will be cherry-picked to release/202302 after PR completion.

kenlautner (approved):

MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
```diff
@@ -236,34 +236,24 @@ PlatformBootManagerBeforeConsole (
...
-        //
-        if (((*PlatformConsoles).ConnectType & CONSOLE_IN) == CONSOLE_IN) {
-          EfiBootManagerUpdateConsoleVariable (ConIn, (*PlatformConsoles).DevicePath, NULL);
-        }
-
-        if (((*PlatformConsoles).ConnectType & CONSOLE_OUT) == CONSOLE_OUT) {
-          EfiBootManagerUpdateConsoleVariable (ConOut, (*PlatformConsoles).DevicePath, NULL);
-        }
+  if (PlatformConsoles != NULL) {
+    while ((*PlatformConsoles).DevicePath != NULL) {
```
kuqin12: nit: Any idea on why we use the `*Pointer.Field` instead of `Pointer->Field`? This is not related to the change itself, so no need to block on it.

kuqin12 (approved):

MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
makubacki: No. I'd also prefer it to use `->`. I can add that in a separate commit to prevent confusion with the functional change being made.

makubacki: Updated

PR #271, opened 2023-07-06
Added logic to delete the Memory Type Information variable on a capsule update.

kenlautner:
## Description

Added logic to delete the Memory Type Information variable on a capsule update.  This is because in rare circumstance where a new memory type is added (or a memory type is removed) a capsule update will cause a mismatch of memory types.  Removing the variable and thus forcing a clean memory bucket discovery solves this issue.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on a platform where there is a capsule update between two versions of firmware with a different amount of memory types.  Without this change the transition from temp ram to permanent memory breaks but with this change the device correctly goes to the default memory bucket configuration and is able to boot.

## Integration Instructions

N/A

### DIFF
--- a/MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
+++ b/MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
@@ -349,8 +349,8 @@ PlatformBootManagerAfterConsole (
   )
 {
   EFI_STATUS  Status;
-  UINTN       VariableSize;
   UINT32      Attributes;
+  UINTN       VariableSize = 0;
 
   //  EFI_INPUT_KEY                 Key;
 
@@ -398,7 +398,7 @@ PlatformBootManagerAfterConsole (
                       &VariableSize,
                       NULL
                       );
-      if (!EFI_ERROR (Status) || (Status == EFI_BUFFER_TOO_SMALL)) {
+      if (Status == EFI_BUFFER_TOO_SMALL) {
         Status = gRT->SetVariable (
                         EFI_MEMORY_TYPE_INFORMATION_VARIABLE_NAME,
                         &gEfiMemoryTypeInformationGuid,
--- a/MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
+++ b/MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
@@ -349,6 +349,8 @@ PlatformBootManagerAfterConsole (
   )
 {
   EFI_STATUS  Status;
+  UINTN       VariableSize;
+  UINT32      Attributes,
 
   //  EFI_INPUT_KEY                 Key;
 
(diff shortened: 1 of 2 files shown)


kuqin12 (approved):

MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
```diff
@@ -388,6 +388,14 @@ PlatformBootManagerAfterConsole (
     case BOOT_ON_FLASH_UPDATE:
       EfiBootManagerConnectAll ();
       DEBUG ((DEBUG_INFO, "[%a] - signalling capsules are ready for processing\n", __FUNCTION__));
+      DEBUG ((DEBUG_INFO, "[%a] - Deleting Memory Type Information variable for capsule update\n", __FUNCTION__));
+      Status = gRT->SetVariable (
```
makubacki: The return from `SetVariable()` assigned to `Status` should be checked.

Within the scope of this BDS code, I think it should be considered that the variable doesn't exist. Trying to delete a variable that does not exist should return `EFI_NOT_FOUND`.

To satisfy both of these, I think `GetVariable()` should first be called to test for existence of the variable. If present, proceed to call `SetVariable()` with zero-sized data to delete the variable. In that case an error is not expected from the `SetVariable()` call so add an `ASSERT_EFI_ERROR (Status)` to catch that in assert-enabled builds.

kuqin12: To add on that, the variable attribute should come from the query of `GetVariable` instead of hardcoded. Otherwise, the deletion might fail.

kenlautner: Should be updated now.

MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
```diff
@@ -388,6 +390,25 @@ PlatformBootManagerAfterConsole (
     case BOOT_ON_FLASH_UPDATE:
       EfiBootManagerConnectAll ();
       DEBUG ((DEBUG_INFO, "[%a] - signalling capsules are ready for processing\n", __FUNCTION__));
+      DEBUG ((DEBUG_INFO, "[%a] - Deleting Memory Type Information variable for capsule update\n", __FUNCTION__));
+      Status = gRT->GetVariable (
```
makubacki, line 394: I suggest setting `VariableSize` to `0` before calling `GetVariable()` so you're not dependent on random values picked up in stack memory as to whether `EFI_BUFFER_TOO_SMALL` will be returned.

Then, solely consider `EFI_BUFFER_TOO_SMALL` to indicate data is present and proceed with deletion.

makubacki (approved):

MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
kenlautner: Updated.

PR #311, opened 2023-09-11
Only call HdwPortWrite if DebugLevel Met

### LINKED ISSUE #309: [Bug]: AdvancedLoggerHdwPortLib should not check PCD for header versions greater than 2

- [X] I have searched existing issues

### Current Behavior

wasted execution time on a hot path comparing the debug level against the PCD at both levels.   
https://github.com/microsoft/mu_plus/blob/a3d53ef6296bc0b1309cf2654017eb394a726dd4/AdvLoggerPkg/Library/AdvancedLoggerHdwPortLib/AdvancedLoggerHdwPortLib.c#L64

Why does calling code just mask the bits rather than avoid the function call?
https://github.com/microsoft/mu_plus/blob/a3d53ef6296bc0b1309cf2654017eb394a726dd4/AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c#L148

### Expected Behavior

Expected to be as efficient as possible and require minimal code in lower layers. 

### Steps To Reproduce

Look at code

### Build Environment

```markdown
NA
```

### Version Information

```text
a3d53ef6296bc0b1309cf2654017eb394a726dd4

all versions.
```

### Urgency

Medium

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

May have impact on downstream/out-of-tree projects and their implementations.

os-d:
## Description

The DebugLevel is checked twice in the hot path on serial path writes and the HdwPortWrite call is made even if the upper layer knows that the message being logged does not meet the DebugLevel criteria.

Closes #309.

In order to maintain backwards compatibility, if the LoggerInfo block is found to have a version less than the hardware logging level version, the PCD is checked to decide whether to call HdwPortWrite or not.

In SEC, because we may not have the LoggerInfo structure, we check the PCD to see if the message should be logged at this DebugLevel.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Build tested.

## Integration Instructions

If a bespoke AdvLoggerHdwPortLib is used, AdvLoggerHdwPortWrite should not check DebugLevel, but simply write the message to the hardware port.


apop5:
The intention of this second Pcd was to allow PcdFixedDebugPrintErrorLevel and  PcdAdvancedLoggerHdwPortDebugPrintErrorLevel to both function independently.  The debug messages written into the advanced logger memory buffer would be based on the PcdFixedDebugPrintErrorLevel  and the debug messages actually written out to a hardware port could be less (since the writes to some things, like serial ports, can be much slower than copying into a memory buffer).

Are you intending to remove this feature?

AdvLoggerPkg/Library/AdvancedLoggerHdwPortLib/AdvancedLoggerHdwPortLib.c
```diff
@@ -58,12 +58,5 @@ AdvancedLoggerHdwPortWrite (
   IN UINTN  NumberOfBytes
   )
 {
-  UINTN  NumberReturned;
-
-  NumberReturned = NumberOfBytes;
-  if (DebugLevel & PcdGet32 (PcdAdvancedLoggerHdwPortDebugPrintErrorLevel)) {
```
shrugupt, line 64: Since the intention is to check debug level in the upper layer only and "DebugLevel" argument is not used anymore in this API, may be "DebugLevel" argument can be removed altogether from this API.

os-d: I agree, but I didn't want to make API level changes in this PR, because there may be other implementations of AdvancedLoggerHdwPortLib for different hardware that uses it.

os-d:
> The intention of this second Pcd was to allow PcdFixedDebugPrintErrorLevel and PcdAdvancedLoggerHdwPortDebugPrintErrorLevel to both function independently. The debug messages written into the advanced logger memory buffer would be based on the PcdFixedDebugPrintErrorLevel and the debug messages actually written out to a hardware port could be less (since the writes to some things, like serial ports, can be much slower than copying into a memory buffer).
> 
> Are you intending to remove this feature?

This is not intended to remove that feature. PcdAdvancedLoggerHdwPortDebugPrintErrorLevel is still checked (it is loaded into LoggerInfo->HwPrintLevel). This just avoids checking this twice (once in AdvancedLoggerLib, once in AdvancedLoggerHdwPortLib) and avoids calling into AdvancedLoggerHdwPortLib when we know we won't be logging anyway.

In the legacy case, where the new code has loaded a LoggerInfo that doesn't have the HwPrintLevel field in it, we'll check the PCD explicitly.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
```diff
@@ -142,13 +142,24 @@ AdvancedLoggerWrite (
...
+  }
+
  #else
   if ((LoggerInfo != NULL) && (!LoggerInfo->HdwPortDisabled)) {
+    // if we are at a high enough version to support HW_LVL logging, only call the HdwPortWrite if this DebugLevel
+    // is asked to be logged
+    // if we are at an older version, check the PCD to see if we should log this message
     if (LoggerInfo->Version >= ADVANCED_LOGGER_HW_LVL_VER) {
-      DebugLevel = (DebugLevel & LoggerInfo->HwPrintLevel);
+      if (DebugLevel & LoggerInfo->HwPrintLevel) {
```
kuqin12, line 156: I do not think "HwPrintLevel" is meant to replace the PCD. The metadata value here is to provide a system wide mask inheritable from its loader. PCD, on the other hand, can be used for per module print level tuning.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
```diff
@@ -142,13 +142,24 @@ AdvancedLoggerWrite (
...
+        AdvancedLoggerHdwPortWrite (DebugLevel, (UINT8 *)Buffer, NumberOfBytes);
+      }
+    } else if (DebugLevel & PcdGet32 (PcdAdvancedLoggerHdwPortDebugPrintErrorLevel)) {
+      AdvancedLoggerHdwPortWrite (DebugLevel, (UINT8 *)Buffer, NumberOfBytes);
     }
+  }
 
  #endif
-    AdvancedLoggerHdwPortWrite (DebugLevel, (UINT8 *)Buffer, NumberOfBytes);
```
kuqin12, line 152: Continuing with the previous comment, if we want to stick to the same functionality, we can add a PCD check here and avoid the function call.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
kuqin12: Turned out this check once characteristic is needed for another reason. Resolve this comment here.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
kuqin12: Ditto.

kuqin12 (approved):

apop5 (approved):

PR #322, opened 2023-09-29
[REBASE && FF] Add FlatPageTableLib, Make Spellcheck Fixes, Update Paging Audit to Use FlatPageTableLib

TaylorBeebe:
## Description

Makes some spellcheck fixes.

Creates a new library, FlatPageTableLib, which works on X64 and AARCH64 platforms and converts the page table to a "flat" version. The flat version is a one-dimensional array where each entry is an address, a length, and attributes. The library will walk the page/translation table and combine blocks/leaves with the same attributes into a single entry in the flat array. The attributes mask for each architecture is defined in the header and includes both the upper and lower block/leaf attributes. On both X64 and AARCH64, the hierarchical inheritance of attributes is factored into the determination of block/leaf attributes. This allows the consumer of the library to easily check the attributes of any region in the page/translation table.

Updates DxePagingAuditTestApp to use FlatPageTableLib which allows us to delete the custom parsing logic in the test app.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [x] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by running the paging audit on SBSA and Q35 and by comparing the output against the Memory Attribute Protocol

## Integration Instructions

Platforms which build the paging audit will need to add an instance of FlatPageTableLib to their platform
DSC files.


UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
```diff
@@ -179,6 +242,44 @@ CanRegionBeRWX (
...
+  UT_ASSERT_NOT_NULL (mMap.Entries);
+  UT_ASSERT_NOT_EQUAL (mMap.EntryCount, 0);
+
+  Index           = 0;
+  FoundRWXAddress = FALSE;
+
+  for ( ; Index < mMap.EntryCount; Index++) {
+    if (IsPageExecutable (mMap.Entries[Index].PageEntry) && IsPageReadable (mMap.Entries[Index].PageEntry) && IsPageWritable (mMap.Entries[Index].PageEntry)) {
+      if (!CanRegionBeRWX (mMap.Entries[Index].LinearAddress, mMap.Entries[Index].Length)) {
+        UT_LOG_ERROR ("Memory Range 0x%llx-0x%llx is Read/Write/Execute\n", mMap.Entries[Index].LinearAddress, mMap.Entries[Index].LinearAddress + mMap.Entries[Index].Length);
```
os-d: nit: consider breaking this into multiple lines to avoid line wrap

UefiTestingPkg/Library/FlatPageTableLib/AArch64/FlatPageTableAArch64.c
```diff
@@ -0,0 +1,371 @@
...
+
+  @param[in, out]      Map         Pointer to the PAGE_MAP struct to be populated.
+
+  @retval RETURN_SUCCESS           The translation table is parsed successfully.
+  @retval RETURN_INVALID_PARAMETER MapCount is NULL, or Map is NULL but *MapCount is nonzero.
+  @retval RETURN_BUFFER_TOO_SMALL  *MapCount is too small.
+                                   MapCount is updated to indicate the expected number of entries.
+                                   Caller may still get RETURN_BUFFER_TOO_SMALL with the new MapCount.
+**/
+EFI_STATUS
```
os-d, line 277: nit: they are #defined to the same values, but the API claims to return an EFI_STATUS but returns a RETURN_STATUS. May as well be consistent.

os-d (approved):

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
TaylorBeebe: Updated

UefiTestingPkg/Library/FlatPageTableLib/AArch64/FlatPageTableAArch64.c
TaylorBeebe: Fixed

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
```diff
@@ -31,6 +36,67 @@ IMAGE_RANGE_DESCRIPTOR            *mNonProtectedImageList      = NULL;
 UINTN                             mSpecialRegionCount          = 0;
 EFI_GCD_MEMORY_SPACE_DESCRIPTOR   *mMemorySpaceMap             = NULL;
 UINTN                             mMemorySpaceMapCount         = 0;
+PAGE_MAP                          mMap                         = { 0 };
+
+/**
+  Frees the entries in the mMap global.
```
apop5, line 42: nit: Input parameter missing

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
```diff
@@ -31,6 +36,67 @@ IMAGE_RANGE_DESCRIPTOR            *mNonProtectedImageList      = NULL;
...
+  }
+
+  mMap.ArchSignature       = 0;
+  mMap.EntryCount          = 0;
+  mMap.EntryPagesAllocated = 0;
+}
+
+/**
+  Populate the global flat page table map.
+
```
apop5, line 60: nit: Input parameter missing

UefiTestingPkg/Include/Library/FlatPageTableLib.h
```diff
@@ -0,0 +1,166 @@
+/** @file
+  Library to parse page/translation table entries.
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef PAGE_TABLE_ATTRIBUTE_LIB_H_
```
apop5: Does this include guard need updated?

UefiTestingPkg/Include/Library/FlatPageTableLib.h
```diff
@@ -0,0 +1,166 @@
...
+  Library to parse page/translation table entries.
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef PAGE_TABLE_ATTRIBUTE_LIB_H_
+#define PAGE_TABLE_ATTRIBUTE_LIB_H_
+
```
apop5, line 15: do these structures need #pramga packed?

UefiTestingPkg/Library/FlatPageTableLib/AArch64/FlatPageTableAArch64.c
```diff
@@ -0,0 +1,371 @@
...
+          // BIT7 toggles write access for all ELs
+          ScratchEntry.Tteb.Bits.AccessPermissions |= BIT7; // Set BIT7
+          // 0b11 -> Access from EL0 is not allowed, no write permissions at any EL
+        } else if ((ParentHeritableAttributes.Bits.ApTable & TT_TABLE_AP_MASK) == TT_TABLE_AP_MASK) {
+          ScratchEntry.Tteb.Bits.AccessPermissions |= BIT7;            // Set BIT7
+          ScratchEntry.Tteb.Bits.AccessPermissions &= ~((UINT64)BIT6); // Clear BIT6
+        }
+      }
+
+      if ((*LastEntry != NULL) &&
```
apop5, line 219: LastEntry is not NULL checked before it is dereferenced here.

apop5 (approved):

UefiTestingPkg/Include/Library/FlatPageTableLib.h
TaylorBeebe: It seems like they should be but because these structures mirror those in CpuPageTableLib (which doesn't #pragma pack) I opted not to

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
TaylorBeebe: Added

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
TaylorBeebe: Added

UefiTestingPkg/Library/FlatPageTableLib/AArch64/FlatPageTableAArch64.c
TaylorBeebe: Added a NULL check on LastEntry at the top of the function

UefiTestingPkg/Include/Library/FlatPageTableLib.h
TaylorBeebe: Updated

UefiTestingPkg/Library/FlatPageTableLib/FlatPageTableLib.c
```diff
@@ -0,0 +1,93 @@
...
+**/
+
+#include <Uefi.h>
+#include <Library/BaseLib.h>
+#include <Library/SafeIntLib.h>
+#include <Library/FlatPageTableLib.h>
+
+// TRUE if A and B have overlapping intervals
+#define CHECK_OVERLAP(AStart, AEnd, BStart, BEnd) \
+  ((AStart <= BStart && AEnd > BStart) || \
```
spbrogan: do you need to confirm starts are less than ends?

UefiTestingPkg/Library/FlatPageTableLib/FlatPageTableLib.c
```diff
@@ -0,0 +1,93 @@
...
+  (BStart <= AStart && BEnd > AStart))
+
+/**
+  Checks the input flat page/translation table for the input region and converts the associated
+  table entries to EFI access attributes.
+
+  @param[in]  Map                 Pointer to the PAGE_MAP struct to be parsed
+  @param[in]  Address             Start address of the region
+  @param[in]  Length              Length of the region
```
spbrogan, line 31: in bytes?

UefiTestingPkg/Library/FlatPageTableLib/FlatPageTableLib.c
```diff
@@ -0,0 +1,93 @@
...
+  @param[out] Attributes          EFI Attributes of the region (EFI_MEMORY_XP, EFI_MEMORY_RO, EFI_MEMORY_RP)
+
+  @retval EFI_SUCCESS             The output Attributes is valid
+  @retval EFI_INVALID_PARAMETER   The flat translation table has not been built or
+                                  Attributes was NULL or Length was 0
+  @retval EFI_NOT_FOUND           The input region could not be found.
```
spbrogan: what happens if you span multiple entries with different attributes?

UefiTestingPkg/Library/FlatPageTableLib/FlatPageTableLib.c
```diff
@@ -0,0 +1,93 @@
...
+
+  FoundRange      = FALSE;
+  Index           = 0;
+  InputEndAddress = 0;
+
+  if (EFI_ERROR (SafeUint64Add (Address, Length - 1, &InputEndAddress))) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  do {
```
spbrogan, line 71: why do/while loop?  wouldn't loop logic be easier to see and more "standard" with just while loop or even for loop with break?

UefiTestingPkg/Library/FlatPageTableLib/FlatPageTableLib.inf
```diff
@@ -0,0 +1,51 @@
+## @file
+#  Library to parse page/translation table entries.
+#
```
spbrogan, line 7: a comment about why for application only would be good

UefiTestingPkg/Include/Library/FlatPageTableLib.h
spbrogan: Just to complete this conversation because I think the answer could be better/more educational for the team.  Packing is one of those odd things is often overdone.  @apop5  - why do they need packed?  

I don't think they need packed because you are not using them across compiler boundaries, and they are not being serialized or transported to other environments?  

It is important to note that there will be 32bits of wasted space added to the page_map to make up for the 32bit/64bit issue.  I doubt the entry will have anything added.

UefiTestingPkg/Include/Library/FlatPageTableLib.h
```diff
@@ -0,0 +1,166 @@
...
+  UINTN             EntryCount;
+  UINTN             EntryPagesAllocated;
+} PAGE_MAP;
+
+// The signature of the PAGE_MAP struct is used to determine the architecture of the page/translation table
+// entries.
+#define AARCH64_PAGE_MAP_SIGNATURE  SIGNATURE_32 ('A','A','6','4')
+#define X64_PAGE_MAP_SIGNATURE      SIGNATURE_32 ('X','6','4',' ')
+
+// This union can be used to interpret each PAGE_MAP_ENTRY on an AARCH64 system.
```
spbrogan, line 34: do these bitfields align with page table entry format or anything arch defined or are they just made up with the details you wanted?

UefiTestingPkg/Include/Library/FlatPageTableLib.h
```diff
@@ -0,0 +1,166 @@
...
+    UINT64    Nx                   : 1;  // BIT63
+  } Bits;
+  UINT64    Uint64;
+} X64_PAGE_MAP_ENTRY;
+
+// When the page/translation table is parsed to create an array of PAGE_MAP_ENTRY, the following bitmasks
+// are used to determine the attributes of one page/translation table entry are the same as another
+// page/translation table entry. If contiguous leaf/block entries have the same attributes, then they
+// will be represented in a single PAGE_MAP_ENTRY.
+#define AARCH64_ATTRIBUTES_MASK  ((0xFFFULL << 52) | (0x3FFULL << 2))
```
spbrogan: are these needed in public api and used by caller?

Do these operate on *_PAGE_MAP_ENTRY or on the arch defined page table entry.

UefiTestingPkg/Library/FlatPageTableLib/AArch64/FlatPageTableAArch64.c
```diff
@@ -0,0 +1,371 @@
...
+} TRANSLATION_TABLE_ENTRY_TABLE;
+
+typedef AARCH64_PAGE_MAP_ENTRY TRANSLATION_TABLE_ENTRY_BLOCK;
+
+typedef union {
+  TRANSLATION_TABLE_ENTRY_BLOCK        Tteb;
+  TRANSLATION_TABLE_ENTRY_TABLE        Ttet;
+  TRANSLATION_TABLE_ENTRY_HERITABLE    Tteh;
+  UINT64                               Uint64;
+} TRANSLATION_TABLE_ENTRY_UNION;
```
spbrogan, line 72: defining the source of these structures usually helps in the future and/or when arch makes updates known the impact of such an update.  are thse defined in arm spec?  what revision?

spbrogan:
This will just break the test app, right?  I am not sure we want to propagate a breaking change tag in the repo for test app code?  is this our normal practice for test apps?

UefiTestingPkg/Library/FlatPageTableLib/FlatPageTableLib.c
TaylorBeebe: I opted for do-while as opposed to a standard while so I could embed the index iteration within the parameter check instead of adding "Index++" at the end of the loop. A for loop also works -- this was just the style I happened to use.

UefiTestingPkg/Library/FlatPageTableLib/FlatPageTableLib.c
TaylorBeebe: The returned EFI access attributes are the common attributes across the range. If one range is XP/RO and the next range is XP/RP, then XP is returned as the access attribute spanning the range.

The Memory Attribute Protocol ASSERTs and returns invalid parameter if the attributes vary across the range which may be the more obvious route to take. I'll update this so varying attributes returns an error and move this logic into the paging audit app.

UefiTestingPkg/Include/Library/FlatPageTableLib.h
TaylorBeebe: These operate on the architecturally defined table entries. They are not needed in the public API and I put them here just to inform the library user how the logic should determine likeness between two entries in the page table.

The bitmask definitions can just be put here as a comment, though. I'll move the #define into the arch specific .c and leave a comment noting what the bitmasks are.

UefiTestingPkg/Include/Library/FlatPageTableLib.h
TaylorBeebe: These align with the table entry format. I'll add a comment noting that and from which architecture revisions they were pulled.

UefiTestingPkg/Library/FlatPageTableLib/FlatPageTableLib.c
TaylorBeebe: Updated

UefiTestingPkg/Library/FlatPageTableLib/FlatPageTableLib.c
TaylorBeebe: Updated

UefiTestingPkg/Library/FlatPageTableLib/FlatPageTableLib.inf
TaylorBeebe: added

UefiTestingPkg/Include/Library/FlatPageTableLib.h
TaylorBeebe: Added the architecture revision

UefiTestingPkg/Library/FlatPageTableLib/AArch64/FlatPageTableAArch64.c
TaylorBeebe: added

PR #327, opened 2023-10-06
[REBASE && FF] Update DxePagingAuditTestApp with Additional Shell and HTML Tests

TaylorBeebe:
Adds 8 tests to the paging audit shell app. Which check the following:
1. Unallocated memory is EFI_MEMORY_RP
2. Memory Attribute Protocol is present
3. Calls to allocate pages and pools return buffers with restrictive access attributes
4. NULL page is EFI_MEMORY_RP
5. MMIO Regions are Non Executable
6. Image code sections are EFI_MEMORY_RO and and data sections are EFI_MEMORY_XP
7. BSP stack is EFI_MEMORY_XP and has EFI_MEMORY_RP guard page
8. Memory outside of the EFI Memory Map is inaccessible

Adds 5 tests to the HTML templates:
1. Test that the NULL page is EFI_MEMORY_RP
2. Check that MMIO memory is non-executable.
3. Check that EfiConventionalMemory is non-executable.
4. Check that memory not in the EFI memory map is not accessible.
5. Check that the memory attribute protocol is present on the platform.

This also refactors much of the HTML, adds some quality of life updates to the output
HTML paging audit, adds logical OR filtering capability, and adds the collection of
Memory Attribute Protocol presence on the tested platform.

Tested on Q35, SBSA, and on development devices at UEFI Plugfest.

### DIFF
--- a/UefiTestingPkg/AuditTests/PagingAudit/UEFI/AArch64/PagingAuditProcessor.c
+++ b/UefiTestingPkg/AuditTests/PagingAudit/UEFI/AArch64/PagingAuditProcessor.c
@@ -8,6 +8,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 **/
 
 #include <Library/ArmLib.h>
+#include <Protocol/MemoryAttribute.h>
 #include "../PagingAuditCommon.h"
 
 extern MEMORY_PROTECTION_DEBUG_PROTOCOL  *mMemoryProtectionProtocol;
@@ -251,10 +252,11 @@ DumpPlatforminfo (
   VOID
   )
 {
-  CHAR8  TempString[MAX_STRING_SIZE];
-  UINTN  ExecutionLevel;
-  CHAR8  *ElString;
-  UINTN  StringIndex;
+  CHAR8                          TempString[MAX_STRING_SIZE];
+  UINTN                          ExecutionLevel;
+  CHAR8                          *ElString;
+  UINTN                          StringIndex;
+  EFI_MEMORY_ATTRIBUTE_PROTOCOL  *MemoryAttributeProtocol;
 
   ExecutionLevel = ArmReadCurrentEL ();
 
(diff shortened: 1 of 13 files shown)


apop5 (approved):

cfernald (approved):

UefiTestingPkg/AuditTests/PagingAudit/UEFI/AArch64/PagingAuditProcessor.c
```diff
@@ -268,13 +267,17 @@ DumpPlatforminfo (
     ElString = "Unknown";
   }
 
+  MemoryAttributeProtocol = NULL;
+  gBS->LocateProtocol (&gEfiMemoryAttributeProtocolGuid, NULL, (VOID **)&MemoryAttributeProtocol);
```
makubacki: I think this should check that `EFI_NOT_FOUND` is not returned in addition to checking whether the point is non-null to consider finding the attribute successful.

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
```diff
@@ -14,95 +14,209 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
+// Global for the flat page table
+PAGE_MAP  mMap = { 0 };
+
+// -------------------------------------------------
+//    GLOBALS SUPPORT FUNCTIONS
+// -------------------------------------------------
 
 /**
-  Frees the entries in the mMap global.
+  Return if the PE image section is aligned.
```
makubacki: It would be nice to mention in the function description that this will assert if data types memory types are given.

UefiTestingPkg/AuditTests/PagingAudit/UEFI/X64/PagingAuditProcessor.c
```diff
@@ -769,12 +771,16 @@ DumpPlatforminfo (
     PhaseString = "DXE";
   }
 
+  MemoryAttributeProtocol = NULL;
+  gBS->LocateProtocol (&gEfiMemoryAttributeProtocolGuid, NULL, (VOID **)&MemoryAttributeProtocol);
```
makubacki: Same comment as in PagingAuditProcessor.c. I think this should check that EFI_NOT_FOUND is not returned in addition to checking whether the point is non-null to consider finding the attribute successful.

makubacki (approved):

UefiTestingPkg/AuditTests/PagingAudit/UEFI/AArch64/PagingAuditProcessor.c
TaylorBeebe: Updated to set MemoryAttributeProtocol to NULL if the locate protocol call returns an error.

UefiTestingPkg/AuditTests/PagingAudit/UEFI/X64/PagingAuditProcessor.c
TaylorBeebe: Updated to set MemoryAttributeProtocol to NULL if the locate protocol call returns an error.

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
TaylorBeebe: Description updated to note that the function should only be called with the image code memory type otherwise it will ASSERT.

PR #334, opened 2023-10-17
HidPkg/UsbHidDxe: Continue on failure to get descriptor

makubacki:
## Description

In case a HID device fails to return a valid HID descriptor, this change
will return the error status from UsbHidDriverBindingStart() rather
than assert to match previous behavior from HID drivers that required
the boot protocol. The HID IO protocol will not be installed on these
devices.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Verified functionality is unchanged on physical platform with UsbHidDxe integrated.
- Verified boot previously encountering an assert on the QEMU virtual platform is
  not blocking boot.

## Integration Instructions

Update to a Mu Plus release with this change.

### DIFF
--- a/HidPkg/UsbHidDxe/UsbHidDxe.c
+++ b/HidPkg/UsbHidDxe/UsbHidDxe.c
@@ -723,7 +723,6 @@ ReadDescriptors (
 
   Status = UsbGetFullHidDescriptor (UsbHidDevice->UsbIo, UsbHidDevice->InterfaceDescriptor.InterfaceNumber, &UsbHidDevice->HidDescriptor);
   if (EFI_ERROR (Status)) {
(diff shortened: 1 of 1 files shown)


makubacki:
+@joschock for visibility. We discussed this offline to reduce the stringency on devices outside keyboards and mice that we don't believe were strictly probed for HID reports before.

apop5 (approved):

kuqin12 (approved):

joschock (approved):

PR #340, opened 2023-10-20
Add PCD to allow device exclusions for UsbHidDxe

joschock:
## Description

Adds a PCD for configuring device exclusions for UsbHidDxe. Prior to this change, the UsbHidDxe driver included logic to exclude keyboard devices from being handled by HID, due to other parts of the stack not being fully implemented yet. This removes that code and replaces it with a PCD list that the platform integrator can use to disable particular devices as desired for the platform.

- [x] Impacts functionality?
  - Changes how UsbHidDxe exclusions are handled.
- [ ] Impacts security?
- [x] Breaking change?
  Previously keyboard exclusion was hard-coded; to replicate this behavior platforms will need to add a PCD specification in the DSC files to exclude keyboards. See Integration Instructions below for more details.
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified with functional testing of several different exclusions in Qemu Q35. 

## Integration Instructions

To replicate existing behavior, platforms will need to add PCD specification to the DSC like the following to exclude USB keyboards:

```
[PcdsFixedAtBuild]
  gHidPkgTokenSpaceGuid.PcdExcludedHidDevices|{0x3, 0x1, 0x1, 0x0, 0x0, 0x0}

### DIFF
--- a/HidPkg/HidPkg.dec
+++ b/HidPkg/HidPkg.dec
@@ -35,6 +35,13 @@
   gHidKeyboardLayoutPackageGuid      = {0x57adcfa9, 0xc08a, 0x40b9, {0x86, 0x87, 0x65, 0xe2, 0xec, 0xe0, 0xe5, 0xe1}}
   gHidKeyboardLayoutKeyGuid          = {0xdcafaba8, 0xcde5, 0x40a1, {0x9a, 0xd3, 0x48, 0x76, 0x44, 0x71, 0x7e, 0x47}}
 
+[PcdsFixedAtBuild, PcdsPatchableInModule]
+  ## List of {InterfaceClass, InterfaceSubClass, InterfaceProtocol} tuples that should be excluded from being managed by
+  # the UsbHidDxe driver. The list is terminated by a record of all zeros.
+  # For example, to exclude USB keyboards, this would be set to {0x3, 0x1, 0x1, 0x0, 0x0, 0x0}
+  # @Prompt Exclude devices from UsbHidDxe management.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by makubacki — APPROVED** [review]

File: HidPkg/HidPkg.dec [hunk 1]
```diff
@@ -35,6 +35,13 @@
   gHidKeyboardLayoutPackageGuid      = {0x57adcfa9, 0xc08a, 0x40b9, {0x86, 0x87, 0x65, 0xe2, 0xec, 0xe0, 0xe5, 0xe1}}
   gHidKeyboardLayoutKeyGuid          = {0xdcafaba8, 0xcde5, 0x40a1, {0x9a, 0xd3, 0x48, 0x76, 0x44, 0x71, 0x7e, 0x47}}
 
+[PcdsFixedAtBuild, PcdsPatchableInModule]
+  ## List of {InterfaceClass, InterfaceSubClass, InterfaceProtocol} tuples that should be excluded from being managed by
+  # the UsbHidDxe driver. The list is terminated by a record of all zeros.
+  # For example, to exclude USB keyboards, this would be set to {0x3, 0x1, 0x1, 0x0, 0x0, 0x0}
```
apop5: The example is a good here, but would it be possible to add a link to either a file in the Tree or to a specification that enumerates the Interfaces, Interfaces subclasses and protocols?

apop5 (approved):

HidPkg/HidPkg.dec
joschock: Revised comments to add spec reference and further description.

apop5: Thank you!

PR #341, opened 2023-10-21
Update HelloWorldRustDxe to permit compiling for unit tests

joschock:
## Description

Update HelloWorldRustDxe to allow compiling tests.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
  - adds a sample unit test for the HelloWorldRustDxe driver.
- [ ] Includes documentation?

## How This Was Tested

Cargo test executes sample unit test as expected.
Build for UEFI target and execution QEMU work as expected.

## Integration Instructions
N/A

### DIFF
--- a/MsCorePkg/HelloWorldRustDxe/src/main.rs
+++ b/MsCorePkg/HelloWorldRustDxe/src/main.rs
@@ -8,39 +8,59 @@
 //!
 //! SPDX-License-Identifier: BSD-2-Clause-Patent
 //!
+
 #![no_std]
-#![no_main]
 #![allow(non_snake_case)]
+// UEFI requires "no_main" and a panic handler. To facilitate unit tests, gate "no_main", UEFI entry point, and panic
+// handler behind target_os configuration.
+#![cfg_attr(target_os = "uefi", no_main)]
+#[cfg(target_os = "uefi")]
+mod uefi_entry {
+  extern crate alloc;
 
-extern crate alloc;
+  use alloc::vec;
+  use core::panic::PanicInfo;
+  use r_efi::efi::Status;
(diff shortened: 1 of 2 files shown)


MsCorePkg/HelloWorldRustDxe/src/main.rs
```diff
@@ -8,39 +8,56 @@
 //!
 //! SPDX-License-Identifier: BSD-2-Clause-Patent
 //!
-#![no_std]
-#![no_main]
+#![cfg_attr(target_os = "uefi", no_std)]
```
makubacki: Within the context of this PR, this is obvious, but since this is a sample module, I think an explanatory comment describing that this is done to support `no_std` for UEFI builds while also allowing `std` to be used for unit testing would be helpful for people looking at it.

makubacki: Should also help them track the pattern throughout the rest of the file.

makubacki (approved):

MsCorePkg/HelloWorldRustDxe/src/main.rs
joschock: Technically, I think you could allow no_std for purposes of testing, though I haven't tried it; the important ones for enabling test builds are no_main and the panic handler being gated. I'll experiment a bit with whether no_std actually needs to be gated and update the PR.

joschock: Updated - removed target gate on no_main, and added a descriptive comment.

apop5 (approved):

PR #342, opened 2023-10-21
Add HiiKeyboardLayout crate to support UEFI HII Keyboard Layouts

joschock:
## Description

This crate provides a rust wrapper around UEFI HII Keyboard Layout structures. The relevant structures defined in the UEFI spec are not well-suited for direct definition in rust; so this crate defines analogous rust structures and provides serialization/deserialization support for converting the rust structures into byte buffers and vice versa. This crate uses the `scroll` crate (https://github.com/m4b/scroll) to facilitate serialization/deserialization of the Hii structures. 

## Examples and Usage
Retrieving a default (en-US) layout, writing it to a buffer, and then reading the buffer back into a rust structure:

```
 use hii_keyboard_layout::{get_default_keyboard_pkg, HiiKeyboardPkg};
 use scroll::{Pread, Pwrite};
 let mut buffer = [0u8; 4096];

 let package = get_default_keyboard_pkg();
 buffer.pwrite(&package, 0).unwrap();

 let package2: HiiKeyboardPkg = buffer.pread(0).unwrap();
 assert_eq!(package, package2);
```

- [x] Impacts functionality?
  - Introduces a new crate providing support for HII layouts.
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
  - Includes standard rust unit tests. 
- [x] Includes documentation?
  - Includes standard rust documentation.

## How This Was Tested

Verified by included unit tests.

## Integration Instructions
This crate requires the "scroll," "num-traits" and "num-drive" crates, so platforms intending to use it will need to add these as dependencies in their workspace Cargo.toml files. This PR does this for the workspace Cargo.toml that is at the root of mu_plus. 

Sample:
``` 
scroll = { version = "0.11", default-features = false, features = ["derive"]}
num-traits = { version = "0.2", default-features = false}
num-derive = { version = "0.4", default-features = false}

### CODE REVIEW

File: HidPkg/Crates/HiiKeyboardLayout/src/lib.rs [hunk 1]
```diff
@@ -0,0 +1,782 @@
...
+
+/// Enumeration of physical keys.
+/// Refer to UEFI spec version 2.10 section 34.8.10 and section 33.2.4.1.
+#[rustfmt::skip]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+//Note: UEFI specifies this as an C enum. That means the size is a bit ambiguous; but most compilers
+//will make it 32-bit, so that's what this implementation assumes.
+#[repr(u32)]
+pub enum EfiKey {
+  EfiKeyLCtrl, EfiKeyA0, EfiKeyLAlt, EfiKeySpaceBar, EfiKeyA2, EfiKeyA3, EfiKeyA4, EfiKeyRCtrl, EfiKeyLeftArrow,
```
makubacki: Do you think the enum members each need to be prefixed with `EfiKey` since they're in the `EfiKey` enum?

It makes usage like this `EfiKey::EfiKeyLCtrl` a bit repetitive.

HidPkg/Crates/HiiKeyboardLayout/src/lib.rs
```diff
@@ -0,0 +1,782 @@
...
+  hii::{self, PACKAGE_END},
+  protocols::hii_database::*,
+};
+use scroll::{ctx, Pread, Pwrite};
+
+/// HII Keyboard Package List
+/// Refer to UEFI spec version 2.10 section 33.3.1.2 which defines the generic header structure. This implementation
+/// only supports HII Keyboard Packages; other HII package types (or mixes) are not supported.
+#[derive(Debug, PartialEq, Eq)]
+pub struct HiiKeyboardPkgList {
```
makubacki, line 49: It seems like there is an opportunity to group these `Hii` related structures and implementation into a module named `hii` (like hii.rs) and then refer to them from this main library file as something like `hii::KeyboardPkgList`. That would keep the structures and associated implementation in one place and the other library functions in another place while reducing the overall size of this file.

Though the library is very "HII" centric so I could see preference to keep them in the main library file. Just a thought to separate things out some.

HidPkg/Crates/HiiKeyboardLayout/src/lib.rs
```diff
@@ -0,0 +1,782 @@
...
+  EfiKeyE1, EfiKeyE2, EfiKeyE3, EfiKeyE4, EfiKeyE5, EfiKeyE6, EfiKeyE7, EfiKeyE8, EfiKeyE9, EfiKeyE10, EfiKeyE11,
+  EfiKeyE12, EfiKeyBackSpace, EfiKeyIns, EfiKeyHome, EfiKeyPgUp, EfiKeyNLck, EfiKeySlash, EfiKeyAsterisk, EfiKeyMinus,
+  EfiKeyEsc, EfiKeyF1, EfiKeyF2, EfiKeyF3, EfiKeyF4, EfiKeyF5, EfiKeyF6, EfiKeyF7, EfiKeyF8, EfiKeyF9, EfiKeyF10,
+  EfiKeyF11, EfiKeyF12, EfiKeyPrint, EfiKeySLck, EfiKeyPause, EfiKeyIntl0, EfiKeyIntl1, EfiKeyIntl2, EfiKeyIntl3,
+  EfiKeyIntl4, EfiKeyIntl5, EfiKeyIntl6, EfiKeyIntl7, EfiKeyIntl8, EfiKeyIntl9,
+}
+
+impl TryFrom<u32> for EfiKey {
+  type Error = &'static str;
+  fn try_from(value: u32) -> Result<Self, Self::Error> {
```
makubacki, line 137: This is a common enough pattern (converting between an enum variant and integer), should we go ahead and bring in the [`num_derive`](https://docs.rs/num-derive/0.4.1/num_derive/#derives) crate and then `#[derive(FromPrimitive)]` on the enum?

makubacki: That should result in an `Option` and be able to avoid the unsafe transmute here.

joschock: happy to do that, it would be cleaner. But my initial intuition was to avoid the dependency.

makubacki: I thought the same when I ran into this, but I think we're inevitably going to keep having to do this given the code we're writing so it's a useful dependency.

makubacki (approved):
Approving with a few suggestions. If you feel those change the code substantially enough to request a rereview, please let me know.

HidPkg/Crates/HiiKeyboardLayout/src/lib.rs
joschock: I think there is not enough here to break out to a separate module, especially since these are rust wrapper structs (as opposed to hard FFI spec structs that r_efi has here: https://docs.rs/r-efi/latest/r_efi/protocols/hii_database/index.html). I'd like to keep the structs and their impls together.

HidPkg/Crates/HiiKeyboardLayout/src/lib.rs
joschock: Implemented suggestion.

HidPkg/Crates/HiiKeyboardLayout/src/lib.rs
joschock: Implemented suggestion.

HidPkg/Crates/HiiKeyboardLayout/src/lib.rs
```diff
@@ -0,0 +1,775 @@
...
+pub fn get_default_keyboard_layout() -> HiiKeyboardLayout {
+  HiiKeyboardLayout {
+    guid: efi::Guid::from_fields(0x3a4d7a7c, 0x18a, 0x4b42, 0x81, 0xb3, &[0xdc, 0x10, 0xe3, 0xb5, 0x91, 0xbd]),
+    keys: vec![
+      key!(EfiKey::C1,          'a',    'A',    '\0',   '\0',   NULL_MODIFIER,                AFFECTED_BY_STANDARD_SHIFT | AFFECTED_BY_CAPS_LOCK),
+      key!(EfiKey::B5,          'b',    'B',    '\0',   '\0',   NULL_MODIFIER,                AFFECTED_BY_STANDARD_SHIFT | AFFECTED_BY_CAPS_LOCK),
+      key!(EfiKey::B3,          'c',    'C',    '\0',   '\0',   NULL_MODIFIER,                AFFECTED_BY_STANDARD_SHIFT | AFFECTED_BY_CAPS_LOCK),
+      key!(EfiKey::C3,          'd',    'D',    '\0',   '\0',   NULL_MODIFIER,                AFFECTED_BY_STANDARD_SHIFT | AFFECTED_BY_CAPS_LOCK),
+      key!(EfiKey::D3,          'e',    'E',    '\0',   '\0',   NULL_MODIFIER,                AFFECTED_BY_STANDARD_SHIFT | AFFECTED_BY_CAPS_LOCK),
+      key!(EfiKey::C4,          'f',    'F',    '\0',   '\0',   NULL_MODIFIER,                AFFECTED_BY_STANDARD_SHIFT | AFFECTED_BY_CAPS_LOCK),
```
Javagedes, line 475: nit: The first couple of keys end in `AFFECTED_BY_STANDARD_SHIFT | AFFECTED_BY_CAPS_LOCK),` and then later an extra space is added: `AFFECTED_BY_STANDARD_SHIFT | AFFECTED_BY_CAPS_LOCK ),`.

Also, starting at `E1` the lines extend further than before and do not match.

I noticed that you turned off the rust formatter here, so if this is by design, or is not worth changing, just close this.

Javagedes (approved):

HidPkg/Crates/HiiKeyboardLayout/src/lib.rs
joschock: I turned off the formatter because by default it will put every argument to key!() on its own line, which is anarchy. But I will fix the formatting issues you identified.

joschock: Fixed.

Issue #350, opened 2023-11-04
[Bug]: Cannot build DisplayEngineDxe on ARM


sonic011gamer:
### Is there an existing issue for this?

- [X] I have searched existing issues

### Current Behavior

On 32-bit ARM DisplayEngineDxe will not compile because of a linking error caused by sqrt_d in MathLib.

### Expected Behavior

DisplayEngineDxe to build without errors.

### Steps To Reproduce

Compile a project with ARM target that has DisplayEngineDxe requirement.

### Build Environment

```markdown
- OS(s): Windows 11 with WSL2 Ubtuntu 20.04
- Tool Chain(s): CLANG with arm-linux-gnueabihf-
- Targets Impacted: ARM
```

### Version Information

```text
Tag: release/202302
```

### Urgency

High

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

Maintainer feedback requested

### Anything else?

[BUILDLOG_ExampleApplePkg.txt](https://github.com/microsoft/mu_plus/files/13256356/BUILDLOG_ExampleApplePkg.txt)

PR #351, opened 2023-11-04
Update MathLib.c to fix DisplayEngineDxe not compiling on ARM

sonic011gamer:
## Description

This change was made for fixing DisplayEngineDxe giving a linking error on 32-bit ARM architecture because of sqrt_d in MathLib. It fixes issue #350 

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

<_Please describe the test(s) that were run to verify the changes._>

## Integration Instructions

N/A

### DIFF
--- a/MsCorePkg/Library/MathLib/MathLib.c
+++ b/MsCorePkg/Library/MathLib/MathLib.c
@@ -218,7 +218,7 @@ sqrt_d (
   IN CONST double  input
   )
 {
-  UINT64  firstGuess = (UINT64)input;
+  UINT32  firstGuess = (UINT32)input;
   double  x          = 0;
   double  prevX      = -1;
 


MsCorePkg/Library/MathLib/MathLib.c
```diff
@@ -218,7 +218,7 @@ sqrt_d (
   IN CONST double  input
   )
 {
-  UINT64  firstGuess = (UINT64)input;
+  UINT32  firstGuess = (UINT32)input;
```
kuqin12, line 221: I might be missing something on this, but wouldn't this cast cause the upper 32 bit getting truncated as a double should be 64bit in width?

sonic011gamer: This is exactly why I asked for a review, while this fixed building, it's not the best solution, hence I need someone with experience to look at this

makubacki: @sonic011gamer, looking at your build log `__aeabi_ul2d` looks like a compiler built-in function that is not defined. We define those for Arm an AARCH64 in https://github.com/microsoft/mu_basecore/tree/release/202302/MdePkg/Library/CompilerIntrinsicsLib.

That usually gets linked to every module by applying it like this in a DSC:
`NULL|MdePkg/Library/CompilerIntrinsicsLib/ArmCompilerIntrinsicsLib.inf`

Could you try defining that function there?

makubacki: It looks like there is a definition in `ArmSoftFloatLib` in https://github.com/microsoft/mu_silicon_arm_tiano/tree/release/202302/ArmPkg/Library/ArmSoftFloatLib. Are you linking that against this module?

sonic011gamer: > It looks like there is a definition in `ArmSoftFloatLib` in https://github.com/microsoft/mu_silicon_arm_tiano/tree/release/202302/ArmPkg/Library/ArmSoftFloatLib. Are you linking that against this module?

yes, I am building with this module and the one above mentioned included in the .dsc

PR #358, opened 2023-11-14
Resolve deadlock in RustBootServicesAllocatorDxe

joschock:
## Description

Fix an issue where rust boot services allocator implementation could deadlock if memory allocations were attempted at different TPLs. Deadlock occurs in following scenario:
1. Task running at low TPL initiates an allocation request.
2. While that allocation is in progress and the lock is held, an interrupt occurs, and a new task executes at a higher TPL
3. new task attempts an allocation. 

Deadlock occurs because the higher TPL task cannot acquire the lock because it is held by the lower TPL task, and the lower TPL task cannot make forward progress because it has been interrupted by the higher TPL task.

To resolve this, this PR updates the allocation implementation to remove the spinlock. An AtomicPtr is used to give well-ordered access to the bootservices pointer used as the basis for the allocation implementation. Other aspects of the implementation (i.e. creation of the allocation tracker) are already thread-safe or are the responsibility of the boot services implementation. 

PR also contains some minor style cleanup. 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
  - RustBootServicesAllocatorDxe unit tests updated to accommodate new implementation.
- [ ] Includes documentation?

## How This Was Tested

Verified boot on QEMU.

## Integration Instructions

N/A

### DIFF
--- a/MsCorePkg/Crates/RustBootServicesAllocatorDxe/src/lib.rs
+++ b/MsCorePkg/Crates/RustBootServicesAllocatorDxe/src/lib.rs
@@ -34,19 +34,14 @@
 use core::{
   alloc::{GlobalAlloc, Layout},
   ffi::c_void,
+  sync::atomic::AtomicPtr,
 };
 
-use r_efi::{
-  efi::{BootServices, Status},
-  system::BOOT_SERVICES_DATA,
-};
+use r_efi::efi;
 
 /// Static GLOBAL_ALLOCATOR instance that is marked with the `#[global_allocator]` attribute.
-///
-/// See [`core::alloc::GlobalAlloc`] for more details on rust global allocators. This allocator must be initialized
-/// before use, see [`SpinLockedAllocator::init()`].
 #[cfg_attr(not(test), global_allocator)]
-pub static GLOBAL_ALLOCATOR: SpinLockedAllocator = SpinLockedAllocator::new();
+pub static GLOBAL_ALLOCATOR: BootServicesAllocator = BootServicesAllocator::new();
 
 const ALLOC_TRACKER_SIG: u32 = 0x706F6F6C; //arbitrary sig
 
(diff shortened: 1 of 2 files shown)


makubacki (approved):

MsCorePkg/Crates/RustBootServicesAllocatorDxe/src/lib.rs
```diff
@@ -153,31 +138,57 @@ impl BootServicesAllocator {
...
   unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
-    self.inner.lock().boot_services_alloc(layout)
+    let bs_ptr = self.boot_services.load(core::sync::atomic::Ordering::SeqCst);
+    if let Some(boot_services) = unsafe { bs_ptr.as_ref() } {
+      let ptr;
+      let old_tpl = (boot_services.raise_tpl)(efi::TPL_NOTIFY);
+      if let Some(allocator) = self.inner.try_lock() {
+        ptr = allocator.boot_services_alloc(layout, boot_services);
+      } else {
+        panic!("failed to acquire memory allocation lock at TPL_NOTIFY");
```
makubacki: My understanding is that most of the panic handler implementations are currently going to print a `DEBUG_ERROR` message and then loop indefinitely. Is this expected to potentially occur in a production integration scenario? If so, is that the behavior you're intending for this case?

joschock: I could alternatively silently return a null pointer to indicate allocation failed. However, gBS->Allocate*() is going to ASSERT with a TPL violation in the same condition, so I was thinking that panic would be a better aid to immediately identifying bad code. Happy to change it though, I actually implemented the null pointer return first, but then decided I was more opinionated about it - but I'm not _that_ opinionated.

makubacki: I'm not super opinionated either and I agree with bringing attention to unanticipated edge cases, but lock contention is something that is hard to predict, as it's very dynamic based on the flows involved on a per-platform basis.

My main concern is about potentially placing the system in an unrecoverable state if one of those instances that occurs more often than expected. Typically, we use the same assert behavior across our platforms which is similar to the panic handler implementations. But I thought the panic might apply that behavior (i.e. in `RELEASE` builds) diverging from our normal assert behavior in those cases.

That's why I'm leaning a little more in favor of relying upon the asserts in `gBS->Allocate()` for now.

joschock: I think just removing the lock implementation altogether to defer to gBS->Allocate*() TPL management is probably the right way to go. Will update this PR with that change.

joschock: Implementation updated; ready for a re-review.

makubacki: Thanks for pushing the update.

Javagedes (approved):

PR #361, opened 2023-11-15
Add SetProtocol command to force devices to Report Mode

joschock:
## Description

USB HID specification 1.11 (https://www.usb.org/document-library/device-class-definition-hid-111) section 7.2.6 states:
```
When initialized, all devices default to report protocol. However the host should 
not make any assumptions about the device’s state and should set the desired 
protocol whenever initializing a device. 
```

In testing actual devices, it has been observed that some actual endpoint HID devices come up in "boot protocol" rather than "report protocol." This PR implements the recommendation that the host (in this case, the UsbHidDxe driver) not make any assumptions about device state, and explicitly sets report protocol for devices that implement the "boot" interface subclass. 

The PR also makes a minor adjustment to debug verbosity. 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified with QEMU that command is sent; verified with hardware that has this issue that SetProtocol resolves it. 

## Integration Instructions

N/A

### DIFF
--- a/HidPkg/UsbHidDxe/UsbHidDxe.c
+++ b/HidPkg/UsbHidDxe/UsbHidDxe.c
@@ -26,7 +26,12 @@
 
 #include <IndustryStandard/Usb.h>
 
-#define CLASS_HID  3
+#define CLASS_HID      3
+#define SUBCLASS_BOOT  1
+
+// Refer to USB HID 1.11, section 7.2.6.
+#define BOOT_PROTOCOL    0
+#define REPORT_PROTOCOL  1
 
 //
 // A common header for usb standard descriptor.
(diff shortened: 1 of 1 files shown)


HidPkg/UsbHidDxe/UsbHidDxe.c
```diff
@@ -26,7 +26,12 @@
 
 #include <IndustryStandard/Usb.h>
 
-#define CLASS_HID  3
+#define CLASS_HID      3
+#define SUBCLASS_BOOT  1
+
+// Refer to USB HID 1.11, section 7.2.6.
+#define BOOT_PROTOCOL    0
+#define REPORT_PROTOCOL  1
```
makubacki, line 34: I know these are also defined in UsbKbHidDxe and UsbMouseHid. Do you think it is time to put it somewhere common given it is defined in the USB HID spec?

makubacki (approved):

HidPkg/UsbHidDxe/UsbHidDxe.c
joschock: No objection, seems like they would go in MdePkg\Include\IndustryStandard\Usb.h, which currently has some UsbMass class-specific stuff. Would you like me to raise a PR against basecore for this?

joschock: In any case, if this is added to basecore or someplace else, I propose raising that as a separate change.

makubacki: > In any case, if this is added to basecore or someplace else, I propose raising that as a separate change.

Yes, that's fine and all the uses would need to be updated.

apop5 (approved):

PR #376, opened 2023-12-01
MemoryProtectionTestApp: Separate Reset Method Init to Arch Specific Files

TaylorBeebe:
## Description

The reset test method is not supported on ARM platforms currently (support will be added later, but it's not high priority because this test method often takes 40+ minutes). This PR separates the initialization of this testing method to architecture specific logic so the test method is not attempted on ARM platforms.

This also fixes a build error on ARM platforms caused by DefaultExceptionHandlerLib being included (which does not support UEFI_APPLICATION without a MU override).

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Running the updated test app on Q35 and SBSA

## Integration Instructions

N/A


UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/AArch64/AArch64Functions.c
```diff
@@ -0,0 +1,46 @@
...
+  @retval     others              Error occurred during installation.
+**/
+EFI_STATUS
+EFIAPI
+RegisterMemoryProtectionTestAppInterruptHandler (
+  VOID
+  )
+{
+  // TODO: Add support for reset test method on Arm
+  return EFI_UNSUPPORTED;
```
kuqin12, line 27: Shouldn't ARM support similar exception handling as the x64 routine?

kuqin12 (approved):

apop5 (approved):

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/AArch64/AArch64Functions.c
TaylorBeebe: Yes, but I need to refresh which synchronous exceptions will occur on translation fault and it will take a quite a while to actually test that solution so I have it on the back burner for now.

PR #388, opened 2023-12-15
Add prefix for advanced logger memory message entries

### LINKED ISSUE #375: [Feature]: Add prefix for different firmware entities in memory logging
### Feature Overview

Currently all firmware entities that supports advanced memory logging will dump messages into the memory buffer, without any distinguishable notations. Once the parsed log is retrieved from target, all prints are mushed together, making it hard to read.

The request is to add prefix for different instances of advance logging parties, making it easier to parse and locate logs.

### Solution Overview

Adding an API for each instance that produces `AdvancedLoggerGetLoggerInfo`, call it `AdvancedLoggerGetStringPrefix`. This function will return a constant string and its size. This function will be invoked from `AdvancedLoggerMemoryLoggerWrite`, where the return will be copied to the front of each memory log message.

We should make the prefix string minimal, as it will bloat the memoy usage in the memory logging buffer.

### Alternatives Considered

1. Create a static PCD for the prefix, so that each component type can use that PCD to prefix the messages.
    - This is a simpler solution by offloading the PCD config to platforms. However, the change will require in-depth analysis of the build configuration and could lead to integration errors.
3. Create a message entry v2. Instead of materializing the prefix into the memory logging buffer, add a field indicating the library instance (i.e. PEI core for 0, PEIM for 1, etc.) so that the parser can handle the prefix.
    - This is a more ideal solution, but the issue is that the parser application will need to be updated accordingly as this is introducing a breaking change, of which the impact is unknown.

### Urgency

Medium

I will implement the feature

Maintainer feedback requested

### Anything else?

_No response_

kuqin12:
# Preface

## Description

This change introduces a new message entry structure, which contains extra information of log producer and can be used to differentiate the different boot phases, i.e. PEI, DXE, SMM, etc. Platforms that have multiple firmware entities can leverage this extra information to distinguish coalesced memory logging regions.

The updated applications (UEFI shell app and host OS Python script) are also updated to support new message entry structure while maintaining the backwards compatibility.

Resolves https://github.com/microsoft/mu_plus/issues/375.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [x] Breaking change?
    in build or boot behavior?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change was tested on QEMU Q35 and verified that prefix are properly injected for DXE phase prints.

## Integration Instructions

The log parsers are updated, thus platforms use the existing parser should pick up the latest tool from mu_plus. For platforms that carry their own implementation, they should add the proper support in their own implementation before picking up this change from mu_plus.


MsGraphicsPkg/DisplayEngineDxe/DisplayEngineDxe.inf
```diff
@@ -7,8 +7,6 @@
 #
 ##
 
-#Override : 00000002 | MdeModulePkg/Universal/DisplayEngineDxe/DisplayEngineDxe.inf | 60ed12cbc6548db7016bd842943cb8be | 2022-02-07T20-47-30 | 683ed68b7ecab2be6740359535a52a3ea086dd8a
```
kuqin12: Revert...

kuqin12: Reverted...

os-d: Looks like this hasn't been reverted.

os-d (approved):

MsGraphicsPkg/DisplayEngineDxe/DisplayEngineDxe.inf
kuqin12: Shoot, this might be added back when I was tuning it. Thanks for catching that!

MsGraphicsPkg/DisplayEngineDxe/DisplayEngineDxe.inf
```diff
@@ -7,7 +7,7 @@
 #
 ##
 
-#Override : 00000002 | MdeModulePkg/Universal/DisplayEngineDxe/DisplayEngineDxe.inf | 60ed12cbc6548db7016bd842943cb8be | 2022-02-07T20-47-30 | 683ed68b7ecab2be6740359535a52a3ea086dd8a
+# Override : 00000002 | MdeModulePkg/Universal/DisplayEngineDxe/DisplayEngineDxe.inf | 60ed12cbc6548db7016bd842943cb8be | 2022-02-07T20-47-30 | 683ed68b7ecab2be6740359535a52a3ea086dd8a
```
kuqin12, line 10:
```suggestion
#Override : 00000002 | MdeModulePkg/Universal/DisplayEngineDxe/DisplayEngineDxe.inf | 60ed12cbc6548db7016bd842943cb8be | 2022-02-07T20-47-30 | 683ed68b7ecab2be6740359535a52a3ea086dd8a
```

MsGraphicsPkg/DisplayEngineDxe/DisplayEngineDxe.inf
kuqin12: Now it is really reverted.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
```diff
@@ -40,16 +40,16 @@ AdvancedLoggerMemoryLoggerWrite (
...
+  ADVANCED_LOGGER_INFO              *LoggerInfo;
+  EFI_PHYSICAL_ADDRESS              CurrentBuffer;
+  EFI_PHYSICAL_ADDRESS              NewBuffer;
+  EFI_PHYSICAL_ADDRESS              OldValue;
+  UINT32                            OldSize;
+  UINT32                            NewSize;
+  UINT32                            CurrentSize;
+  UINTN                             EntrySize;
+  UINTN                             UsedSize;
+  ADVANCED_LOGGER_MESSAGE_ENTRY_V2  *Entry;
```
kuqin12, line 52: Need to add a PCD to make this backwards compatible for tooling transitioning?

kuqin12: Not really, all known consumers are using the tooling together with the source.

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
```diff
@@ -128,22 +180,43 @@ AdvancedLoggerAccessLibGetNextMessageBlock (
...
+      LogEntryV2 = NEXT_LOG_ENTRY_V2 (LogEntryV2);
+    } else {
+      LogEntry = NEXT_LOG_ENTRY (LogEntry);
     }
+  }
 
-    LogEntry = NEXT_LOG_ENTRY (LogEntry);
+  // At this point, if LogEntryV2 is not NULL, it points to the next entry to be read.
+  // Otherwise LogEntry will contain the next entry. So we simplify the logic by only
+  // always start from LogEntry. However, note that regardless of how we inherit the pointer
```
kenlautner: The line: `So we simplify the logic by only always start from LogEntry.` is confusing.  I'd recommend rewording to something like this: `So we simplify the logic by only using LogEntry and overwriting it to use the LogEntryV2 data as necessary.`

kenlautner (approved):

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
kuqin12: Updated. Thanks.

PR #389, opened 2023-12-15
Add PeiCore method to find AdvancedLogger log buffer if LoggerInfo is…

apop5:
## Description

This was discovered in a platform that uses AdvancedLogger Starting in Pei, but without permanent memory until after MRC runs.
(i.e. AdvancedLoggerLib PeiCore instance allocates CAR heap for the log buffer)

The platform makes uses of EFI_PEI_TEMPORARY_RAM_DONE_PPI, where is tears down all variable range MTRRs.

When PeiCore is reentered after memory has become available, it will relocate Hobs to system memory, then it will go through the Hobs and fix up the memory allocation pointers to point to the system memory addresses instead of the Heap addresses.

During this transition, the PlatformBlob pointer will become invalid, after CAR is torn down.

During the PeiCore's `PeiServicesInstallPpi (&mMemoryDiscoveredPpi);`, a debug message will attempt to be output about the `Install PPI: ` of the Memory Discovered Ppi, and the current logic will attempt to use the PlatformBlob, which results in a memory exception (the pointer is all Fs).

This PR updates the logic in `AdvancedLoggerGetLoggerInfo` to verify the PlatformBlob pointer contains the correct signature before returning it, and if that fails, it then will go through Memory Allocation Hobs to try to find a memory allocation that corresponds to the AdvancedLogger buffer. If it fails, it will fall through to the existing functionality of creating a new allocation hob.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified that the system that was triggering an exception was able to function correctly after this change was incorporated.

## Integration Instructions

N/A


makubacki (approved):
As discussed in our offline discussion, this is fine for now. Ideally there would be a more explicit way for services depended on by the PEI core to transition their contents based on time of use of those services in the core.

kuqin12 (approved):

PR #395, opened 2023-12-18
[REBASE && FF] FlatPageTableLib and DxePagingAuditTestApp Updates

TaylorBeebe:
## Description

This change set was primarily created to improve the accuracy of the DXE paging audit shell test. The problem was two-fold:
1. The layout of memory would change as data was collected. This is solved by pre-allocating space for two of the memory maps (flat page table and EFI memory map). Those two maps will have their size requirement checked at the start of each function to ensure enough heap space has been allocated for them prior to populating them with memory information.
2. The shell test was imprecise when reporting which regions failed the various tests. I'll use the UnallocatedMemoryIsRP() test as an example. The test parses the EFI memory map for EfiConventionalMemory regions which indicate unallocated memory. If any subsection of the region was not read protected, then the entire region would be reported as failing. With this update, every subsection which does not comply will be individually reported giving platform developers more precise information on where the problem areas are.

Patch breakdown:

1. Updates the DXE paging audit to always write out all .dat files even if the buffers are empty. The presence of the file acts as a receipt that the audit was run successfully and helps with automated unit testing.
2. Adds a function to FlatPageTableLib to dump the contents of a flat page table to the console. This is useful for debugging.
3. Updates GetRegionAccessAttributes() in FlatPageTableLib to return the attributes of the first attribute-contiguous range found in the region and report the actual size of that range. This allows the caller to collect the attributes of the region by calling the function repeatedly instead of needing to guess which subsection of the region has contiguous attributes.
4. Updates the DXE paging audit shell tests to pre-allocate memory for memory maps. The page table map, EFI memory map, and EFI memory space map all describe the layout of the system address space. Because of this, if allocations are performed while these maps are being generated, then the maps generated the earliest will be inaccurate.
5. Adds ValidatePageTableAttributes() to the DXE paging audit shell test logic. This function validates the attributes of the input memory region. It uses GetRegionAccessAttributes() to get the attributes of the region and compares them to the attributes passed in. This function will be used by each test case to check that the page/translation table attributes match the memory protection requirements and enables more precise reporting of regions which do not match the requirements.
6. Updates the DXE paging audit shell tests to use ValidatePageTableAttributes() to determine policy compliance.

## How This Was Tested

This was tested on Q35 and an ARM platform. Some changes in test caused the UnallocatedMemoryIsRP() test to erroneously fail due to reason 1 above which has been fixed with this patch set. Additionally, this patch set was checked by comparing the shell test output to the full HTML paging audit to verify that errant regions are exact and consistent.

## Integration Instructions

N/A

### DIFF
--- a/UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
+++ b/UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
@@ -568,6 +568,121 @@ GetRegionCommonAccessAttributes (
   return FoundRange ? EFI_SUCCESS : EFI_NOT_FOUND;
 }
 
+/**
+  Checks the input flat page/translation table for the input region and validates
+  the attributes match the input attributes.
+
+  @param[in]  Map                   Pointer to the PAGE_MAP struct to be parsed
+  @param[in]  Address               Start address of the region
+  @param[in]  Length                Length of the region
+  @param[in]  RequiredAttributes    The required EFI Attributes of the region
+  @param[in]  MatchAnyAttribute     If TRUE, the region must contain at least one of the
+                                    required attributes. If FALSE, the region must contain
+                                    all of the required attributes.
+  @param[in]  AllowUnmappedRegions  If TRUE, unmapped regions are excepted from the check.
+  @param[in]  LogAttributeMismatch  If TRUE, log the attribute mismatch via DEBUG().
+
+  @retval TRUE                   The region has the required attributes
+  @retval FALSE                  The region does not have the required attributes
+**/
+STATIC
+BOOLEAN
+ValidateRegionAttributes (
+  IN PAGE_MAP  *Map,
+  IN UINT64    Address,
+  IN UINT64    Length,
+  IN UINT64    RequiredAttributes,
+  IN BOOLEAN   MatchAnyAttribute,
+  IN BOOLEAN   AllowUnmappedRegions,
+  IN BOOLEAN   LogAttributeMismatch
+  )
+{
+  UINT64      RegionAttributes;
+  UINT64      CheckedLength;
+  EFI_STATUS  Status;
+  BOOLEAN     AttributesMatch;
+
+  AttributesMatch = TRUE;
+
+  do {
+    RegionAttributes = 0;
+    CheckedLength    = 0;
+    Status           = GetRegionAccessAttributes (
+                         Map,
+                         Address,
+                         Length,
+                         &RegionAttributes,
+                         &CheckedLength
+                         );
+
+    // If the region was completely or partially matched, check the returned attributes against the
+    // expected attributes
+    if ((Status == EFI_SUCCESS) || (Status == EFI_NOT_FOUND)) {
+      if (((!MatchAnyAttribute && ((RegionAttributes & RequiredAttributes) != RequiredAttributes)) ||
+           (MatchAnyAttribute && ((RegionAttributes & RequiredAttributes) == 0))))
+      {
+        if (LogAttributeMismatch) {
+          UT_LOG_ERROR (
+            "Region 0x%llx-0x%llx does not %a%a%a%a\n",
+            Address,
+            Address + CheckedLength,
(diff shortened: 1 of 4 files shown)


makubacki:
I think it's useful to add back to the testing and integration sections to the PR description and summarize that information there. Gets rolled up into release notes for integrators to quickly review.

UefiTestingPkg/Include/Library/FlatPageTableLib.h
```diff
@@ -104,29 +104,54 @@ CreateFlatPageTable (
...
-  @retval EFI_INVALID_PARAMETER   The flat translation table has not been built or
-                                  Attributes was NULL or Length was 0
-  @retval EFI_NOT_FOUND           The input region could not be found
-  @retval EFI_NO_MAPPING          The access attributes are not consistent across the region.
+  table entries to EFI access attributes (EFI_MEMORY_XP, EFI_MEMORY_RO, EFI_MEMORY_RP). The caller
+  of this function is responsible for checking ActualCheckedLength if the return value is
+  EFI_NOT_FOUND or EFI_NO_MAPPING. EFI_NOT_FOUND indicates that the attributes vary across
+  the region. EFI_NO_MAPPING indicates that the section from RegionStart to RegionStart +
+  ActualCheckedLength is not mapped. This implies that if ActualCheckedLength == RegionLength,
+  the entire input region is not mapped.
```
os-d:
```suggestion
  the region. EFI_NO_MAPPING indicates that the section from RegionStart to RegionStart +
  ActualCheckedLength is not mapped. If ActualCheckedLength == RegionLength,
  when EFI_NO_MAPPING is returned, the entire input region is not mapped.
```

nit: suggesting for clarity and a stronger statement

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
```diff
@@ -405,113 +428,171 @@ PopulateEfiMemoryMap (
...
 /**
-  Checks the input flat page/translation table for the input region and converts the associated
-  table entries to EFI access attributes.
+  Checks the input flat page/translation table for the input region and validates
+  the attributes match the input attributes.
 
   @param[in]  Map                 Pointer to the PAGE_MAP struct to be parsed
   @param[in]  Address             Start address of the region
   @param[in]  Length              Length of the region
   @param[out] Attributes          EFI Attributes of the region
```
os-d: These are not the right parameters any longer

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
```diff
@@ -1310,36 +1303,38 @@ ImageCodeSectionsRoDataSectionsXp (
...
+               SectionEnd - SectionStart,
+               EFI_MEMORY_RO,
+               FALSE,
+               FALSE,
+               FALSE
+               ))
+        {
           UT_LOG_ERROR (
-            "Image %a: Section 0x%llx-0x%llx is not EFI_MEMORY_RO\n",
+            "Image %a: Section 0x%llx-0x%llx is not EFI_MEMORY_XP\n",
```
os-d: This should still be RO, right?

os-d (approved):

UefiTestingPkg/Include/Library/FlatPageTableLib.h
TaylorBeebe: Updated

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
TaylorBeebe: Updated the input parmeters

UefiTestingPkg/AuditTests/PagingAudit/UEFI/Dxe/App/DxePagingAuditTestApp.c
TaylorBeebe: Yes, updated to print EFI_MEMORY_RO

kuqin12 (approved):

PR #400, opened 2024-01-04
DxePagingAudit: Skip Stack Publishing if Stack Info Isn't Valid

TaylorBeebe:
## Description

The MemoryInfoDatabase.dat file contains information about the content of memory regions. If an entry in the database file is invalid, it will break the generation of the paging audit. Before adding the stack info to the database file, check that the stack size is greater than zero.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 by generating the paging audit

## Integration Instructions

N/A

### DIFF
--- a/UefiTestingPkg/AuditTests/PagingAudit/UEFI/PagingAuditCommon.c
+++ b/UefiTestingPkg/AuditTests/PagingAudit/UEFI/PagingAuditCommon.c
@@ -1281,14 +1281,16 @@ SpecialMemoryDump (
       }
 
       // Capture the stack
-      AsciiSPrint (
-        TempString,
-        MAX_STRING_SIZE,
-        "Stack,0x%016lx,0x%016lx\n",
-        StackBase,
-        StackLength
-        );
-      AppendToMemoryInfoDatabase (TempString);
+      if (StackLength > 0) {
+        AsciiSPrint (
+          TempString,
+          MAX_STRING_SIZE,
+          "Stack,0x%016lx,0x%016lx\n",
+          StackBase,
+          StackLength
+          );
+        AppendToMemoryInfoDatabase (TempString);
+      }
 
       break;
     }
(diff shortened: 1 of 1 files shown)


UefiTestingPkg/AuditTests/PagingAudit/UEFI/PagingAuditCommon.c
```diff
@@ -1281,14 +1281,16 @@ SpecialMemoryDump (
...
-        );
-      AppendToMemoryInfoDatabase (TempString);
+      if (StackLength > 0) {
+        AsciiSPrint (
+          TempString,
+          MAX_STRING_SIZE,
+          "Stack,0x%016lx,0x%016lx\n",
+          StackBase,
+          StackLength
+          );
```
os-d, line 1289: These prints would still work with a zero size stack, right? Presumably the report generation code chokes on the zero size stack? If so, perhaps it is still valuable to know the stack wasn't populated and fix the report generation code. Unless I'm missing why this breaks here?

Which leads me to, in what cases do we have a zero length stack? Do some of these stacks not exist on some platforms?

TaylorBeebe: The report generation code will break in various ways if any of the input data is nonsense. There are definitely ways to handle weird cases like this one with try-catch blocks and data auditing, but because the DxePagingAudit is both the producer and consumer of this data, fixing it on either side seems valid. In this case, it's much easier to fix it on the producer side.

Yes, we do have platforms which have nonsense in the memory descriptor HOBs. I'll avoid calling out which ones here but I'm sure you can guess.

os-d (approved):

makubacki (approved):

PR #408, opened 2024-01-17
Wrap advanced logger buffer cursor when the logging area is full

kuqin12:
## Description

The advanced logging at runtime is lack of usage. However, this becomes increasingly important for server platforms, where the system could potentially stay up for months, or even years. This will make the advanced logger fill up the buffer during runtime/MM logging well ahead of a reset event, making the buffer content stale when being reviewed.

This change added a feature PCD guarded feature that, when enabled, will automatically wrap the LogCurrent cursor to the beginning and continue to log. This will effectively create a circular buffer as the stale content will remain in place. The tooling update is still under development.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU Q35 along with the UEFI shell based test app to verify that the entry after wrapping is correct and is multi-threading safe.

## Integration Instructions

Platforms can enable this feature by setting `gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerAutoClearEnable|TRUE`


AdvLoggerPkg/AdvLoggerPkg.dec
```diff
@@ -81,6 +81,10 @@
   #
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedFileLoggerForceEnable|TRUE|BOOLEAN|0x00010184
 
+  ## PcdAdvancedFileLoggerForceEnable - Forces the creation of the Logs subdirectory on non USB devices
```
os-d: I don't think this is what this does :)

AdvLoggerPkg/AdvLoggerPkg.dec
```diff
@@ -81,6 +81,10 @@
   #
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedFileLoggerForceEnable|TRUE|BOOLEAN|0x00010184
 
+  ## PcdAdvancedFileLoggerForceEnable - Forces the creation of the Logs subdirectory on non USB devices
+  #
+  gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerAutoClearEnable|FALSE|BOOLEAN|0x00010188
```
os-d: Maybe this should be AutoWrap instead of AutoClear?

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
```diff
@@ -69,20 +69,42 @@ AdvancedLoggerMemoryLoggerWrite (
...
+                              (UINT64 *)&LoggerInfo->LogCurrent,
+                              (UINT64)CurrentBuffer,
+                              (UINT64)NewBuffer
+                              );
+          } while (OldValue != CurrentBuffer);
+
+          // Now that we have a buffer that starts from the beginning, proceed to log the current message, from the beginning.
+          // Note that in this case, if there are other threads in the middle of logging a message, they will write to the lala
+          // land where we will just drop the content.
+          // If there is another clearing attempt on the other thread, i.e. through the clear variable command, the later clear
```
kuqin12: No more clear command, but this simultaneous clear could still happen.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
```diff
@@ -69,20 +69,42 @@ AdvancedLoggerMemoryLoggerWrite (
...
+          // will take effect and the log entry together with that clear will be dropped.
```
os-d: There is no clear command, right? Just this wrapping? Or will clear still be supported?

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
```diff
@@ -69,20 +69,42 @@ AdvancedLoggerMemoryLoggerWrite (
...
+            //
+            CurrentBuffer = LoggerInfo->LogCurrent;
+
+            NewBuffer     = LoggerInfo->LogBuffer;
+            OldValue      = InterlockedCompareExchange64 (
+                              (UINT64 *)&LoggerInfo->LogCurrent,
+                              (UINT64)CurrentBuffer,
+                              (UINT64)NewBuffer
+                              );
+          } while (OldValue != CurrentBuffer);
```
os-d: If I am understanding correctly, this is a do-while loop to catch the case where `LogCurrent` is updated between our read and the CmpExch. But, if it was updated, wouldn't we not want to update `LogCurrent` then, because something else updated it, presumably back to `LogBuffer` and we would overwrite new messages?

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
kuqin12: aye, you got to it first :) i added the comment and did not submit :)

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
kuqin12: My original thought was to just let the last wrap take precedence. But I think we can leave the re-attempt to the outer loop. So that if this update failed, we can recalculate the pointers and see if the message will just fit in? If not, this logic can engage again?

os-d: That makes sense to me. If there were a case of very noisy neighbor threads who were spamming the log before this could write in, the platform would need to fix those neighbors or increase the log buffer size, but I think that should be a remote enough edge case we don't need to worry about getting stuck here.

kuqin12: Okay, it is updated. Can you please check to see if that does the job to let go the first wrap and keep the log messages?

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
kuqin12: Updated.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
kuqin12: Updated.

AdvLoggerPkg/AdvLoggerPkg.dec
kuqin12: Updated.

AdvLoggerPkg/AdvLoggerPkg.dec
kuqin12: Indeed :)

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
```diff
@@ -69,20 +69,44 @@ AdvancedLoggerMemoryLoggerWrite (
...
-                          );
-        } while (OldSize != CurrentSize);
-
-        return LoggerInfo;
+        if (FeaturePcdGet (PcdAdvancedLoggerAutoWrapEnable) && (LoggerInfo->AtRuntime)) {
+          //
+          // Wrap around the current cursor when auto wrap is enabled on buffer full during runtime.
+          //
+          NewBuffer = LoggerInfo->LogBuffer;
+          OldValue  = InterlockedCompareExchange64 (
```
os-d, line 77: Hmm, as I look at this now, this looks like it will work, but the flow is do CmpExch, reset to the beginning, then do another CmpExch and move LogCurrent to the end of our message. Should we not do this all in one atomic step, check if we are going to be going over the end and if so, set LogCurrent to be LogBuffer + MsgSize?

os-d: This is also only when we hit the end, which should be fairly infrequent, so having the duplicate call is okay, I think, where we wrap the pointer, then come back and try again for it.

os-d (approved):

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
kuqin12: Oh, you mean just do `CurrentBuffer = LoggerInfo->LogBuffer;` here? That seems better and should work.

AdvLoggerPkg/AdvLoggerPkg.dec
```diff
@@ -81,6 +81,10 @@
   #
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedFileLoggerForceEnable|TRUE|BOOLEAN|0x00010184
 
+  ## PcdAdvancedLoggerAutoWrapEnable - Automatically wrap around the LogCurrent cursor when it reaches the end of the log
```
apop5: We should mention here that this only occurs at runtime (after exit boot service even is triggered)

apop5 (approved):

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
kuqin12: Hmm.. Simply `CurrentBuffer = LoggerInfo->LogBuffer;` will not work as the `InterlockedCompareExchange64` below will not go through because CurrentBuffer does not match current LoggerInfo->LogCurrent. So either we invoke CmpExch twice, or update the existing `InterlockedCompareExchange64` call below to plum through logic.

os-d: Actually, I realized we cannot do this, because the read of `LogCurrent` may be stale, so if we always assume we are resetting, someone else may have done it already and we reset and write over valid stuff.

AdvLoggerPkg/AdvLoggerPkg.dec
kuqin12: Updated.

PR #411, opened 2024-01-20
Add call to HdwPortInitialize() when instantiating logger in DXE

joschock:
## Description

This change adds a call to AdvancedLoggerHdwPortInitialize() when the advanced logger is created in the DxeCoreAdvancedLoggerLibConstructor(). Previously, hardware port was only initalized in AdvancedLoggerGetInfo() if the logging info structure was present at a fixed location or in a HOB; this would cause writes to the hardware port to potentially fail in the path where the logger is not initialized until the Constructor is called. 

- [x] Impacts functionality?
  - AdvancedLoggerHdwPortInitialize() is now called in the DxeCoreAdvancedLoggerLibConstructor if the logger is instantiated there.
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified on a platform where this was not working due to missing HdwPort initialization; verified after this patch that it works as expected. 

## Integration Instructions

N/A

### DIFF
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
@@ -384,10 +384,12 @@ DxeCoreAdvancedLoggerLibConstructor (
       LoggerInfo->LogBufferSize = EFI_PAGES_TO_SIZE (FixedPcdGet32 (PcdAdvancedLoggerPages)) - sizeof (ADVANCED_LOGGER_INFO);
       LoggerInfo->LogCurrent    = LoggerInfo->LogBuffer;
       LoggerInfo->HwPrintLevel  = FixedPcdGet32 (PcdAdvancedLoggerHdwPortDebugPrintErrorLevel);
-      AdvancedLoggerHdwPortInitialize ();
-      mLoggerInfo->HdwPortInitialized = TRUE;
-      mMaxAddress                     = LoggerInfo->LogBuffer + LoggerInfo->LogBufferSize;
-      mBufferSize                     = LoggerInfo->LogBufferSize;
+      if (LoggerInfo->HdwPortInitialized == FALSE) {
+        AdvancedLoggerHdwPortInitialize ();
+        LoggerInfo->HdwPortInitialized = TRUE;
+      }
+      mMaxAddress = LoggerInfo->LogBuffer + LoggerInfo->LogBufferSize;
+      mBufferSize = LoggerInfo->LogBufferSize;
     } else {
       DEBUG ((DEBUG_ERROR, "%a: Error allocating Advanced Logger Buffer\n", __FUNCTION__));
     }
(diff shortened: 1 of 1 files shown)


kuqin12 (approved):

kuqin12:
Can we target this to release/2311 and then cherry-pick to release/2302?

joschock:
> Can we target this to release/2311 and then cherry-pick to release/2302?

I'm good with that.

joschock:
> > Can we target this to release/2311 and then cherry-pick to release/2302?
> 
> I'm good with that.

Rebase to 202311 completed.

apop5 (approved):

AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
```diff
@@ -384,8 +384,10 @@ DxeCoreAdvancedLoggerLibConstructor (
       LoggerInfo->LogBufferSize = EFI_PAGES_TO_SIZE (FixedPcdGet32 (PcdAdvancedLoggerPages)) - sizeof (ADVANCED_LOGGER_INFO);
       LoggerInfo->LogCurrent    = LoggerInfo->LogBuffer;
       LoggerInfo->HwPrintLevel  = FixedPcdGet32 (PcdAdvancedLoggerHdwPortDebugPrintErrorLevel);
-      mMaxAddress               = LoggerInfo->LogBuffer + LoggerInfo->LogBufferSize;
-      mBufferSize               = LoggerInfo->LogBufferSize;
+      AdvancedLoggerHdwPortInitialize ();
```
makubacki: Should this call be conditionalized on `!mLoggerInfo->HdwPortInitialized`?

The code obviously depends on `LoggerInfo` presence to implicitly determine if the port was initialized above but given that `mLoggerInfo->HdwPortInitialized` appears to be an explicit guard around the port being initialized more than once, it would be a bit more robust to guard the call with it regardless of other dependencies like the implicit side effect/return value from `AdvancedLoggerGetLoggerInfo()`.

```suggestion
      if (!mLoggerInfo->HdwPortInitialized) {
        AdvancedLoggerHdwPortInitialize ();
      }
```

makubacki (approved):

AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
joschock: Well, the main reason I didn't do that is because LoggerInfo was just created and zero'd immediately prior, so there is no code path where LoggerInfo->HdwPortInitialized would be true - making the conditional redundant. I can add it if you think that is more self-describing though. Also - at this point mLoggerInfo is probably null, so the conditional would need to be on LoggerInfo anyway.

makubacki: Within the current context I agree it's not needed. But the flag really is associated with that call so I think it should always surround the call. People change code without being aware of all the things defined elsewhere, copy/paste calls elsewhere and don't realize associated flags otherwise existed, etc. So, this is more a "defensive programming" suggestion.

joschock: Ok, added the conditional. Also fixed the mLoggerInfo reference which should have been a LoggerInfo reference.

PR #415, opened 2024-01-23
Adding a more adapt python tooling for Advanced Logger v4

### LINKED ISSUE #401: [Feature]: AdvLoggerPkg should never discard messages
### Feature Overview

Once the buffer used for the AdvLoggerPkg is filled, any new messages are discarded. This make it impossible to get actionable logs on a long running system.

### Solution Overview

The buffer should either be flushed to persistent storage when it is full / periodically or it should at least be circular so that the most recent (& likely most relevant) logs are always available.

### Alternatives Considered

_No response_

### Urgency

High

Someone else needs to implement the feature

No maintainer feedback needed

### Anything else?

_No response_

kuqin12:
# Preface

## Description

This change will allow the advanced logger parser to handle the v4, which would get lines from the same UEFI boot phase without intercepting each other. It will also search for the beginning of returned buffer with more sophisticated logic.

Resolves https://github.com/microsoft/mu_plus/issues/401

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on both QEMU Q35 and proprietary hardware platform returned advanced logger buffer binary.

## Integration Instructions

N/A


AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
```diff
@@ -456,9 +456,6 @@ def _GetNextMessageBlock(self, LoggerInfo):
         if LoggerInfo["LogBuffer"] == LoggerInfo["LogCurrent"]:
             return (self.END_OF_FILE, MessageBlock)
 
-        if InFile.tell() >= LoggerInfo["LogCurrent"]:
```
kuqin12: Need to double check backwards compat on this.

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
```diff
@@ -598,22 +595,123 @@ def _GetLines(self, lines, LoggerInfo):
...
         StartLine = LoggerInfo["StartLine"]
 
-        while (Status == self.SUCCESS):
-            (Status, MessageLine) = self._GetNextFormattedLine(MessageLine, LoggerInfo)
-            if Status != self.SUCCESS:
-                if Status != self.END_OF_FILE:
-                    print(f"Error {Status} from GetNextFormattedLine")
-                break
+        if LoggerInfo["Version"] == self.V4_LOGGER_INFO_VERSION:
+            # There is a potential that we may have enabled auto wrap.
```
Javagedes: I think the main suggestion is there is a lot going on here that makes it hard / slow to understand for soneone new to this script. I would break all of this into a few distinct functions and then provide documentation for each function. This logic really does three things. 1. Find the first uncorrupted block, 2. Generate the list of messages, 3. Turn messages into strings

1. Turn the logic that finds the offset of the first uncorrupted block of records into it's own function.

`StartOffset = FirstValidBlock(LoggerInfo)`

2. Turn the logic that builds the list of Message Entries into it's own function.

`MessageEntries = ParseMessageEntries(LoggerInfo, StartOffset)`

3. Convert the Message entries into a list of lines

`Lines = BuildLinesFromMessages(MessageEntries)`

It does not have to be exactly this, but by breaking it up, and then documenting each function, I think it will make it easier for someone new to understand what is going on.

Javagedes:
Hi Kun - Overall, this looks good. I tried working on some performance improvements in regard to turning records into lines, as it has an inner loop I was concerned about. However, in a real scenario, the number of records that don't end in a "\n" will be far and few between (and even when they do occur, there will only be a few records between that and the break condition), so the performance impact is pretty negligible in my opinion.

I left some comments on your comments in regard to splitting the overall logic into three separate functions, to make understanding the logic a bit easier.

Overall, I don't see any glaring issues.

Javagedes (approved):

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
```diff
@@ -598,22 +595,123 @@ def _GetLines(self, lines, LoggerInfo):
...
+            # If so, we need to find the first legible line. Given that
+            # we read the logger information from the head of the buffer,
+            # we can start from this cursor as an acceptable estiamted
```
os-d:
```suggestion
            # we can start from this cursor as an acceptable estimated

```

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
```diff
@@ -598,22 +595,123 @@ def _GetLines(self, lines, LoggerInfo):
...
+            # If we need to wrap, we need to start from the beginning.
+            if NeedWrap:
+                CurrentOffset = OriginalOffset
+                MaximumOffset = LoggerInfo["LogCurrent"]
+                LoggerInfo["InFile"].seek(CurrentOffset)
+                while CurrentOffset < MaximumOffset:
+                    (MessageEntry, _) = self._ReadMessageEntry(LoggerInfo)
+                    CurrentOffset = LoggerInfo["InFile"].tell()
+                    if MessageEntry is None:
+                        break
```
os-d: If we hit this case, this would be an error, right? Because we had a log where we wrapped, but had garbage before LogCurrent, i.e. the most recent message(s) got garbled. We should probably at least raise a warning here, something along the lines of, I tried to parse the log, but the most recent messages have been lost, here is all I could figure out

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
```diff
@@ -598,22 +595,123 @@ def _GetLines(self, lines, LoggerInfo):
...
+                while CurrentOffset < LoggerInfo["LogBufferSize"] and\
+                        CurrentOffset < StartOffset + self.MAX_MESSAGE_SIZE:
+                    (MessageEntry, _) = self._ReadMessageEntry(LoggerInfo)
+                    CurrentOffset = LoggerInfo["InFile"].tell()
+                    if MessageEntry is None:
+                        ThisIsGood = False
+                        break
+
+                if ThisIsGood:
+                    LoggerInfo["InFile"].seek(StartOffset)
```
os-d: This logic was somewhat confusing to me, can we add a comment here. This case I believe is we found a valid message after LogCurrent, so we are in the wrap case, right?

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
```diff
@@ -598,22 +595,123 @@ def _GetLines(self, lines, LoggerInfo):
...
+                        if StartOffset == -1:
+                            print("DecodeUefiLog unable to find ALM* signature. Using the beginning as start")
+                            StartOffset = OriginalOffset
+                            break
+                        else:
+                            StartOffset += CurrentOffset
+                    else:
+                        StartOffset += CurrentOffset
+
+            # By the time we are here, we LoggerInfo["InFile"] pointing at the first StartOffset.
```
os-d:
```suggestion
            # By the time we are here, we have LoggerInfo["InFile"] pointing at the first StartOffset.

```

os-d (approved):
We now have many conditions in this script, I suspect through more testing we'll find something that didn't quite work out in one of the odd cases, but as read it makes some sense to me :)

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
kuqin12: This is saying that after the `LogCurrent`, we have found a starting point where the content has been good for at least an entire block (512 bytes). This is to prevent a memory writing across the block boundary during block readings.

We always assume we are in the wrap case and start to find the potential starting point. But if we have not identified a block that is fully formatted after `LogCurrent` till the end of the logging region, it means that we are not in the wrap case.

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
kuqin12: Updated.

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
kuqin12: Reverted,

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
kuqin12: Updated to closer to your recommendation. Thanks for the suggestion!

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
kuqin12: I updated the flow as requested by Joey. Hopefully this is clearer now,

kuqin12:
> Hi Kun - Overall, this looks good. I tried working on some performance improvements in regard to turning records into lines, as it has an inner loop I was concerned about. However, in a real scenario, the number of records that don't end in a "\n" will be far and few between (and even when they do occur, there will only be a few records between that and the break condition), so the performance impact is pretty negligible in my opinion.
> 
> I left some comments on your comments in regard to splitting the overall logic into three separate functions, to make understanding the logic a bit easier.
> 
> Overall, I don't see any glaring issues.

Thanks. I did update this change as you requested. I agree that this looks much cleaner! I will try to merge this change if possible and then we can work to make this better on a needed base.

PR #432, opened 2024-02-06
SecureBootRecovery: Add Logic to detect 2023 Windows Production CA

Flickdm:
## Description

This change adds the ability for the SecureBootRecovery application to detect the existence of the 2023 Windows Production CA and prevent updating the DB if found.

- [X] Impacts functionality?
  - **Functionality** - Prevents updating the DB if the existence of the 2023 Windows CA is found
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This application was self signed on a production system using custom Secure Boot keystore. 

This was tested with:

No Secure boot Certificates
Only the 2011 Certificates
All Certificates (2011 and 2023)

## Integration Instructions

N/A


MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
```diff
@@ -48,6 +97,210 @@ StatusToHexString (
...
+
+  //
+  // Check that the DB has the expected attributes
+  //
+  if ((Attributes & VARIABLE_ATTRIBUTE_NV_BS_RT_AT) != VARIABLE_ATTRIBUTE_NV_BS_RT_AT) {
+    Status = EFI_UNSUPPORTED;
+    //
+    // Likely this will continue to fail on reboot
+    //
+    return FALSE;
```
apop5: should EFI_UNSUPPORTED get bubbled up through OptionalStatus?

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
```diff
@@ -48,6 +97,210 @@ StatusToHexString (
...
+                  );
+
+  //
+  // Confirm we got the expected error that the buffer was too small
+  //
+  if (Status != EFI_BUFFER_TOO_SMALL) {
+    //
+    // Likely this will continue to fail on reboot
+    //
+    return FALSE;
```
apop5: Should EFI_BUFFER_TOO_SMALL get bubbled up through OptionalStatus?

apop5 (approved):

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
```diff
@@ -48,6 +97,210 @@ StatusToHexString (
...
+
+  // Confirm the DB is at least as large as the EFI_SIGNATURE_LIST header
+  if (DbSize < sizeof (EFI_SIGNATURE_LIST)) {
+    Status = EFI_BUFFER_TOO_SMALL;
+    goto Exit;
+  }
+
+  //
+  // The DB is a list of EFI_SIGNATURE_LISTs but we only care about certificates
+  // they are not garunteed to be in any particular order
```
kenlautner: Nit: gauranteed

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
kenlautner: Shouldn't this be `goto Exit`?

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
kenlautner: Same as above.

kenlautner (approved):

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
Flickdm: This is actually the expected error, since I'm giving it a NULL buffer with a size of 0. It in return tells me the full size that I can then use to allocate. Are you asking if I should replace EFI_BUFFER_TOO_SMALL with EFI_SUCCESS after this check?

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
```diff
@@ -11,15 +11,64 @@
...
+#define CERT_COMMON_NAME_OFFSET   0x116
+#define CERT_SIGNATURE            { 0x9f, 0xc9, 0xb6, 0xff, 0x6e, 0xe1, 0x9c, 0x3b, 0x55, 0xf6, 0xfe, 0x8b, 0x39, 0xdd, 0x61, 0x04 }
+#define CERT_SIGNATURE_OFFSET     0x3AE
+
+//
+// The system table pointers
+//
+EFI_SYSTEM_TABLE      *mST = NULL;
+EFI_RUNTIME_SERVICES  *mRT = NULL;
+EFI_BOOT_SERVICES     *mBS = NULL;
```
makubacki: Why not just use UefiBootServicesTableLib and UefiRuntimeServicesTableLib? They maintain the pointers for you.

I understand dependencies are trying to be reduced but you're already dependent on much more complex logic elsewhere (like the functions backing those pointers) to the point the code redundancy vs complexity tradeoff I think favors just using the MdePkg library instances.

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
```diff
@@ -11,15 +11,64 @@
...
+//
 #define STALL_10_SECONDS  10000000
 
+//
+// The size of an EFI_STATUS in bytes
+//
 #define STATUS_SIZE         (sizeof(EFI_STATUS) * 2)
 #define STATUS_STRING_SIZE  (STATUS_SIZE + sizeof(L"\0"))
 
+//
```
makubacki: I suggest moving some of these macros and definitions to a header file.

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
```diff
@@ -11,15 +11,64 @@
...
+
+//
+// The system table pointers
+//
+EFI_SYSTEM_TABLE      *mST = NULL;
+EFI_RUNTIME_SERVICES  *mRT = NULL;
+EFI_BOOT_SERVICES     *mBS = NULL;
+
+typedef struct {
+  UINT8    *Organization;      // The organization name in the certificate
```
makubacki: I don't know the details of this string's contents, but was there a particular reason for using `UINT8` vs `CHAR8`?

makubacki: It looks like it is expected to be an ASCII string like `"Microsoft Corporation"`? In that case, I'd recommend `CHAR8`.

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
```diff
@@ -48,6 +97,210 @@ StatusToHexString (
...
+
+  @retval TRUE             The certificate was found in the database.
+  @retval FALSE            The certificate was not found in the database.
+                           If OptionalStatus is not NULL, it contains an error code indicating the reason for the failure.
+
+**/
+BOOLEAN
+IsCertificateInDB (
+  IN  CERTIFICATE_ENTRY  *Certificate,
+  OUT EFI_STATUS         *OptionalStatus
```
makubacki:
```suggestion
  OUT EFI_STATUS         *OptionalStatus OPTIONAL

```

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
```diff
@@ -48,6 +97,210 @@ StatusToHexString (
...
+ * Prints an error message to the console
+ * @param[in] Message     The message to print
+ * @param[in] Status      The EFI_STATUS to print
+*/
+VOID
+PrintError (
+  CHAR16      *Message,
+  EFI_STATUS  Status
+  )
+{
```
makubacki, line 70: I suggest at least an assert to simplify debugging in case of a bad value given to the function.

```suggestion
{
  ASSERT (Message != NULL);
```

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
```diff
@@ -48,6 +97,210 @@ StatusToHexString (
...
+  //
+  // If we get here, we did not find the certificate in the DB, however, we did not encounter any errors
+  //
+  Status = EFI_SUCCESS;
+
+Exit:
+
+  if (OptionalStatus != NULL) {
+    *OptionalStatus = Status;
+  }
```
makubacki: If I understand the flow correctly, `Db` is no longer needed and the buffer can be freed (on success or through an error path).

```suggestion
Exit:

  if (Db != NULL) {
    mBS->FreePool (Db);
  }

  if (OptionalStatus != NULL) {
    *OptionalStatus = Status;
  }

```

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
```diff
@@ -48,6 +97,210 @@ StatusToHexString (
...
+        Cert     = CertData->SignatureData;
+        CertSize = CertList->SignatureSize - sizeof (EFI_GUID);
+
+        //
+        // First lets ensure that the size of the certificate is at least as large as the certificate we are looking for
+        // Each offset in the certficate entry plus the size of the string must be less than the size of the certificate
+        //
+        if (((Certificate->OrganizationOffset + sizeof (Certificate->Organization)) <= CertSize) &&
+            ((Certificate->CommonNameOffset + sizeof (Certificate->CommonName)) <= CertSize) &&
+            ((Certificate->SignatureOffset + sizeof (Certificate->Signature)) <= CertSize))
```
makubacki: In `sizeof (Certificate->Organization)` are you trying to get the size of the struct field (i.e. the pointer) or the size of the string itself?

At a quick glance, it seems that you're trying to get the size of the string here. In that case, `AsciiStrSize()` would be needed to get the size including the NULL character.

This is also where making `Organization` being a `CHAR8` makes it compatible with string functions.

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
makubacki: If you're trying to completely avoid using `DebugLib`, then just return on NULL.

makubacki (approved):
Approved assuming noted issues are resolved.

I suggest pinging reviewers when all updates are done so we can take a quick second look before merging.

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
Flickdm: Since this is going to be consumed by the Boot team and built with windows tooling instead of UEFI ultimately, I'm avoiding the additional libraries

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
Flickdm: This is correct, somehow this got dropped when I was rebasing commits trying to clean up the commits. Thanks!

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
Flickdm: This is getting the size of the string itself. Although in this example I have OrganizationLength and CommonNameLength and don't need to perform the size_of again.

Since the strings lengths are really just a char array which the size of is known at compile time I figured I could avoid pulling in an additional library to get AsciiStrSize(..) and just use str_len

Further, ASN.1 strings (that we're looking for) aren't null terminated and thus we just need the length of the array

makubacki: `OrganizationLength` and `CommonNameLength` are fine to use because they computed the string length using `sizeof` on a byte-sized string literal.

`sizeof (Certificate->Organization)` is the size of the `Organization` member which is `UINT8*` (pointer is 8 bytes on 64-bit machine). `AsciiStrLen()` could be used to not include the null character. But, as you said you can just use the ready-to-go values.

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
makubacki: In that case, I suggest leaving a note saying why this is done since it is against typical edk2 development patterns.

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
Flickdm: I tried breaking up the macros and definitions and I wasn't a fan of breaking it immediately. I might leave this to boot team to figure out how they want to organize it.

Flickdm: Moved the certificate defines to a different header file, like you suggested. With a proper comment I like it.

makubacki: Yes, it makes it easier to focus on implementation in the C file and quickly find data definitions in the header file.

MsApplicationPkg/SecureBootRecovery/SecureBootRecovery.c
```diff
@@ -48,6 +55,233 @@ StatusToHexString (
...
+  // The DB is a list of EFI_SIGNATURE_LISTs but we only care about certificates
+  // they are not gauranteed to be in any particular order
+  //
+  CertList = (EFI_SIGNATURE_LIST *)Db;
+
+  //
+  // Iterate through the DB entries
+  // The DB is a list of EFI_SIGNATURE_LISTs
+  //
+  while ((DbSize > 0) && (DbSize >= CertList->SignatureListSize)) {
```
makubacki, line 214: Because `DbSize` is an unsigned integer type (`UINTN`), it is always going to be greater than or equal to zero. If you're going to subtract and expect subtraction to potentially go below zero, then this needs to be an `INTN`. If zero is very certain regardless of user input, etc. then this is fine.

Flickdm: 0 should be very certain regardless of the user input because it's coming from the GetVariable(..) api. This would mean that the platform is wildly out of spec and I don't think I can handle that.

Issue #434, opened 2024-02-07
[Bug]: AdvLoggerPkg and probably other are not built for AARCH64 in CI


cfernald:
### Is there an existing issue for this?

- [X] I have searched existing issues

### Current Behavior

Currently AdvLoggerPkg and probably others are not built in CI with AARCH64 because of the SUPPORTED_ARCHITECTURES, despite having AARCH64 specific code that needs CI.

### Expected Behavior

Package that are used for AARCH64 should have CI builds. 

### Steps To Reproduce

N/A

### Build Environment

```markdown
- OS(s): Ubuntu
- Tool Chain(s): GCC
- Targets Impacted: CI
```

### Version Information

```text
N/A
```

### Urgency

Medium

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

PR #435, opened 2024-02-07
Switch MmCoreArm advanced logger to use updated MMU functions

cfernald:
## Description

As part of 202311, ArmMmuLib switched to use ArmSetMemoryAttributes instead of the individual set/clear routines. This was also implemented in the MU change to converge ArmMmuStandaloneMmLib's use with ArmMmuLib. This changed fixes MmCoreArm/AdvancedLoggerLib.c to use the updated functions. 

It seems this went unnoticed originally due to issue #434 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Manually tested.

## Integration Instructions

N/A

### DIFF
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
@@ -15,7 +15,7 @@
 #include <Library/BaseMemoryLib.h>
 #include <Library/DebugLib.h>
 #include <Library/SynchronizationLib.h>
-#include <Library/ArmMmuLib.h>
+#include <Library/MmuLib.h>
 
 #include "../AdvancedLoggerCommon.h"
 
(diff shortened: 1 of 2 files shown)


os-d (approved):
The switch looks fine, that being said we don't check the return Status, probably worthwhile to have a check in there to either assert or log to the hw port (if available)

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -114,12 +114,20 @@ AdvancedLoggerGetLoggerInfo (
     // If this is the first time for MM core to get here, the memory attributes of this module
     // may not be fully set yet. Thus set the memory for global variables attributes to RW first.
     Address = ALIGN_VALUE ((EFI_PHYSICAL_ADDRESS)(UINTN)&mInitialized - EFI_PAGE_SIZE + 1, EFI_PAGE_SIZE);
-    Status  = ArmSetMemoryRegionNoExec (Address, EFI_PAGE_SIZE);
-    Status  = ArmClearMemoryRegionReadOnly (Address, EFI_PAGE_SIZE);
+    Status  = ArmSetMemoryAttributes (
```
kuqin12: I think we should better use `MmuSetAttributes (Address, EFI_PAGE_SIZE, EFI_MEMORY_XP)` here. And include the "<Library/MmuLib.h>" instead.

Also, we also need to clear the read only flag, right?

kuqin12 (approved):

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
cfernald: I agree on using MmuLib, to get off ARM specific libs, Will make that change.

Side rant, I'm not a fan of this whole "clear/set" semantic when it comes to attributes. A caller will always want/need a specific set of attributes for the use case of the buffer, so it sees really odd to me to allow a caller to only ask about specific attributes. I believe they should set to an explicit combination each time. _steps off soapbox_

kuqin12: I think this is to prevent the caller always need to retrieve the current attribute? But I agree that this pattern is confusing when we want to replace the current page attributes with ours.

cfernald: My point is, why do we care about the current attributes? If we have a buffer for a purpose, we know what all 3 attributes should be, so we should set them so what we want despite what they currently are. but yeah, not relevant here.

PR #437, opened 2024-02-09
Remove AdvancedLogger MmCoreArm dependence on global variables.

cfernald:
## Description

The Advanced logger Arm MM Core library may be invoked before the imagine has completed it's PeCOFF section attribute fixup. Previously this was handled by explicitly mapping the data page need for the adv logger globals to be writable, but this leads to potential infinite recursion if any of the functions in the MMU stack calls a print. Instead this change is to remove the dependence on global variables and so remove the need for external calls in these library functions.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested and validated presence of MM logs in serial output and adv logger buffer.

## Integration Instructions

N/A

### DIFF
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
@@ -15,63 +15,15 @@
 #include <Library/BaseMemoryLib.h>
 #include <Library/DebugLib.h>
 #include <Library/SynchronizationLib.h>
-#include <Library/MmuLib.h>
 
 #include "../AdvancedLoggerCommon.h"
 
 #define ADV_LOGGER_MIN_SIZE  (65536)
 
-STATIC ADVANCED_LOGGER_INFO  *mLoggerInfo = NULL;
-STATIC UINT32                mBufferSize  = 0;
-STATIC EFI_PHYSICAL_ADDRESS  mMaxAddress  = 0;
-STATIC BOOLEAN               mInitialized = FALSE;
-
-/**
-  Validate Info Blocks
-
-  The address of the ADVANCE_LOGGER_INFO block pointer is captured during the first debug print.  The
-  pointers LogBuffer and LogCurrent, and LogBufferSize, could be written to by untrusted code.  Here,
-  we check that the pointers are within the allocated mLoggerInfo space, and that LogBufferSize, which
-  is used in multiple places to see if a new message will fit into the log buffer, is valid.
-
-  @param          NONE
-
-  @return         BOOLEAN     TRUE = mLoggerInfo Block passes security checks
-  @return         BOOLEAN     FALSE= mLoggerInfo Block failed security checks
-
-**/
-STATIC
-BOOLEAN
-ValidateInfoBlock (
-  VOID
-  )
-{
-  if (mLoggerInfo == NULL) {
-    return FALSE;
-  }
-
-  if (mLoggerInfo->Signature != ADVANCED_LOGGER_SIGNATURE) {
-    return FALSE;
-  }
-
-  if (mLoggerInfo->LogBuffer != (PA_FROM_PTR (mLoggerInfo + 1))) {
-    return FALSE;
-  }
-
-  if ((mLoggerInfo->LogCurrent > mMaxAddress) ||
-      (mLoggerInfo->LogCurrent < mLoggerInfo->LogBuffer))
-  {
-    return FALSE;
-  }
-
-  if (mBufferSize == 0) {
-    mBufferSize = mLoggerInfo->LogBufferSize;
-  } else if (mLoggerInfo->LogBufferSize != mBufferSize) {
-    return FALSE;
-  }
-
-  return TRUE;
-}
+//
+// NO GLOBALS! This routine may run before data sections are writable, and so
+// cannot presume globals will be available.
+//
 
 /**
   The logger Information Block is carved from the Trust Zone at a specific fixed address.
@@ -87,7 +39,7 @@ ValidateInfoBlock (
   PcdAdvancedLoggerCarBase     -- NOT USED, leave at default
   PcdAdvancedLoggerPreMemPages -- NOT USED, leave at default
 
-  NOTE:  A debug statement here will cause recursion.  Insure that the recursion will be
+  NOTE:  A debug statement here will cause recursion. Ensure that the recursion will be
          a straight path just to return the existing mLoggerInfo.
 
   @param       - None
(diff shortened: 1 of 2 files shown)


AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -102,48 +54,59 @@ AdvancedLoggerGetLoggerInfo (
...
+  // Ensure the start of the log is in the correct location.
+  if (LoggerInfo->LogBuffer != (PA_FROM_PTR (LoggerInfo + 1))) {
+    return NULL;
+  }
+
+  // Make sure the size of the buffer does not overrun it's fixed size.
+  if ((LoggerInfo->LogBuffer + LoggerInfo->LogBufferSize) >
+      (FixedPcdGet32 (PcdAdvancedLoggerPages) * EFI_PAGE_SIZE))
+  {
+    return FALSE;
```
cfernald: oops

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -87,7 +39,7 @@ ValidateInfoBlock (
   PcdAdvancedLoggerCarBase     -- NOT USED, leave at default
   PcdAdvancedLoggerPreMemPages -- NOT USED, leave at default
 
-  NOTE:  A debug statement here will cause recursion.  Insure that the recursion will be
+  NOTE:  A debug statement here will cause recursion. Ensure that the recursion will be
          a straight path just to return the existing mLoggerInfo.
```
makubacki: Looks like `mLoggerInfo` here should just be "logger info" now?

```suggestion
         a straight path just to return the existing logger information.
```

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -15,63 +15,15 @@
 #include <Library/BaseMemoryLib.h>
 #include <Library/DebugLib.h>
 #include <Library/SynchronizationLib.h>
-#include <Library/MmuLib.h>
 
 #include "../AdvancedLoggerCommon.h"
 
 #define ADV_LOGGER_MIN_SIZE  (65536)
```
makubacki: I don't see `ADV_LOGGER_MIN_SIZE` used. Is it needed?

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
cfernald: Going to just update this since the reality here is bit different.

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
cfernald: Will remove. Can't think of a reason it would be needed.

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -102,48 +54,59 @@ AdvancedLoggerGetLoggerInfo (
...
-  if (((mLoggerInfo) != NULL) && !ValidateInfoBlock ()) {
-    mLoggerInfo = NULL;
-    DEBUG ((DEBUG_ERROR, "%a: LoggerInfo marked invalid\n", __FUNCTION__));
+  // Ensure the start of the log is in the correct location.
+  if (LoggerInfo->LogBuffer != (PA_FROM_PTR (LoggerInfo + 1))) {
+    return NULL;
+  }
+
+  // Make sure the size of the buffer does not overrun it's fixed size.
+  if ((LoggerInfo->LogBuffer + LoggerInfo->LogBufferSize) >
```
makubacki: Is `LoggerInfo->LogBuffer` an address? Should this just be comparing `LoggerInfo->LogBufferSize` against the size computed from the PCD?

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -102,48 +54,59 @@ AdvancedLoggerGetLoggerInfo (
...
-    mBufferSize = mLoggerInfo->LogBufferSize;
+  LoggerInfo = (ADVANCED_LOGGER_INFO *)(VOID *)FixedPcdGet64 (PcdAdvancedLoggerBase);
+  if (LoggerInfo == NULL) {
+    return NULL;
+  }
+
+  // Initialize HdwPrintLevel if needed.
+  HwPrintLevel = FixedPcdGet32 (PcdAdvancedLoggerHdwPortDebugPrintErrorLevel);
+  if (LoggerInfo->HwPrintLevel != HwPrintLevel) {
+    LoggerInfo->HwPrintLevel = HwPrintLevel;
```
os-d: Can the HwPrintLevel be overwritten by configuration code? There are some slightly complicated flows to configuring what is getting logged, I'm wondering if this could overwrite later configuration

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -102,48 +54,59 @@ AdvancedLoggerGetLoggerInfo (
...
+
+  //
+  // The pointers LogBuffer and LogCurrent, and LogBufferSize, could be written
+  // to by untrusted code.  Here, we check that the pointers are within the
+  // allocated LoggerInfo space, and that LogBufferSize, which is used in
+  // multiple places to see if a new message will fit into the log buffer, is
+  // valid.
+  //
+
+  if (LoggerInfo->Signature != ADVANCED_LOGGER_SIGNATURE) {
```
os-d, line 76: Should we check the signature before we write the HwPrintLevel to the structure? If this is the wrong memory region (or something else overran into it) presumably we wouldn't want to write to it first.

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -87,8 +39,8 @@ ValidateInfoBlock (
   PcdAdvancedLoggerCarBase     -- NOT USED, leave at default
```
makubacki, line 37: I noticed "PcdAdvancedLoggerPages" is stated as being assumed to greater than 64KB. Is there a hard constraint on that somewhere or is it what seems reasonable for a log buffer?

makubacki (approved):

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
cfernald: As discussed offline, I will remove this. Right now, this implementation only can handle a adv logger buffer that was previously intialized. As such it should not presume to set any initial values as it cannot know the origin of the existing value. Only the initial creator of the structure decides default values.

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
cfernald: Will no longer matter because of the above comment.

kuqin12 (approved):

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
cfernald: good catch

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
cfernald: Yeah, this comments seems to have originated in the ARM BaseLib, but i don't see any enforcement or reason given. So unsure why this is here.

PR #458, opened 2024-04-09
Improve robustness of XmlTreeLib

yangrongwei:
## Description

Improve robustness of XmlTreeLib for handling tailing invalid (?) characters.

Current XmlTreeLib will run into **endless loop** if there are some non-whitespace chars after root node. Take `malloc()` on Windows as an example, the memory will be initialized to `0xcd` on debug build. If we put data on this buffer without a proper ending char (null), the `CreateXmlTree()` call will **never return** on this buffer. 

**This is strange.** A better parser lib should (a) report this as an error or (2) stop parsing at the place. 

This PR will stop parsing after _first_ Root node was parsed, discarding any characters after it. As a side effect,  if we have more than one root, only the first root will be reported. This is reasonable, since XML allows only one root node.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ X] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Added test cases and passed unit test on host test.

## Integration Instructions

None.

### DIFF
--- a/XmlSupportPkg/Library/XmlTreeLib/XmlTreeLib.c
+++ b/XmlSupportPkg/Library/XmlTreeLib/XmlTreeLib.c
@@ -877,14 +877,15 @@ BuildNodeList (
     //
     Status = RtlXmlNextToken (&State, &Next, FALSE);
     if (EFI_ERROR (Status)) {
-      DEBUG ((EFI_D_ERROR, "Failed to get the next token, Status = 0x%x\n", Status));
+      DEBUG ((EFI_D_ERROR, "Failed to get the next token, Status = %r\n", Status)); // MS_CHANGE
       goto Exit;
     } else if (Next.fError) {
       //
       // Errors in parse, such as bad characters, come out here in the
       // fError member
       //
-      DEBUG ((EFI_D_ERROR, "Error during tokenization, Status = 0x%x\n", Next.fError));
+      DEBUG ((EFI_D_ERROR, "Error during tokenization, Next.fError = %d\n", Next.fError)); // MS_CHANGE
+      Status = EFI_ABORTED;                                                                // MS_CHANGE
       goto Exit;
     }
 
@@ -1071,7 +1072,7 @@ BuildNodeList (
       //
       Status = AddAttributeToNode (CurrentNode, AttributeName, AttributeValue);
       if (EFI_ERROR (Status)) {
-        DEBUG ((EFI_D_ERROR, "ERROR:  AddAttributeToNode() failed, Status = 0x%x\n", Status));
+        DEBUG ((EFI_D_ERROR, "ERROR:  AddAttributeToNode() failed, Status = %r\n", Status));  // MS_CHANGE
         goto Exit;
       }
     } else if (Next.State == XTSS_ENDELEMENT_NAME) {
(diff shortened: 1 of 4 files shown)


yangrongwei:
@microsoft-github-policy-service agree company="Raymond Yang"

XmlSupportPkg/Library/XmlTreeLib/XmlTreeLib.c
```diff
@@ -877,14 +877,15 @@ BuildNodeList (
     //
     Status = RtlXmlNextToken (&State, &Next, FALSE);
     if (EFI_ERROR (Status)) {
-      DEBUG ((EFI_D_ERROR, "Failed to get the next token, Status = 0x%x\n", Status));
+      DEBUG ((EFI_D_ERROR, "Failed to get the next token, Status = %r\n", Status)); // MS_CHANGE
```
os-d, line 880: The `MS_CHANGE` tags aren't needed for the mu_plus repo (since this entire repo represents mu value adds). For future reference, `MU_CHANGE` is preferred for repos that do have upstream components that we track changes against.

XmlSupportPkg/Library/XmlTreeLib/XmlTreeLib.c
```diff
@@ -1105,6 +1106,20 @@ BuildNodeList (
       //
       if (CurrentNode) {
         CurrentNode = CurrentNode->ParentNode;
+        // MS_CHANGE  [begin]
+        //
+        // Topmost node, ignore tailing characters
+        //
+        // DONE-CCC:
+        //   See also,  TODO-AAA and TODO-BBB in `xml_fasterxml.c`
```
os-d, line 1114: Let's leave full comments, please don't use annotations like DONE-CCC or TODO-AAA.

XmlSupportPkg/Library/XmlTreeLib/fasterxml/xml_fasterxml.c
```diff
@@ -2700,6 +2700,22 @@ RtlXmlNextToken (
...
         return success;
       }
+      // MS_CHANGE
+      //
+      // TODO-BBB:  See TODO-AAA
+      //
+      // if (pRawToken->TokenName == NTXML_RAWTOKEN_ERROR) {
+      //   pToken->fError = TRUE;
+      //   success = EFI_ABORTED;
+      //   return success;
```
os-d, line 2710: Instead of commenting out the code, we can remove it. If it is helpful, do leave a full comment explaining why we don't check if the TokenName == NTXML_RAWTOKEN_ERROR.

XmlSupportPkg/Library/XmlTreeLib/fasterxml/xml_fasterxml.c
```diff
@@ -2946,6 +2962,14 @@ RtlXmlNextToken (
 
         cbTotalTokenLength = pRawToken->Run.cbData;
 
+        // MS_CHANGE
+        //
+        // TODO-AAA:
```
os-d, line 2967: Is this TODO fixable now?

PR #460, opened 2024-04-10
Update the Mem Map and MAT Shell Test to Support RUNTIME_PAGE_ALLOCATION_GRANULARITY Alignment

os-d:
## Description

The Mem Map and MAT shell test currently has a hard dependency on 4k alignment of EFI_MEMORY_MAP and MAT table entries. This is only valid when the RUNTIME_PAGE_ALLOCATION_GRANULARITY is 4k. To make this more universal, compare the MAT entries to RUNTIME_PAGE_ALLOCATION_GRANULARITY as well as the UEFI spec defined EFI_MEMORY_MAP types that should have RUNTIME_PAGE_ALLOCATION_GRANULARITY.

As a small cleanup, rename LegacyMemoryMap to EfiMemoryMap as it is certainly not legacy and to avoid confusion :).

This also removes a duplicate test.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by running the shell app on Q35 and confirming it still succeeds.

## Integration Instructions

N/A. This has no dependency on mu_basecore 64k changes because the RUNTIME_PAGE_ALLOCATION_GRANULARITY is what we expect the MAT and runtime EFI_MEMORY_MAP entries to be aligned to, regardless of whether that is 4k or 64k.

### DIFF
--- a/UefiTestingPkg/FunctionalSystemTests/MemmapAndMatTestApp/MemmapAndMatTestApp.c
+++ b/UefiTestingPkg/FunctionalSystemTests/MemmapAndMatTestApp/MemmapAndMatTestApp.c
@@ -695,7 +695,7 @@ MemmapAndMatTestApp (
   AddTestCase (TableStructureTests, "MAT size should be a multiple of the Descriptor size", "Security.MAT.Size", MatMapSizeShouldBeAMultipleOfDescriptorSize, NULL, NULL, NULL);
   AddTestCase (TableStructureTests, "No standard MemoryMap entries should have a 0 size", "Security.MAT.MemMapZeroSizeEntries", NoEfiMemoryMapEntriesShouldHaveZeroSize, NULL, NULL, NULL);
   AddTestCase (TableStructureTests, "No MAT entries should have a 0 size", "Security.MAT.MatZeroSizeEntries", NoMatMapEntriesShouldHaveZeroSize, NULL, NULL, NULL);
-  AddTestCase (TableStructureTests, "All standard MemoryMap entries should be page aligned", "Security.MAT.MemMapAlignment", AllEfiMemoryMapEntriesShouldBeAligned, NULL, NULL, NULL);
+  AddTestCase (TableStructureTests, "All standard MemoryMap entries should be correctly aligned", "Security.MAT.MemMapAlignment", AllEfiMemoryMapEntriesShouldBeAligned, NULL, NULL, NULL);
 
   //
   // Populate the MatTableContentTests Unit Test Suite.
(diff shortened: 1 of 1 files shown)


UefiTestingPkg/FunctionalSystemTests/MemmapAndMatTestApp/MemmapAndMatTestApp.c
```diff
@@ -172,49 +172,41 @@ NoMatMapEntriesShouldHaveZeroSize (
...
+    // per the UEFI spec, these types need to have
+    // RUNTIME_PAGE_ALLOCATION_GRANULARITY
+    if ((Descriptor->Type == EfiRuntimeServicesCode) ||
+        (Descriptor->Type == EfiRuntimeServicesData) ||
+        (Descriptor->Type == EfiACPIMemoryNVS) ||
+        (Descriptor->Type == EfiReservedMemoryType))
     {
-      Status = UNIT_TEST_ERROR_TEST_FAILED;
-      break;
+      if (((Descriptor->PhysicalStart & (RUNTIME_PAGE_ALLOCATION_GRANULARITY - 1)) != 0) ||
```
apop5, line 199: I have some concerns about this implementation.

now that RUNTIME_PAGE_ALLOCATION_GRANULARITY define can be changed, including this in the code directly means that this has to be compiled on the system that it is targeting, or on a system that is similar enough in settings.

Since RUNTIME_PAGE_ALLOCATION_GRANULARITY can not be modified, I think we should move it to either an input parameter for the utility, or it should be determined through some runtime mechanism.

os-d: Hmm, I actually think that it is good that we target RUNTIME_PAGE_ALLOCATION_GRANULARITY here. That is UEFI spec defined. So while there are some older systems that cannot support say 64k as a RPAG, they are breaking the UEFI spec. I don't think this test should pass for them. Although I guess this would lead to inconsistent results, if you compiled on a standard system, the test would break, if you compiled against the old system, it would pass. I still sort of thing that's okay. If you run your own test, fine it works. If you run the standard compiled test, you see you are breaking the spec. The other piece here is that all new platforms should be using the standard RPAG, so this should not be an issue going forward. So I would be inclined to leave this as is, from the test perspective, directly tested against RPAG is the right thing to do.

apop5 (approved):

TaylorBeebe (approved):

PR #469, opened 2024-04-17
Support the platform specific event hooks which can be customized.

Eathonhsu:
## Description

The current event triggers lack flexibility and cannot be effectively used on some platforms.

1. Create hook functions at the end of dxe event before and after the event.
2. Create a hook after the boot manager dispatch the deferred 3rd party images Event.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [x] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

It will not affect the original process, mainly serving as an extension for platform use.

## Integration Instructions

Create a new library to be replaced by the platform to expand functionality.


apop5:
@Eathonhsu can you please fill out the Description details?

MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
```diff
@@ -41,11 +41,15 @@ ExitPmAuth (
 
   DEBUG ((DEBUG_INFO, "ExitPmAuth ()- Start\n"));
 
+  BeforeEndOfDxeEventHook (); // MS_CHANGE
```
kuqin12: Since this is in MU_PLUS, we should not need to use `MS_CHANGE` tags in this PR.

MsCorePkg/Library/PlatformBootManagerLib/BdsPlatform.c
```diff
@@ -41,11 +41,15 @@ ExitPmAuth (
...
   DEBUG ((DEBUG_INFO, "ExitPmAuth ()- Start\n"));
 
+  BeforeEndOfDxeEventHook (); // MS_CHANGE
+
   //
   // Since PI1.2.1, we need signal EndOfDxe as ExitPmAuth
   //
   EfiEventGroupSignal (&gEfiEndOfDxeEventGroupGuid);
 
+  AfterEndOfDxeEventHook (); // MS_CHANGE
```
kuqin12: What are the usage of these hooks?

Also, is the abstraction method better than the event signaling? i.e. line 49?

@apop5 @spbrogan thoughts?

Eathonhsu: In practical cases, we would use these hooks to send platform-specific events to enable the functionality to be activated properly.

Why I choose to use a library instead of send `EventGroupSignal`, the reason is that after the platform receives the event callback, the callback function will send its platform-specific events. I'm concerned about potential nesting of muti-events.

Eathonhsu:
> @Eathonhsu can you please fill out the Description details?

Updated, thanks for the suggestion

apop5:
> > @Eathonhsu can you please fill out the Description details?
> 
> Updated, thanks for the suggestion

I added the "breaking change" because this would require existing platforms to include a new library instance in their project to allow this to build.

PR #473, opened 2024-04-26
Moving the global variable from header file into library instances

kuqin12:
## Description

The current implementation of instantiating global variables will make each c file that includes this header to carry a copy of such variable, which is space redundant.

This change removed the variable into the single module that uses it and removed the structure it defines. The function descriptions are also updated.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

This is a non-functional change.

## Integration Instructions

N/A

### DIFF
--- a/MfciPkg/Include/Library/MfciDeviceIdSupportLib.h
+++ b/MfciPkg/Include/Library/MfciDeviceIdSupportLib.h
@@ -8,48 +8,45 @@
   SPDX-License-Identifier: BSD-2-Clause-Patent
 **/
 
-#ifndef __MFCI_DEVICE_ID_SUPPORT_LIB_H__
-#define __MFCI_DEVICE_ID_SUPPORT_LIB_H__
+#ifndef MFCI_DEVICE_ID_SUPPORT_LIB_H_
+#define MFCI_DEVICE_ID_SUPPORT_LIB_H_
 
 #include <MfciVariables.h>
 
-// define a generic function prototype shared by all library functions
-
 /**
- * Get a device-unique targeting value
- *
- * @param[out] String   Device targeting value, a UTF-16 little endian string.
- *                      Includes a wide NULL terminator.
- *                      Refer to MfciPkg/Include/MfciVariables.h for more details.
- *
- * @param[out] StringSize  (OPTIONAL) String size in bytes including the wide NULL terminator.
- *                         NULL may be supplied if the size is not requested (it is NULL terminated after all)
- *
- * It is the callers responsibility to free the String buffer returned using FreePool()
- *
- * @return EFI_STATUS
- *
- * @retval EFI_UNSUPPORTED         Likely using the NULL library instance
- * @retval EFI_SUCCESS             Successfully retrieved the string and length
- */
+  Function pointer definition to get a device-unique targeting value.
+  It is the callers responsibility to free the String buffer returned using FreePool().
 
+  @param[out] String      Device targeting value, a UTF-16 little endian string.
+                          Includes a wide NULL terminator.
+                          Refer to MfciPkg/Include/MfciVariables.h for more details.
+  @param[out] StringSize  (OPTIONAL) String size in bytes including the wide NULL terminator.
+                          NULL may be supplied if the size is not requested (it is NULL terminated after all)
+
+  @retval EFI_UNSUPPORTED       Likely using the NULL library instance.
+  @retval EFI_SUCCESS           Successfully retrieved the string and length.
+  @retval EFI_OUT_OF_RESOURCES  There is not enough memory to allocate the string.
+  @retval EFI_INVALID_PARAMETER The String is NULL.
+**/
 typedef
 EFI_STATUS
 (EFIAPI *MFCI_DEVICE_ID_FN)(
   OUT CHAR16 **String,
-  OUT UINTN *StringSize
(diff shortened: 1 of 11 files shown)


MfciPkg/Include/Library/MfciDeviceIdSupportLib.h
```diff
@@ -110,40 +112,22 @@ MfciIdSupportGetOem1 (
...
-
-// define a structure that pairs up the function pointer with the UEFI variable name
-typedef struct {
-  MFCI_DEVICE_ID_FN    DeviceIdFn;
-  CHAR16               *DeviceIdVarName;
-} MFCI_DEVICE_ID_FN_TO_VAR_NAME_MAP;
-
-// populate the array of structures that pair up the functions with variable names
-#define MFCI_TARGET_VAR_COUNT  5
-STATIC CONST MFCI_DEVICE_ID_FN_TO_VAR_NAME_MAP  gDeviceIdFnToTargetVarNameMap[MFCI_TARGET_VAR_COUNT] = {
```
os-d, line 141: gPolicyBlobFieldName and gPolicyTargetFieldVarNames have the same pattern

kuqin12: thanks, let me do another search on it.

MfciPkg/Include/Library/MfciDeviceIdSupportLib.h
```diff
@@ -8,48 +8,45 @@
...
+                          NULL may be supplied if the size is not requested (it is NULL terminated after all)
+
+  @retval EFI_UNSUPPORTED       Likely using the NULL library instance.
+  @retval EFI_SUCCESS           Successfully retrieved the string and length.
+  @retval EFI_OUT_OF_RESOURCES  There is not enough memory to allocate the string.
+  @retval EFI_INVALID_PARAMETER The String is NULL.
+**/
 typedef
 EFI_STATUS
 (EFIAPI *MFCI_DEVICE_ID_FN)(
```
makubacki, line 33: Does this still need to be in the public library interface?

MfciPkg/MfciDxe/MfciDxe.c
```diff
@@ -31,6 +31,27 @@
 
 #include "MfciDxe.h"
 
+/**
+ * the following helps iterate over the functions and set the corresponding target variable names
+ */
```
makubacki:
```suggestion
//
// The following helps iterate over the functions and set the corresponding target variable names.
//
```

MfciPkg/MfciDxe/MfciDxe.c
```diff
@@ -31,6 +31,27 @@
...
+
+// define a structure that pairs up the function pointer with the UEFI variable name
+typedef struct {
+  MFCI_DEVICE_ID_FN    DeviceIdFn;
+  CHAR16               *DeviceIdVarName;
+} MFCI_DEVICE_ID_FN_TO_VAR_NAME_MAP;
+
+// populate the array of structures that pair up the functions with variable names
+#define MFCI_TARGET_VAR_COUNT  5
```
makubacki: Unless `MFCI_TARGET_VAR_COUNT` is used elsewhere, I think it it'd be easier to maintain the code without it since the array size can be deduced from the initializers here.

kuqin12: Yeah, hold on this PR. There seems to be some complication with other platforms...

MfciPkg/Include/Library/MfciDeviceIdSupportLib.h
kuqin12: Fixed those.

MfciPkg/Include/Library/MfciDeviceIdSupportLib.h
kuqin12: Yes, unfortunately the global data is still referenced in some platform code.

MfciPkg/MfciDxe/MfciDxe.c
kuqin12: The count is in place so as ARRAY_SIZE will not throw an exception on sizeof returning 0.

os-d (approved):

MfciPkg/MfciDxe/MfciDxe.c
makubacki: I see it became an extern after so that makes sense in that case.

MfciPkg/Include/Library/MfciDeviceIdSupportLib.h
makubacki: It would be nice to clean up the interface some more, but given the potential impact on platforms, I'll approve as-is.

makubacki (approved):

MfciPkg/Include/Library/MfciDeviceIdSupportLib.h
kuqin12: We should see if the platform can drop the dependency on these array items at some point. If so, we could easily move this into a module for better sanity.

kuqin12:
Merging the change to unblock the upcoming `STATIC` change from basecore.

Issue #426, opened 2024-01-30
[Bug]: Math in OnScreenKeyboardDxe can result in dimensions that overrun max resolution.


rogurr:
### Is there an existing issue for this?

- [X] I have searched existing issues

### Current Behavior

The transforms used in the OnScreenKeyboardDxe driver in the DisplayTransform.c file all use floats to perform calculations for rotation, positioning, scaling, etc.  When doing the math for putting the keyboard in the lower right corner of the screen on certain resolutions results in a 1 pixel overrun of either the X or Y max resolution.  This incorrect X or Y point eventually makes it's way to the RenderingEngineDxe driver that makes a mParentGop->Blt() call into the GOP with the 1 pixel overrun.  Depending on the silicon vendor's GOP, that can have differing results from being ignored to on screen corruption.

### Expected Behavior

Three areas need to be addressed:

1) The OnScreenKeyboardDriver.c driver file uses that math for several functions to scale, rotate, and position the keyboard.  They all have access to a member GOP pointer, so logic should probably be added to those functions to guarantee any point in the keyboard does not go over the expected resolution.
2) The RenderingEngine.c file in the RenderingEngineDxe driver publishes a protocol that the keyboard driver uses.  Since it is a public API, the driver should probably validate the inputs are acceptable and either return an error code, or debug print and clip the overrun.
3) The RenderingEngine.c file in the RenderingEngineDxe driver makes multiple calls to mParentGop->Blt() which is using the GOP protocol without checking the return status code.  Again, we probably need to check the return code and pass up to the caller of the rendering protocol user.

### Steps To Reproduce

Due to this being an error handler issue and needing a vendor specific GOP with specific keyboard size and specific screen resolution to get the proper rounding error in the float math, it is very difficult to reproduce.  Easiest way to see the error is to modify the SREActivateSurface function call in the RenderingEngine.c file to increase the FrameWidth parameter in the mParentGop->Blt calls to be 1 pixel past the X resolution.

### Build Environment

```markdown
- OS(s): Windows
- Tool Chain(s): VS2022
- Targets Impacted: All
```

### Version Information

```text
url = https://github.com/microsoft/mu_plus
branch = release/202302
SHA = 69d1c094ca61f6d12e8c8e172b07576120e6d259
```

### Urgency

Medium

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

Maintainer feedback requested

### Anything else?

[...]

PR #512, opened 2024-07-04
MsGraphicsPkg: Fix issues in coordinate calculation

wenbhou:
## Description

Fix issues in coordinate calculation to solve On-screen keyboard render issue, and potential issues in other UI widgets.
See https://github.com/microsoft/mu_plus/issues/518

The SWM_RECT structure defines a standard bounding rectangle by specifying the left, top, right, and bottom coordinates. For any SWM_RECT instance, the width is calculated as `width = right - left + 1`, and the height is determined by `height = bottom - top + 1`. When the origin point of a rectangle is known, along with its width and height, the right coordinate can be derived using `right = left + width - 1`, and the bottom coordinate is given by `bottom = top + height - 1`. Ensuring the correct application of these formulas prevents the On-Screen Keyboard (OSK) boundary from exceeding its defined limits.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Boot to UEFI menu, all UI items rendered correctly and OSK functioned well.

## Integration Instructions

N/A

### DIFF
--- a/MsGraphicsPkg/DisplayEngineDxe/FormDisplay.c
+++ b/MsGraphicsPkg/DisplayEngineDxe/FormDisplay.c
@@ -923,8 +923,8 @@ CreateFormControls (
   //
   FormRect.Left   = mMasterFrameWidth;
   FormRect.Top    = mTitleBarHeight;
-  FormRect.Right  = mGop->Mode->Info->HorizontalResolution;
-  FormRect.Bottom = mGop->Mode->Info->VerticalResolution;
+  FormRect.Right  = mGop->Mode->Info->HorizontalResolution - 1;
+  FormRect.Bottom = mGop->Mode->Info->VerticalResolution - 1;
 
   // Create a canvas for rendering the HII form.
   //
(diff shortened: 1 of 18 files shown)


os-d (approved):
Can you add some more context in the PR description as to why the -1's and +1's make sense? I.e. where the issue was. Can you also add whether this PR fully resolves the bug you list or whether it only partially resolves it. It doesn't look like this PR does any bounds checking, which the bug references.

MsGraphicsPkg/DisplayEngineDxe/FormDisplay.c
makubacki, line 1: @wenbhou, I had similar questions to Oliver so I'm leaving this as a file comment to address.

1. Can you please explain conceptually why the +1/-1 code is needed? My first generic thought is that this is a lot of off by one adjustments to get correct throughout the code. Is there a simpler way to represent/calculate these values that is not so fragile? I assume you've looked through the code enough to determine that this is the cleanest solution. Those details would be helpful.

2. #426 mentions validation and bounds checking that is not currently implemented here. Do you plan to implement that portion of the issue?

makubacki: For (1), I think it's implied this is related to 0 start and -1 end, but a brief explanation of that would probably help most people with understanding the change in the future. In regards to making the code less fragile, you could consider using a macro to return common values such as the endpoint and length. That would centralize the calculation reducing repetition elsewhere in the code.

wenbhou: I will try to make the code less fragile.
As for (2), I am not going to implement the validation and bounds checking. Personally, I don't think the overrunning is due to the math with floating-point numbers, but caused by the missing of +1 or -1 in some width or coordinates calculations. Do you think if I should open a new bug for my PR?

makubacki: I suggest opening a new issue. Since you've invested some time to look at this, it would also help if you could add some data/context to the existing issue.

wenbhou: New issue opened: https://github.com/microsoft/mu_plus/issues/518

makubacki: You're still planning to push a change to make the code less fragile, right?

wenbhou: yes

wenbhou: Pushed. Please review again. Thanks

makubacki, line 1: @wenbhou, can you confirm you tested again after the latest changes?

makubacki (approved):

MsGraphicsPkg/DisplayEngineDxe/FormDisplay.c
wenbhou: Yes, I tested on both 13" and 15" devices, went thru all UEFI front page items. No regression observed and OSK can render properly.

PR #516, opened 2024-07-10
Add prefix for DEBUG_ERROR prints for data analysing purpose

liqiqiii:
## Description

Add prefix for DEBUG_ERROR prints for data analysing purpose
DEBUG_ERROR level prints are generally very important and worth analyzing. One blocking thing now is that we can't tell the difference between different debug level msgs because not all DEBUG_ERROR will have an "error:" as the prefix of the debug msg. This PR is to add a non-used keyword so we can easily track the DEBUG_ERRORs.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with a platform that uses advancedlogger and can see [LOG_ERROR] in front of every debug_error msg.

## Integration Instructions

N/A

### DIFF
--- a/AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
+++ b/AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
@@ -263,6 +263,59 @@ class AdvLogParser ():
     PHASE_STRING_LIST = ["[UNSPECIFIED] ", "[SEC] ", "[PEI] ", "[PEI64] ",
                          "[DXE] ", "[RUNTIME] ", "[MM_CORE] ", "[MM] ",
                          "[SMM_CORE] ", "[SMM] ", "[TFA] "]
+
+    # Debug levels from MU_BASECORE\MdePkg\Include\Library\DebugLib.h
+    # //
+    # // Declare bits for PcdDebugPrintErrorLevel and the ErrorLevel parameter of DebugPrint()
+    # //
+    # #define DEBUG_INIT      0x00000001       // Initialization
+    # #define DEBUG_WARN      0x00000002       // Warnings
+    # #define DEBUG_LOAD      0x00000004       // Load events
+    # #define DEBUG_FS        0x00000008       // EFI File system
+    # #define DEBUG_POOL      0x00000010       // Alloc & Free (pool)
+    # #define DEBUG_PAGE      0x00000020       // Alloc & Free (page)
+    # #define DEBUG_INFO      0x00000040       // Informational debug messages
+    # #define DEBUG_DISPATCH  0x00000080       // PEI/DXE/SMM Dispatchers
+    # #define DEBUG_VARIABLE  0x00000100       // Variable
+    # #define DEBUG_SMI       0x00000200       // MS_CHANGE_141550: Added for SMI audting options.
+    # #define DEBUG_BM        0x00000400       // Boot Manager
+    # #define DEBUG_BLKIO     0x00001000       // BlkIo Driver
+    # #define DEBUG_NET       0x00004000       // Network Io Driver
+    # #define DEBUG_UNDI      0x00010000       // UNDI Driver
+    # #define DEBUG_LOADFILE  0x00020000       // LoadFile
+    # #define DEBUG_EVENT     0x00080000       // Event messages
+    # #define DEBUG_GCD       0x00100000       // Global Coherency Database changes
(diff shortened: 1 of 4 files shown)


os-d:
I don't think this adds value. DEBUG_ERROR gets used in a lot of places for things that aren't actual errors. Adding something like this will only cause spurious errors to get reported. At the end of the day, nothing will beat high quality print statements and actually analyzing a log.

makubacki:
> I don't think this adds value. DEBUG_ERROR gets used in a lot of places for things that aren't actual errors. Adding something like this will only cause spurious errors to get reported. At the end of the day, nothing will beat high quality print statements and actually analyzing a log.

That's my main concern. `DEBUG_ERROR` is often generically used as a "high level" for messages so they can appear in RELEASE builds and so on. I'm not going to defend that though; it is a bad design/pattern. Error should mean error and developers should and probably would pay more attention to those messages if that was the case.

I could maybe see this being controlled by a platform with something like `PcdDebugPropertyMask` (default off), but I think that extends the mess from `DEBUG_ERROR` by encoding it into a PCD rather than addressing the problem like just adding another level for "important" but not error messages. I also feel that if this were enabled (however that is done), the prefix of all levels (e.g. `info:`, `warn:`, etc.) should be printed for consistency.

liqiqiii:
> > I don't think this adds value. DEBUG_ERROR gets used in a lot of places for things that aren't actual errors. Adding something like this will only cause spurious errors to get reported. At the end of the day, nothing will beat high quality print statements and actually analyzing a log.
> 
> That's my main concern. `DEBUG_ERROR` is often generically used as a "high level" for messages so they can appear in RELEASE builds and so on. I'm not going to defend that though; it is a bad design/pattern. Error should mean error and developers should and probably would pay more attention to those messages if that was the case.
> 
> I could maybe see this being controlled by a platform with something like `PcdDebugPropertyMask` (default off), but I think that extends the mess from `DEBUG_ERROR` by encoding it into a PCD rather than addressing the problem like just adding another level for "important" but not error messages. I also feel that if this were enabled (however that is done), the prefix of all levels (e.g. `info:`, `warn:`, etc.) should be printed for consistency.

Thanks Oliver and Michael, I think it's a good point that DEBUG_ERROR is often used in non error cases(kind of a routine), I was planning to do either cleaning up non-error debug errors + provide an ignore file that includes all non-error ones, so we can get really useful results, or I go and add error: into the real errors like some of them already have right now(non consistent format of print statements)  and use error: as the keyword. Both will take time but I think if it's finished properly, it will really add values to our detailed uefi logs. 

Now the uefi logs are gold that's hard to mine, Errors that don't block boot will be very hard to be found as no one notices in the tens of thousands of lines of log.

The pcd idea is sth worth a trial if everyone likes it. 
I was thinking about adding prefix for all levels but was little bit worried about the extra space it would consume, adding the [LOG_ERROR] will increase binary size by around 50KB with several thousand DEBUG_ERRORs total; if we add it for all debug prints that would be a significant amount of size increase, and we don't really care about other level prefixes that much so I didn't add it. If we have plan to add it, I will also make some assessment and see if it really explodes the size or not.

liqiqiii:
Hi @os-d @makubacki Oliver and Michael, I had an offline talk with Kun @kuqin12 and he mentioned a better way to use the existing debug_level in the metadata and in this way, we don't need to touch the firmware itself. I have tried with it and it's working perfectly fine, we got all the debug levels printed out. 
22:36:22.840 : [MM_CORE] [DEBUG_INFO]AdvancedLoggerGetLoggerInfo: LoggerInfo=70F37000
22:36:22.857 : [MM_CORE] [DEBUG_ERROR]ERROR - The FV at 0xFFE60000 is invalid!

makubacki:
> Hi @os-d @makubacki Oliver and Michael, I had an offline talk with Kun @kuqin12 and he mentioned a better way to use the existing debug_level in the metadata and in this way, we don't need to touch the firmware itself. I have tried with it and it's working perfectly fine, we got all the debug levels printed out. 22:36:22.840 : [MM_CORE] [DEBUG_INFO]AdvancedLoggerGetLoggerInfo: LoggerInfo=70F37000 22:36:22.857 : [MM_CORE] [DEBUG_ERROR]ERROR - The FV at 0xFFE60000 is invalid!

Closing the PR.

PR #523, opened 2024-07-15
Add debug level prefix for advanced logger memory message entries

liqiqiii:
## Description

This change added the existing metadata - debug level into the final advanced logger memory entries, it can differentiate the different debug levels, DEBUG_INFO, DEBUG_ERROR, etc. 
We don't need to touch firmware as debug level info is already part of metadata, so we only need to update the decode script.
It's useful to be able to filter out the specific debug level msgs that we care about and make data processing of uefi logs possible. 
For example, DEBUG_ERROR level prints are generally very important and worth analyzing. One blocking thing now is that we can't tell the difference between different debug level msgs because not all DEBUG_ERROR will have an "error:" as the prefix of the debug msg, and search for "error" will also be noisy as it's a commonly used word. After this PR checked in, we can easily track the DEBUG_ERRORs.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with a platform that uses advancedlogger and can seedebug level prefixes in the msg.
22:36:22.840 : [MM_CORE] [DEBUG_INFO]AdvancedLoggerGetLoggerInfo: LoggerInfo=70F37000 
22:36:22.857 : [MM_CORE] [DEBUG_ERROR]ERROR - The FV at 0xFFE60000 is invalid!

## Integration Instructions

N/A


AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
```diff
@@ -1022,4 +1086,4 @@ def main():
 # --------------------------------------------------------------------------- #
 if __name__ == '__main__':
 
-    main()
+    main()
```
os-d: We can leave this as is.

os-d (approved):

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
liqiqiii: Thanks Oliver, I ate the blank line.

kenlautner (approved):

PR #524, opened 2024-07-15
MsCorePkg: Update telemetry logging with PI Spec EFI Status Code

VivianNK:
## Description

Previously in MemoryTypeInformationChangeLib, we were using a custom Status Code. That code is being removed from MU_BASECORE  in [PR #1001](https://github.com/microsoft/mu_basecore/pull/1001) and we're using an existing code from PI Spec, EFI_SW_EC_ILLEGAL_SOFTWARE_STATE instead.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?
 
## How This Was Tested

CI

## Integration Instructions

Upgrading to or already using mu_basecore release/202311, change any reference to the macro as follows:
Any instances of `EFI_SW_EC_MEMORY_TYPE_INFORMATION_CHANGE` will need to be replaced with `EFI_SW_EC_ILLEGAL_SOFTWARE_STATE`

### DIFF
--- a/MsCorePkg/Library/MemoryTypeInformationChangeLib/MemoryTypeInformationChangeLib.c
+++ b/MsCorePkg/Library/MemoryTypeInformationChangeLib/MemoryTypeInformationChangeLib.c
@@ -30,12 +30,12 @@ ReportMemoryTypeInformationChange (
 {
   //
   // This telemetry event is logged by a lib linked to BdsDxe, thus a EFI_SOFTWARE_DXE_BS_DRIVER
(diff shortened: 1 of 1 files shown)


os-d (approved):
This should be labeled a breaking change, with the integration steps including information about what version of basecore this is linked to (whatever release has the code), unless this code is not newly added.

apop5 (approved):

PR #525, opened 2024-07-19
Add debug level prefix for advanced logger memory message entries - Driver implementation

liqiqiii:
## Description

Add debug level prefix for advanced logger memory message entries - Driver implementation
This change added the existing metadata - debug level into the final
advanced logger memory entries through AdvancedFileLogger.  This is a followup PR for https://github.com/microsoft/mu_plus/commit/243bd187e4aa6ac68a3eb323d07554576d39ad47.
After this PR checked in, we can easily track the DEBUG_ERRORs through advanced logger files on UEFI that included this driver.
Added an extra space in the decodeuefilog.py to improve readability of the log.
For the for loop in the code, ran with perf_trace enabled, and influence is very low and can be ignored.
Updated the prefix of python script to match the design here. 
Use [ERROR] instead of [DEBUG_ERROR], which could save overall log file sizes and memory buffer usage.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with two platforms that uses AdvancedFileLogger and can see debug level
prefixes in the logs in the EFI Partition and USB UefiLogs folder.
[MM_CORE] [ERROR] Image - MmSupervisorCore.pdb

## Integration Instructions

N/A


AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
```diff
@@ -293,27 +293,27 @@ class AdvLogParser ():
     # #define DEBUG_ERROR  0x80000000          // Error
 
     debug_levels_dict = {
```
kuqin12, line 295: Would it be easier to change how these strings are used instead of changing the string themselves?

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
```diff
@@ -40,10 +40,42 @@ CONST   CHAR8                 *AdvMsgEntryPrefix[ADVANCED_LOGGER_PHASE_CNT] = {
   "[TFA] ",
 };
 
-#define ADV_TIME_STAMP_FORMAT  "%2.2d:%2.2d:%2.2d.%3.3d : "
-#define ADV_TIME_STAMP_RESULT  "hh:mm:ss:ttt : "
-#define ADV_PHASE_ERR_FORMAT   "[%04X] "
-#define ADV_PHASE_MAX_SIZE     32
+// Define a structure to hold debug flag information
+typedef struct {
+  const char    *name;
```
kuqin12:
```suggestion
  CONST CHAR8    *name;
```

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
```diff
@@ -136,6 +168,40 @@ FormatPhasePrefix (
...
+FormatDebugLevelPrefix (
+  IN CHAR8   *MessageBuffer,
+  IN UINTN   MessageBufferSize,
+  IN UINT32  DebugLevel
+  )
+{
+  UINTN  DebugLevelStringLen;
+
+  // Add the matched debug level
+  for (UINT8 i = 0; i < sizeof (debug_levels) / sizeof (debug_levels[0]); i++) {
```
kuqin12: Can we please change this to UEFI style? Typically, we do not declare variables in the loop.

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
```diff
@@ -40,10 +40,42 @@ CONST   CHAR8                 *AdvMsgEntryPrefix[ADVANCED_LOGGER_PHASE_CNT] = {
...
-#define ADV_PHASE_ERR_FORMAT   "[%04X] "
-#define ADV_PHASE_MAX_SIZE     32
+// Define a structure to hold debug flag information
+typedef struct {
+  const char    *name;
+  UINT32        value;
+} DebugLevel;
+
+// Create an array of DebugLevel structures
+DebugLevel  debug_levels[] = {
```
kuqin12: Please capitalize the variable and its type.

kuqin12 (approved):

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
makubacki: In addition to Kun's comment:

- Name the variable something other than `i` like `Index`
- Change the type of the index variable to `UINTN` instead of `UINT8`

makubacki: You should be able to use `ARRAY_SIZE()` instead of `sizeof (x) / sizeoff (x[0])`.

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
```diff
@@ -136,6 +168,40 @@ FormatPhasePrefix (
...
+  returns the debug_level prefix in from the AdvMsgEntryPrefix, otherwise raw debug level value is returned.
+
+  @param  MessageBuffer
+  @param  MessageBufferSize
+  @param  Phase
+
+  @retval Number of characters printed
+*/
+STATIC
+UINT16
```
makubacki, line 186: Typically, length/count is represented by `UINTN`.

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
```diff
@@ -136,6 +168,40 @@ FormatPhasePrefix (
...
 
+/**
+  FormatDebugLevelPrefix
+
+  Adds a debug level indicator to the message being returned.  If debug level is recognized and specified,
+  returns the debug_level prefix in from the AdvMsgEntryPrefix, otherwise raw debug level value is returned.
+
+  @param  MessageBuffer
+  @param  MessageBufferSize
+  @param  Phase
```
makubacki:
```suggestion
  @param[in]  MessageBuffer
  @param[in]  MessageBufferSize
  @param[in]  Phase
```

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
```diff
@@ -136,6 +168,40 @@ FormatPhasePrefix (
   return (UINT16)PhaseStringLen;
 }
 
+/**
+  FormatDebugLevelPrefix
+
+  Adds a debug level indicator to the message being returned.  If debug level is recognized and specified,
+  returns the debug_level prefix in from the AdvMsgEntryPrefix, otherwise raw debug level value is returned.
```
makubacki: Putting the function name at the beginning of the function description is not necessary.

```suggestion
  Adds a debug level indicator to the message being returned.  If debug level is recognized and specified,
  returns the debug_level prefix in from the AdvMsgEntryPrefix, otherwise raw debug level value is returned.
```

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
```diff
@@ -40,10 +40,42 @@ CONST   CHAR8                 *AdvMsgEntryPrefix[ADVANCED_LOGGER_PHASE_CNT] = {
...
+  { "[DEBUG_LOADFILE] ",      0x00020000 },
+  { "[DEBUG_EVENT] ",         0x00080000 },
+  { "[DEBUG_GCD] ",           0x00100000 },
+  { "[DEBUG_CACHE] ",         0x00200000 },
+  { "[DEBUG_VERBOSE] ",       0x00400000 },
+  { "[DEBUG_MANAGEABILITY] ", 0x00800000 },
+  { "[DEBUG_ERROR] ",         0x80000000 }
+};
+
+#define ADV_TIME_STAMP_FORMAT     "%2.2d:%2.2d:%2.2d.%3.3d : "
```
makubacki: I recommend prefixing these with `ADV_LOG` for clarity. So, `ADV_LOG_TIME_STAMP_FORMAT`.

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
makubacki: - `DEBUG_LEVEL`
- `DebugLevels`

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
makubacki: `*Name` not `*name`

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
```diff
@@ -40,10 +40,42 @@ CONST   CHAR8                 *AdvMsgEntryPrefix[ADVANCED_LOGGER_PHASE_CNT] = {
...
 };
 
-#define ADV_TIME_STAMP_FORMAT  "%2.2d:%2.2d:%2.2d.%3.3d : "
-#define ADV_TIME_STAMP_RESULT  "hh:mm:ss:ttt : "
-#define ADV_PHASE_ERR_FORMAT   "[%04X] "
-#define ADV_PHASE_MAX_SIZE     32
+// Define a structure to hold debug flag information
+typedef struct {
+  const char    *name;
+  UINT32        value;
```
makubacki:
```suggestion
  UINT32        Value;
```

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
```diff
@@ -293,27 +293,27 @@ class AdvLogParser ():
     # #define DEBUG_ERROR  0x80000000          // Error
 
     debug_levels_dict = {
-        0x00000001: "[DEBUG_INIT]",
```
makubacki, line 296: Do the strings really need `DEBUG_`? For example, can this just be `[INIT]`?

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
makubacki: I don't think they should be defined with a space. That is a formatting decision that should be made at formatting time.

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
```diff
@@ -136,6 +168,40 @@ FormatPhasePrefix (
...
+  IN CHAR8   *MessageBuffer,
+  IN UINTN   MessageBufferSize,
+  IN UINT32  DebugLevel
+  )
+{
+  UINTN  DebugLevelStringLen;
+
+  // Add the matched debug level
+  for (UINT8 i = 0; i < sizeof (debug_levels) / sizeof (debug_levels[0]); i++) {
+    if (DebugLevel == debug_levels[i].value) {
```
apop5: DebugLevels are a bitwise map. 

Some debug messages in C code will use multiple of the bitwise map i.e. `DEBUG ((DEBUG_ERROR | DEBUG_INIT, "Message Here\n"));`

You will need to account for messages that may have multiple bits set.

```suggestion
    if ( (DebugLevel & debug_levels[i].value) == debug_levels[i].value) {

```

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
```diff
@@ -136,6 +168,40 @@ FormatPhasePrefix (
...
+STATIC
+UINT16
+FormatDebugLevelPrefix (
+  IN CHAR8   *MessageBuffer,
+  IN UINTN   MessageBufferSize,
+  IN UINT32  DebugLevel
+  )
+{
+  UINTN  DebugLevelStringLen;
+
```
apop5, line 210: Can you please validate the input buffers before you use them? and MessageBufferSize. If a size zero buffer is passed in, there is no need to do anything else.

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
liqiqiii: Changed, thanks Kun

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
liqiqiii: Changed, thanks Michael.

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
liqiqiii: Set this to UINT16 because the final output type of MessageLen in the BlockEntry is UINT16, so keep these UINT16 would be much easier.

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
liqiqiii: Thanks Aaron, fixed

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
liqiqiii: `DEBUG_` is required as the words like `init` or `error` are very commonly used words in the debug print messages, so it will cause lots of noise if we only use the `error` itself.

AdvLoggerPkg/Library/AdvancedLoggerAccessLib/AdvancedLoggerAccessLib.c
liqiqiii: Done, thanks Aaron!

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
makubacki: @liqiqiii, can you describe the issue you encountered trying to change this here?

liqiqiii: I was seeing a red null when trying to open the output log file in vscode:

I was trying to do this:
  // Print the debug levels
  for (Index = 0; Index < ARRAY_SIZE (DebugLevels); Index++) {
    if ( (DebugLevel & DebugLevels[Index].Value) == DebugLevels[Index].Value) {
      DebugLevelStringLen = AsciiSPrint (MessageBuffer, MessageBufferSize, DebugLevels[Index].Name);
      // Add an extra space to the end of the string for readability purposes
      AsciiSPrint(MessageBuffer + DebugLevelStringLen, 1, " ");
      // AsciiSPrint(MessageBuffer + DebugLevelStringLen + 1, 1, "\0");  //with and without this line the same red null
      DebugLevelStringLen++;
      return (UINT16)DebugLevelStringLen;
    }

makubacki: Can you try this?

```c
  if (DebugLevelStringLen < MessageBufferSize - 1) {
    MessageBuffer[DebugLevelStringLen] = ' ';
    MessageBuffer[DebugLevelStringLen + 1] = '\0';
  }
```

liqiqiii: It works! It works without using the AsciiSPrint.
```
      // Verify string length and add an extra space
      if (DebugLevelStringLen < MessageBufferSize - 1) {
        MessageBuffer[DebugLevelStringLen]     = ' ';
        MessageBuffer[DebugLevelStringLen + 1] = '\0';
        DebugLevelStringLen++;
      }
```
When I tried the same thing with the AsciiSPrint it will still have the red null and that was where I stuck, no matter whether I add only a ' ' or also ' ' + '\0' to the end it will always have the red null, so I think the function did sth that will cause it. Because we should only do ascii prints, so I didn't even think of trying without this function call.
Thanks so much Michael.

liqiqiii: It might be relevant to the input of AsciiSPrint, it needs to send in a string instead of a char, so " " might be causing the issue. Although theorectically it should be the same when I only copy the first char in the string, that's why I didn't think it is the problem.

AdvLoggerPkg/Application/DecodeUefiLog/DecodeUefiLog.py
makubacki: I understand `init` and `error` are common, but is `[INIT]`? This seems unnecessarily verbose as a log marker that is going out over interfaces like serial for every message. It will also have a substantial impact on overall log file sizes and memory buffer usage.

liqiqiii: That's a good point, for certain key words like `[ERROR]` we do have some usage already in the tree, but they are at the right place indicating the right debug level so we are good.I have tested and updated the relevant part in both files. Thanks Michael.

makubacki (approved):
This will cause some additional size but I'm okay to approve it if the tradeoff provides more value to platform debug.

PR #536, opened 2024-08-03
Removed all instances of PcdAdvancedLoggerPeiInRam

### LINKED ISSUE #521: [Bug]: PcdAdvancedLoggerPeiInRam Does Not Work

- [X] I have searched existing issues

### Current Behavior

PeiCore has a built in assumption that full memory is not available at its start, even on systems that do have full memory available. As such, this PCD does not work.

### Expected Behavior

The PCD should be removed.

### Steps To Reproduce

N/A.

### Build Environment

```markdown
- OS(s):
- Tool Chain(s):
- Targets Impacted:
All
```

### Version Information

```text
top of tree
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

Raymond-MS:
## Description

The PCD PcdAdvancedLoggerPeiInRam does not work and is being removed. For Intel systems, the PEI phase uses cache-as-ram and determines that there is no way to allocate memory. This PCD was supposed to signal to the logger that it could allocate memory, however, it would error as memory allocation was not allowed in that phase.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Ran the CI build for all packages in mu_plus for the X64 architecture. All tests passing.

## Integration Instructions

N/A

### DIFF
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.c
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.c
@@ -453,7 +453,6 @@ AdvancedLoggerGetLoggerInfo (
   UINTN                             Pages;
   CONST EFI_PEI_SERVICES            **PeiServices;
   EFI_STATUS                        Status;
-  EFI_MEMORY_TYPE                   Type;
   ADVANCED_LOGGER_MESSAGE_ENTRY_V2  *LogEntry;
 
   // Try to do the minimum work at the start of this function as this
@@ -554,13 +553,11 @@ AdvancedLoggerGetLoggerInfo (
       //
 
       Pages =  FixedPcdGet32 (PcdAdvancedLoggerPreMemPages);
-      // This is to avoid the interim buffer being allocated to consume 64KB on ARM64 platforms.
-      Type =  EfiBootServicesData;
 
       BufferSize = EFI_PAGES_TO_SIZE (Pages);
 
       Status = PeiServicesAllocatePages (
-                 Type,
+                 EfiBootServicesData,
                  Pages,
                  &NewLoggerInfo
                  );
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.c
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.c
@@ -553,14 +553,9 @@ AdvancedLoggerGetLoggerInfo (
       // Memory Discovered Ppi.  At that time, the full in memory log buffer is allocated.
       //
 
-      if (FeaturePcdGet (PcdAdvancedLoggerPeiInRAM)) {
-        Pages =  FixedPcdGet32 (PcdAdvancedLoggerPages);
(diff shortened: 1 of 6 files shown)


AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.c
```diff
@@ -553,14 +553,9 @@ AdvancedLoggerGetLoggerInfo (
...
-        Pages =  FixedPcdGet32 (PcdAdvancedLoggerPages);
-        Type  =  EfiRuntimeServicesData;
-      } else {
-        Pages =  FixedPcdGet32 (PcdAdvancedLoggerPreMemPages);
-        // This is to avoid the interim buffer being allocated to consume 64KB on ARM64 platforms.
-        Type =  EfiBootServicesData;
-      }
+      Pages =  FixedPcdGet32 (PcdAdvancedLoggerPreMemPages);
+      // This is to avoid the interim buffer being allocated to consume 64KB on ARM64 platforms.
+      Type =  EfiBootServicesData;
```
os-d: Now that the PCD is gone, you can get rid of the `Type` variable and just use EfiBootServicesData directly in the AllocatePages call.

AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.c
```diff
@@ -623,17 +618,7 @@ AdvancedLoggerGetLoggerInfo (
       Status = PeiServicesInstallPpi (mAdvancedLoggerPpiList);
       ASSERT_EFI_ERROR (Status);
 
-      if (FeaturePcdGet (PcdAdvancedLoggerPeiInRAM)) {
-        LoggerInfo->InPermanentRAM = TRUE;
-        Status                     = MmUnblockMemoryRequest (NewLoggerInfo, Pages);
```
os-d, line 628: Was this the only use of MmUnblockMemoryRequest? If so, we can get rid of that header and unlink the lib in the inf.

os-d (approved):

AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.c
Raymond-MS: This function call is used one other time in this file at around line # 240. So it looks like we will need to keep it.

AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.c
Raymond-MS: Good catch, fixed.

makubacki:
I noticed you created the PR from the `release/202311` branch on your fork. That's okay, but You might want to use separate branches for each change to keep you tracking branch to the upstream clean. That makes it easier to sync it in the future. Up to your preferred workflow in the end.

PR #555, opened 2024-08-22
AdvLoggerPkg: Add GoogleTest for AdvancedLoggerPeiLib

VivianNK:
## Description
Add GoogleTest for AdvancedLoggerPeiLib.
Add a mock for the AdvancedLoggerPpi to be used in the GoogleTest.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Local CI. Building the test locally.

## Integration Instructions

N/A

### DIFF
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.cpp
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.cpp
@@ -1,0 +1,227 @@
+/** @file AdvancedLoggerPeiLibGoogleTest.cpp
+
+    This file contains the unit tests for the Advanced Logger DXE Library.
+
+    Copyright (c) Microsoft Corporation.
+    SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#include <Library/GoogleTestLib.h>
+#include <GoogleTest/Library/MockPeiServicesLib.h>
+#include <GoogleTest/Ppi/MockAdvancedLogger.h>
+
+extern "C" {
+  #include <Uefi.h>
+  #include <Base.h>
+  #include <Library/BaseLib.h>
+  #include <Library/DebugLib.h>
+  #include <AdvancedLoggerInternal.h>
+  #include "../../AdvancedLoggerCommon.h"
+}
+
+using namespace testing;
+
+/**
+  Test class for AdvancedLoggerWrite Function
+**/
+class AdvancedLoggerWriteTestPei : public Test {
+protected:
+  StrictMock<MockPeiServicesLib> PeiServicesMock;
+  StrictMock<MockAdvancedLoggerPpi> AdvancedLoggerPpiMock;
+  UINTN DebugLevel;
+  UINTN Instance;
+  CHAR8 *Buffer;
+  UINTN NumberOfBytes;
+
+  void
+  SetUp (
+    ) override
+  {
+    CHAR8  OutputBuf[] = "MyUnitTestLog";
+
+    NumberOfBytes     = sizeof (OutputBuf);
+    Buffer            = OutputBuf;
+    DebugLevel        = DEBUG_ERROR;
+    Instance          = 0;
+    gALPpi->Signature = ADVANCED_LOGGER_PPI_SIGNATURE;
+    gALPpi->Version   = ADVANCED_LOGGER_PPI_VERSION;
+  }
+};
+
+TEST_F (AdvancedLoggerWriteTestPei, PpiNotFounfFailure) {
+  // Expect the locate PPI call to fail
+  EXPECT_CALL (
+    PeiServicesMock,
+    PeiServicesLocatePpi (
+      BufferEq (&gAdvancedLoggerPpiGuid, sizeof (EFI_GUID)),
+      Eq (Instance),
+      IsNull (),
+      NotNull ()
+      )
+    )
+    .WillOnce (
+       Return (EFI_NOT_FOUND)
+       );
+
+  AdvancedLoggerWrite (DebugLevel, Buffer, NumberOfBytes);
+}
+
+TEST_F (AdvancedLoggerWriteTestPei, AdvLoggerWriteSuccess) {
+  // Expect the call to LocatePpi to return success, "found" AdvancedLoggerPpi (mocked)
+  EXPECT_CALL (
+    PeiServicesMock,
+    PeiServicesLocatePpi (
+      BufferEq (&gAdvancedLoggerPpiGuid, sizeof (EFI_GUID)),
+      Eq (Instance),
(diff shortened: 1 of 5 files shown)


AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.inf
```diff
@@ -0,0 +1,42 @@
...
+
+[LibraryClasses]
+  GoogleTestLib
+  AdvancedLoggerLib
+
+[Ppis]
+  gAdvancedLoggerPpiGuid                                    ## CONSUMES
+
+[Pcd]
+  gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerHdwPortDebugPrintErrorLevel       ## SOMETIMES_CONSUMES
```
VivianNK: @apop5 When is this PCD sometimes consumed? I copied it over from the Lib inf.

AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.cpp
```diff
@@ -0,0 +1,225 @@
...
+      )
+    )
+    .WillOnce (
+       DoAll (
+         SetArgPointee<3>(ByRef (gALPpi)),
+         Return (EFI_SUCCESS)
+         )
+       );
+
+  // TODO - why do we still expect a call to AdvancedLoggerWritePpi? shouldn't we exit?
```
VivianNK: @apop5 Is this expected behavior?

AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.cpp
```diff
@@ -0,0 +1,225 @@
+/** @file AdvancedLoggerPeiLibGoogleTest.cpp
+
+    This file contains the unit tests for the Advanced Logger DXE Library.
```
VivianNK:
```suggestion
    This file contains the unit tests for the Advanced Logger PEI Library.

```

AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.inf
apop5: image


Consider the case where there are multiple firmware images that were compiled separately. 

i.e. one project's DSC was compiled and then another project's DSC was compiled, and that between the two, they might result is a mismatch of the versions.

If the reported version of advanced logger in the system is newer than ADVANCED_LOGGER_HW_LVL_VER, it will not be used. Otherwise it will be used.

AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.cpp
apop5: image

With asserts disabled, it will still call. I agree with your question, and I think that if the singatures are not valid, we should just exit without writing

@kuqin12 @os-d Do you have any problem with the system not writing when the signatures are bad?

apop5 (approved):

AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.cpp
os-d: No, I agree if the signature is bad, we shouldn't write there, we would be corrupting memory.

AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.inf
```diff
@@ -0,0 +1,42 @@
...
+#
+#  Google Test for PEI instance of the Advanced Logger library.
+#
+#  Copyright (c) Microsoft Corporation.
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+[Defines]
+  INF_VERSION                    = 1.26
```
os-d: nit: 1.29 is the current version, might as well use that.

AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.inf
```diff
@@ -0,0 +1,42 @@
...
+  BASE_NAME                      = AdvancedLoggerPeiLibGoogleTest
+  FILE_GUID                      = 95196693-DA35-4775-887C-1FF9C65D5054
+  MODULE_TYPE                    = HOST_APPLICATION
+  VERSION_STRING                 = 1.0
+
+#
+# The following information is for reference only and not required by the build tools.
+#
+#  VALID_ARCHITECTURES           = IA32 X64 AARCH64
```
os-d: nit: GoogleTest in edk2 doesn't actually support AARCH64

os-d (approved):

AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.inf
VivianNK: Should the test inf not match the inf it's testing's version as a general rule of thumb? Or always use the most current?

AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.cpp
VivianNK: Changed in this PR: https://github.com/microsoft/mu_plus/pull/565

AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/GoogleTest/AdvancedLoggerPeiLibGoogleTest.inf
os-d: Always use the current.

PR #565, opened 2024-08-28
AdvLoggerPkg: Exit from write call if signature mismatch

VivianNK:
## Description

Current behavior allows the PPI to be used if there is a signature or version mismatch. This fix returns from the function to exit gracefully and prevent memory corruption. 

Create a bug to track this behavior elsewhere in Advanced Logger Library: https://github.com/microsoft/mu_plus/issues/567

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Local CI and unit test.

## Integration Instructions

N/A

### DIFF
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/AdvancedLoggerLib.c
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/AdvancedLoggerLib.c
@@ -43,8 +43,15 @@ AdvancedLoggerWrite (
              );
 
   if (Status == EFI_SUCCESS) {
-    ASSERT (AdvancedLoggerPpi->Signature == ADVANCED_LOGGER_PPI_SIGNATURE);
-    ASSERT (AdvancedLoggerPpi->Version == ADVANCED_LOGGER_PPI_VERSION);
+    if ((AdvancedLoggerPpi->Signature != ADVANCED_LOGGER_PPI_SIGNATURE) ||
+        (AdvancedLoggerPpi->Version != ADVANCED_LOGGER_PPI_VERSION))
+    {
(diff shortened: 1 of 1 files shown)


AdvLoggerPkg/Library/AdvancedLoggerLib/Pei/AdvancedLoggerLib.c
```diff
@@ -46,6 +46,13 @@ AdvancedLoggerWrite (
     ASSERT (AdvancedLoggerPpi->Signature == ADVANCED_LOGGER_PPI_SIGNATURE);
     ASSERT (AdvancedLoggerPpi->Version == ADVANCED_LOGGER_PPI_VERSION);
 
+    if ((AdvancedLoggerPpi->Signature != ADVANCED_LOGGER_PPI_SIGNATURE) ||
+        (AdvancedLoggerPpi->Version != ADVANCED_LOGGER_PPI_VERSION))
+    {
+      // Signature/Version mismatch, don't attempt to use the PPI.
```
os-d, line 52: We can move the asserts inside the if statement here so we don't check the condition twice except on the failure case.

os-d (approved):
Do the other libs need this same behavior for the protocol? Do the unit tests need updating?

VivianNK:
> Do the other libs need this same behavior for the protocol? Do the unit tests need updating?

@os-d Dxe library was fixed to not rely on asserts, and then its unit test was updated before being merged. Planning on doing the same with Pei Lib.
 I can add an issue to check the rest of the libs?

apop5 (approved):

PR #640, opened 2025-03-04
Fix infinite loop issue in UsbGetFullHidDescriptor

NishanthSanjeevi:
## Description

UsbGetFullHidDescriptor does not check the Descriptor header length and this might result in an infinite loop if a bad descriptor is passed.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Surface devices

## Integration Instructions

N/A

### DIFF
--- a/HidPkg/UsbHidDxe/UsbHidDxe.c
+++ b/HidPkg/UsbHidDxe/UsbHidDxe.c
@@ -697,6 +697,12 @@ UsbGetFullHidDescriptor (
       }
     }
 
+    // Check if the descriptor length is 0
+    if (DescriptorHeader->Len == 0) {
+      DEBUG ((DEBUG_ERROR, "[%a] Descriptor length is 0\n", __FUNCTION__));
+      break;
+    }
+
     // move to next descriptor
     DescriptorCursor += DescriptorHeader->Len;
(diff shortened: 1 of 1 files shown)


HidPkg/UsbHidDxe/UsbHidDxe.c
```diff
@@ -697,6 +697,12 @@ UsbGetFullHidDescriptor (
       }
     }
 
+    // Check if the descriptor length is 0
+    if (DescriptorHeader->Len == 0) {
+      DEBUG ((DEBUG_ERROR, "[%a] Descriptor length is 0\n", __FUNCTION__));
```
makubacki: We're trying to use `__func__` for new changes.

```suggestion
      DEBUG ((DEBUG_ERROR, "[%a] Descriptor length is 0\n", __func__));
```

makubacki (approved):

joschock (approved):
I don't think this should happen with a USB-compliant descriptor buffer, but I think this is a good defense mechanism against buggy USB implementations.

makubacki:
> I don't think this should happen with a USB-compliant descriptor buffer, but I think this is a good defense mechanism against buggy USB implementations.

I agree. Programmatically this case should be handled.

apop5 (approved):

kuqin12 (approved):

PR #662, opened 2025-04-02
Updating MuSecureBootKeySelectorLib to detect MS Serviced

Flickdm:
## Description

MuSecureBootKeySelectorLib is being updated to detect if the keystore doesn't match one of our expected
configurations but does match something MSFT could have serviced. 

This detection logic will facilitate a change to the UEFI frontpage to look something like the following when a configuration of secure boot keys does not match our configurations but does match certificates we would support.

Picture1

For details on how to complete these options and their meaning refer to [CONTRIBUTING.md](https://github.com/microsoft/mu/blob/HEAD/CONTRIBUTING.md).

- [ ] Impacts functionality?
- [ ] Impacts security?
- [X] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?
- [ ] Backport to release branch?

## How This Was Tested

Ran on a Physical Machine 

## Integration Instructions

N/A

### DIFF
--- a/MsCorePkg/Library/MuSecureBootKeySelectorLib/MuSecureBootKeySelectorLib.c
+++ b/MsCorePkg/Library/MuSecureBootKeySelectorLib/MuSecureBootKeySelectorLib.c
@@ -17,9 +17,239 @@
 #include <Library/MemoryAllocationLib.h>            // AllocateZeroPool, etc.
 #include <Library/DebugLib.h>                       // Tracing
 #include <Library/UefiRuntimeServicesTableLib.h>    // gRT
+#include <Library/UefiBootServicesTableLib.h>       // gBS
 #include <Library/SecureBootVariableLib.h>          // Secure Boot Variables Operations
 #include <Library/MuSecureBootKeySelectorLib.h>     // Our header
 #include <Library/SecureBootKeyStoreLib.h>          // GetPlatformKeyStore
+#include <Protocol/Hash2.h>                         // Hash2 Protocol
+#include <Protocol/ServiceBinding.h>                // Service Binding Protocol
+
+//
+// These represent the SHA256 hashes of the Microsoft certificates that could potentially be
+// used in the Secure Boot configuration. A platform may have any one or more of these
+// certificates in the Secure Boot configuration depending on if the platform is carrying
+// Legacy Microsoft keys or Microsoft Serviced keys.
+//
+UINT8  mMicrosoftCertificates[][sizeof (EFI_SHA256_HASH2)] = {
+  { // Microsoft UEFI Certificate 2011
+    0x48, 0xE9, 0x9B, 0x99, 0x1F, 0x57, 0xFC, 0x52, 0xF7, 0x61, 0x49, 0x59, 0x9B, 0xFF, 0x0A, 0x58,
+    0xC4, 0x71, 0x54, 0x22, 0x9B, 0x9F, 0x8D, 0x60, 0x3A, 0xC4, 0x0D, 0x35, 0x00, 0x24, 0x85, 0x07
+  },
+  { // Microsoft Windows First Party Certificate 2011
+    0xE8, 0xE9, 0x5F, 0x07, 0x33, 0xA5, 0x5E, 0x8B, 0xAD, 0x7B, 0xE0, 0xA1, 0x41, 0x3E, 0xE2, 0x3C,
+    0x51, 0xFC, 0xEA, 0x64, 0xB3, 0xC8, 0xFA, 0x6A, 0x78, 0x69, 0x35, 0xFD, 0xDC, 0xC7, 0x19, 0x61
+  },
+  { // Microsoft Option ROM Certificate 2023
+    0xE5, 0xBE, 0x3E, 0x64, 0xC6, 0xE6, 0x6A, 0x28, 0x14, 0x57, 0xEC, 0xDE, 0xCE, 0x0D, 0x6D, 0x07,
+    0x87, 0x57, 0x7A, 0xAD, 0x2A, 0x3A, 0x01, 0x44, 0x26, 0x2C, 0x10, 0xC1, 0x4B, 0xA8, 0xD8, 0xF1
+  },
+  { // Microsoft UEFI Certificate 2023
+    0xF6, 0x12, 0x4E, 0x34, 0x12, 0x5B, 0xEE, 0x3F, 0xE6, 0xD7, 0x9A, 0x57, 0x4E, 0xAA, 0x7B, 0x91,
+    0xC0, 0xE7, 0xBD, 0x9D, 0x92, 0x9C, 0x1A, 0x32, 0x11, 0x78, 0xEF, 0xD6, 0x11, 0xDA, 0xD9, 0x01
+  },
+  { // Microsoft Windows First Party Certificate 2023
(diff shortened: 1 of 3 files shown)


Flickdm:
I'm not convinced that a list of sha256s hashes is the cleanest approach and I'm open to ideas. I briefly considered SecureBootObjects keeping a list of previous hashes that sets a PCD that I read from.

apop5 (approved):

MsCorePkg/Library/MuSecureBootKeySelectorLib/MuSecureBootKeySelectorLib.c
```diff
@@ -17,9 +17,228 @@
...
+  @retval EFI_OUT_OF_RESOURCES    Memory allocation failed during the operation.
+  @retval Other                   An error occurred during the operation.
+
+**/
+EFI_STATUS
+EFIAPI
+IsMsftServicedConfig (
+  UINT8    *DbVar,
+  UINTN    DbVarSize,
+  BOOLEAN  *IsMsServicedConfig
```
makubacki, line 87: Please add directional specifiers (`IN`, `OUT`, etc.).

MsCorePkg/Library/MuSecureBootKeySelectorLib/MuSecureBootKeySelectorLib.c
```diff
@@ -17,9 +17,228 @@
...
+  UINTN                         CertCount;
+  UINT8                         *RootCert;
+  UINTN                         RootCertSize;
+  UINTN                         Index;
+  UINTN                         CertIndex;
+  EFI_SERVICE_BINDING_PROTOCOL  *Hash2ServiceBinding;
+  EFI_HASH2_PROTOCOL            *Hash2Protocol;
+  EFI_HASH2_OUTPUT              HashOutput;
+
+  if (IsMsServicedConfig != NULL) {
```
makubacki: If `IsMsServicedConfig` is optional, please mark it `OPTIONAL` in the parameter list.

MsCorePkg/Library/MuSecureBootKeySelectorLib/MuSecureBootKeySelectorLib.c
```diff
@@ -17,9 +17,228 @@
...
+    *IsMsServicedConfig = FALSE;
+  }
+
+  Status = gBS->LocateProtocol (&gEfiHash2ProtocolGuid, NULL, (VOID **)&Hash2Protocol);
```
makubacki: If the platform publishes `gEfiHash2ServiceBindingProtocolGuid` that means they're expecting you to use it create a hash protocol child instance for use here, so shouldn't that be checked first and then fall back on a default `gEfiHash2ProtocolGuid` if it's not available?

MsCorePkg/Library/MuSecureBootKeySelectorLib/MuSecureBootKeySelectorLib.c
```diff
@@ -17,9 +17,228 @@
...
+                    );
+    if (EFI_ERROR (Status) || (Hash2ServiceBinding == NULL) || (Hash2ServiceBinding->CreateChild == NULL)) {
+      Status = EFI_UNSUPPORTED;
+      goto Exit;
+    }
+
+    //
+    // Create an instance of the hash protocol for this controller.
+    //
+    Status = Hash2ServiceBinding->CreateChild (Hash2ServiceBinding, &mHash2ServiceHandle);
```
makubacki: If a child instance is created on this handle, you still need to open the protocol on that handle to assign `Hash2Protocol` in this case, right?

MsCorePkg/Library/MuSecureBootKeySelectorLib/MuSecureBootKeySelectorLib.c
```diff
@@ -17,9 +17,228 @@
...
+      Status = EFI_UNSUPPORTED;
+      goto Exit;
+    }
+  }
+
+  //
+  // Check if the current db matches a Microsoft Serviced Key.
+  // walk each entry in the DbVar and check if it matches the MS Serviced Key.
+  //
+  BOOLEAN  IsMsftCertificate = FALSE;
```
makubacki: We're not supposed to declare variables outside the beginning of the function block per convention. I'd move everything above here into a separate function whose responsibility is to get the hash protocol. If that works, then proceed into this logic.

MsCorePkg/Library/MuSecureBootKeySelectorLib/MuSecureBootKeySelectorLib.c
```diff
@@ -17,9 +17,228 @@
...
+      CertCount = (CertList->SignatureListSize - sizeof (EFI_SIGNATURE_LIST) - CertList->SignatureHeaderSize) / CertList->SignatureSize;
+
+      IsMsftCertificate = FALSE;
+      for (Index = 0; Index < CertCount; Index++) {
+        //
+        // Iterate each Signature Data Node within this CertList for verify.
+        //
+        RootCert     = Cert->SignatureData;
+        RootCertSize = CertList->SignatureSize - sizeof (EFI_GUID);
+        Status       = Hash2Protocol->HashInit (Hash2Protocol, &gEfiHashAlgorithmSha256Guid);
```
makubacki, line 174: Is this meant to be hardcoded to SHA256?

MsCorePkg/Library/MuSecureBootKeySelectorLib/MuSecureBootKeySelectorLib.c
```diff
@@ -17,9 +17,228 @@
...
+    }
+
+    DbVarSize -= CertList->SignatureListSize;
+    CertList   = (EFI_SIGNATURE_LIST *)((UINT8 *)CertList + CertList->SignatureListSize);
+  }
+
+  //
+  // If we get here than we can assume that the current db matches some configuration of MS Serviced Keys.
+  //
+  if (IsMsServicedConfig != NULL) {
```
makubacki, line 222: Should the logic around `IsMsftCertificate` be reworked to factor in here? Could passing a `DbVarSize` of `0` cause this condition to hit (skipping the loop)? Is that expected?

MsCorePkg/Library/MuSecureBootKeySelectorLib/MuSecureBootKeySelectorLib.c
```diff
@@ -17,9 +17,228 @@
...
+    0x07, 0x6F, 0x1F, 0xEA, 0x90, 0xAC, 0x29, 0x15, 0x5E, 0xBF, 0x77, 0xC1, 0x76, 0x82, 0xF7, 0x5F,
+    0x1F, 0xDD, 0x1B, 0xE1, 0x96, 0xDA, 0x30, 0x2D, 0xC8, 0x46, 0x1E, 0x35, 0x0A, 0x9A, 0xE3, 0x30
+  }
+};
+
+//
+// This is the handle for the Hash2ServiceBinding Protocol instance this driver produces
+// if the platform does not provide one.
+//
+EFI_HANDLE  mHash2ServiceHandle = NULL;
```
makubacki: Does this need to be a global?

VivianNK (approved):

PR #675, opened 2025-05-05
AdvLoggerPkg: Fix timestamp consistency in LineParserTestApp

apop5:
## Description

LineParserTestApp makes use of InternalGetPerformanceCounter for generating timestamps. The value returned is a fixed value, but additional testing was being performed on the conversion of the time, to make it correspond to what was actually possible based on the performance counter's frequency.

This started failing when when optimized, causing UT_ASSERT_TRUE to fail. When UT_ASSERT_TRUE fails, it returned UNIT_TEST_ERROR_TEST_FAILED, instead of the expected UINT64 value that the InternalGetPerformanceCounter was expected, causing a cascade of failures.

Remove the check on the conversion of the test's build defined value for timestamp. Updated the expected strings to use the timestamp value as calculated from the conversion from timestampe to ticks (note the 11 vs the 12).

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Ran mu_tiano_platform testing and verifying that Audit Tests pass with changes.

## Integration Instructions
No Integration necessary.

### DIFF
--- a/AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
+++ b/AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
@@ -58,47 +58,47 @@ CHAR8  *InternalMemoryLog[] = {
 // The following text represents the output lines from the line parser given the above input
 
 /* spell-checker: disable */
-CHAR8  Line00[] = "09:06:45.012 : [INFO] First normal test line\n";
-CHAR8  Line01[] = "09:06:45.012 : [ERR ] The QueryMode() function returns information for an available graphics mod\n";
-CHAR8  Line02[] = "09:06:45.012 : [ERR ] e that the graphics device and the set of active video output devices supp\n";
-CHAR8  Line03[] = "09:06:45.012 : [ERR ] orts.\n";
-CHAR8  Line04[] = "09:06:45.012 : [ERR ] If ModeNumber is not between 0 and MaxMode - 1, then EFI_INVALID_PARAMETER\n";
-CHAR8  Line05[] = "09:06:45.012 : [ERR ]  is returned.\n";
-CHAR8  Line06[] = "09:06:45.012 : [INFO] MaxMode is available from the Mode structure of the EFI_GRAPHICS_OUTPUT_PR\n";
-CHAR8  Line07[] = "09:06:45.012 : [INFO] OTOCOL.\n";
-CHAR8  Line08[] = "09:06:45.012 : [ERR ] The size of the Info structure should never be assumed and the value of Si\n";
-CHAR8  Line09[] = "09:06:45.012 : [ERR ] zeOfInfo is the only valid way to know the size of Info.\n";
-CHAR8  Line10[] = "09:06:45.012 : [ERR ] \n";
-CHAR8  Line11[] = "09:06:45.012 : [ERR ] If the EFI_GRAPHICS_OUTPUT_PROTOCOL is installed on the handle that repres\n";
-CHAR8  Line12[] = "09:06:45.012 : [INFO] ents a single video output device, then the set of modes returned by this \n";
-CHAR8  Line13[] = "09:06:45.012 : [ERR ] service is the subset of modes supported by both the graphics controller a\n";
-CHAR8  Line14[] = "09:06:45.012 : [ERR ] nd the video output device.\n";
-CHAR8  Line15[] = "09:06:45.012 : [ERR ] \n";
-CHAR8  Line16[] = "09:06:45.012 : [ERR ] If the EFI_GRAPHICS_OUTPUT_PROTOCOL is installed on the handle that repres\n";
-CHAR8  Line17[] = "09:06:45.012 : [ERR ] ents a combination of video output devices, then the set of modes returned\n";
-CHAR8  Line18[] = "09:06:45.012 : [INFO]  by this service is the subset of modes supported by the graphics controll\n";
-CHAR8  Line19[] = "09:06:45.012 : [ERR ] er and all of the video output devices represented by the handle.\n";
+CHAR8  Line00[] = "09:06:45.011 : [INFO] First normal test line\n";
+CHAR8  Line01[] = "09:06:45.011 : [ERR ] The QueryMode() function returns information for an available graphics mod\n";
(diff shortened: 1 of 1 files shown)


AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
```diff
@@ -58,47 +58,47 @@ CHAR8  *InternalMemoryLog[] = {
...
+CHAR8  Line10V2[] = "09:06:45.011 : [DXE  ] [ERR ] \n";
+CHAR8  Line11V2[] = "09:06:45.011 : [DXE  ] [ERR ] If the EFI_GRAPHICS_OUTPUT_PROTOCOL is installed on the handle that repres\n";
+CHAR8  Line12V2[] = "09:06:45.011 : [DXE  ] [INFO] ents a single video output device, then the set of modes returned by this \n";
+CHAR8  Line13V2[] = "09:06:45.011 : [DXE  ] [ERR ] service is the subset of modes supported by both the graphics controller a\n";
+CHAR8  Line14V2[] = "09:06:45.011 : [DXE  ] [ERR ] nd the video output device.\n";
+CHAR8  Line15V2[] = "09:06:45.011 : [DXE  ] [ERR ] \n";
+CHAR8  Line16V2[] = "09:06:45.011 : [DXE  ] [ERR ] If the EFI_GRAPHICS_OUTPUT_PROTOCOL is installed on the handle that repres\n";
+CHAR8  Line17V2[] = "09:06:45.011 : [DXE  ] [ERR ] ents a combination of video output devices, then the set of modes returned\n";
+CHAR8  Line18V2[] = "09:06:45.011 : [DXE  ] [INFO]  by this service is the subset of modes supported by the graphics controll\n";
+CHAR8  Line19V2[] = "09:06:45.011 : [DXE  ] [ERR ] er and all of the video output devices represented by the handle.\n";
```
kuqin12, line 101: Are these changes necessary?

kuqin12 (approved):

os-d (approved):

AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
apop5: When optimizations are on, the resulting string is:

[INFO] = 09:06:45.012 : [INFO] First normal test line = 

When optimiatons are off, the resulting string is:

[INFO] = 09:06:45.011 : [INFO] First normal test line = 

the compiler seems to be optimizing the math operations away, which is the change in the bits.

AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
makubacki, line 1: Are you intentionally not backporting?

AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
makubacki: What is specifically meant by "optimizations" in this context? Can you please include that detail in the PR description?

AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
makubacki, line 1: Do we really need to rely upon the frequency to come from a TimerLib instance here? Can that be fixed too?

```c
Frequency = GetPerformanceCounterProperties (NULL, NULL);
```

AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
apop5: Yes,

A thanks to Kun for talking through this with me.

The observed behavior was that the unit tests were failing, none of the tests, starting with `InitializeInMemoryLog`

`InitializeInMemoryLog` was failing because `InternalGetPerformanceCounter` failed the 
`UT_ASSERT_TRUE (TimeInMs == (TimeInNs / 1000000u));`.

UT_ASSERT_TRUE, should not have been used in this context. It was being treated as an assert by the author, but in reality it contains a return statement, returning `UNIT_TEST_ERROR_TEST_FAILED`, which the calling code ended up treating as a UINT64 value. 

Once that was cleaned up, the unit test then started failing with 
`expected [INFO] = 09:06:45.012 : [INFO] First normal test line =`
`actual [INFO] = 09:06:45.011 : [INFO] First normal test line =`

When digging into this, I added pragma optimize off around the function, and the function started passing again.
This lead me to believe compiler optimization was reducing the math operations that were occurring in calculating the ticks from the time/frequency.

Looking closer at the Ticks calculation, the calculation is incorrect, they are taking the remainder from the divide operation and adding it as if it were significant. i.e. the remainder would be n ticks, and they were adding it as if it were (1000 * n).

Once that was fixed, the Ticks started being calculated correctly, but the resulting actual value was still 09:06:45.011 instead of the expected 09:06:45.012.

Given the conversion between time -> ticker -> time, this is expected. The TimerLib is not considered a high performance timer, but is meant to be as approximate as possible given the timer source (in this case the PM timer).

makubacki: There is enough going on here to mark this as a bug fix.

AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
apop5: It needs to be backported, but I wanted to double check that muautomation bot is setup correct (there is a PR in basecore to bring the functionality to the 202405 branch)

apop5: AdvancedLoggerAccessLib is relying on GetTimeInNanoSecond()

It might be possible to gut out any dependency on the TImerLib library.

makubacki: I suggest we try to do that.

apop5: Investigated, and its not possible to drop the TimerLib. AdvLoggerAccessLib calls out the dependency for TimerLib, which results in the platform specified instance being pulled in.

PR #676, opened 2025-05-06
AdvLoggerPkg: Fix timestamp consistency in LineParserTestApp

apop5:
## Description

LineParserTestApp makes use of InternalGetPerformanceCounter for generating timestamps. The value returned is a fixed value, but additional testing was being performed on the conversion of the time, to make it correspond to what was actually possible based on the performance counter's frequency.

This started failing when when optimized, causing UT_ASSERT_TRUE to fail. When UT_ASSERT_TRUE fails, it returned
UNIT_TEST_ERROR_TEST_FAILED, instead of the expected UINT64 value that the InternalGetPerformanceCounter was expected, causing a cascade of failures.

Remove the check on the conversion of the test's build defined value for timestamp. Updated the expected strings to use the timestamp value as calculated from the conversion from timestampe to ticks (note the 11 vs the 12).

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Ran mu_tiano_platform testing and verifying that Audit Tests pass with changes.

## Integration Instructions
No Integration necessary.

### DIFF
--- a/AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
+++ b/AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
@@ -58,47 +58,47 @@ CHAR8  *InternalMemoryLog[] = {
 // The following text represents the output lines from the line parser given the above input
 
 /* spell-checker: disable */
-CHAR8  Line00[] = "09:06:45.012 : [INFO] First normal test line\n";
-CHAR8  Line01[] = "09:06:45.012 : [ERR ] The QueryMode() function returns information for an available graphics mod\n";
-CHAR8  Line02[] = "09:06:45.012 : [ERR ] e that the graphics device and the set of active video output devices supp\n";
-CHAR8  Line03[] = "09:06:45.012 : [ERR ] orts.\n";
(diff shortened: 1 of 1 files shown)


os-d (approved):

makubacki (approved):

TpmTestingPkg/Overrides/Tcg2Dxe/Tcg2Dxe.inf
makubacki, line 1: Are these TPM changes meant to be in this PR?

apop5: The [Tpm2DebugLib ](https://github.com/microsoft/mu_tiano_plus/pull/405) changes in mu_tiano_plus, broke UefiTestingPkg and TpmTestingPkg.

This changes are here, temporarily, because this is the first PR that attempted to go into mu_plus 202502 since the other changes went into mu_tiano_plus.

I think we are waiting on a ruling for how to correctly address this.

makubacki: Spoke to Oliver offline and he'll follow up.

PR #720, opened 2025-07-10
UefiTestingPkg: Update Paging Audit Static Buffer Size

os-d:
## Description

Add a few more 0's and hope it works.

Less glib: the larger buffer size was still being overrun with new changes in SBSA, so adding much more size to the buffer.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Running unit tests in SBSA.

## Integration Instructions

N/A.

### DIFF
--- a/UefiTestingPkg/AuditTests/PagingAudit/UEFI/PagingAuditCommon.h
+++ b/UefiTestingPkg/AuditTests/PagingAudit/UEFI/PagingAuditCommon.h
@@ -32,7 +32,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Guid/DebugImageInfoTable.h>
 #include <Guid/MemoryAttributesTable.h>
 
-#define    MEM_INFO_DATABASE_REALLOC_CHUNK    0x10000
+#define    MEM_INFO_DATABASE_REALLOC_CHUNK    0x10000000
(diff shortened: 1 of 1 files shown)


apop5:
Having a static size seems to be causing problems. 

Is there the possibility of determining the allocation size required? It might prevent needing to modify this again the future, if it is possible.

makubacki:
Also, can you run the final solution through the mu_tiano_platforms pipelines to verify the result (if not able to repro locally)?

apop5 (approved):

os-d:
> Having a static size seems to be causing problems.
> 
> Is there the possibility of determining the allocation size required? It might prevent needing to modify this again the future, if it is possible.

Unfortunately not. The test does dynamically allocate memory when it determines it is allowed to, but once it is in the state of analyzing the page table, it doesn't want to make more allocations that then change the page table.

I will note that this test is not continuing forward into other iterations of the memory protections code, those will be written in that native environment and not suffer these same issues because of a different architecture.

os-d:
> Also, can you run the final solution through the mu_tiano_platforms pipelines to verify the result (if not able to repro locally)?

It passed CI: https://github.com/microsoft/mu_tiano_platforms/pull/1179.

makubacki (approved):

PR #738, opened 2025-08-18
AdvLoggerPkg: Add semaphore for accessing serial port for messages

apop5:
## Description

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?
- [ ] Backport to release branch?

## How This Was Tested

<_Describe the test(s) that were run to verify the changes._>

## Integration Instructions

<_Describe how these changes should be integrated. Use N/A if nothing is required._>


apop5:
based on conversations, serial port debug messages are not thread safe, and we should not attempt to make this thread safe.

AP processes, when spun up, should not allowed to use the debug macro.

PR #775, opened 2025-10-23
Perf Services Shell App

berlin-with0ut-return:
## Description

Adds a new performance UEFI shell app that benchmarks basic services. The number of calls and functions used are derived from here: https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/background/memory_safety_strategy.md#service-call-coverage

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Runs and gives reasonably accurate results on Q35 and Intel platforms.

## Integration Instructions
This app runs and is compiled independently and does not have to live in mu_plus. This PR is to simply encourage reviews.


UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+                    "| {:<30} | {:>14} | {:>12} | {:>15} |",
+                    bench_name,
+                    cycles,
+                    num_calls,
+                    cycles / num_calls as u64
+                )
+                .map_err(|e| BenchError::WriteFailure("Write table header failed", e))?;
+            }
+            Err(e) => {
+                debugln!(DEBUG_ERROR, "Benchmark {} failed: {:?}", bench_name, e);
```
makubacki, line 93: Have you considered showing the failed benchmarks in the actual output results so the information is always available? As opposed to just the debug print and assert?

UefiTestingPkg/PerfTests/TempTest/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+edition = "2024"
+authors = ["Microsoft"]
+
+[[bin]]
+name = "TempTest"
+path = "src/main.rs"
+
+[dependencies]
+r-efi = {workspace=true}
+patina_sdk = { version = "4", registry = "patina-fw" }
```
makubacki: If this is kept around, please update to the `patina` crate on crates.io.

.cargo/config.toml
```diff
@@ -1,17 +1,21 @@
...
 ]
 
 [target.aarch64-unknown-uefi]
 rustflags = [
     "-C", "link-arg=/base:0x0",
     "-C", "link-arg=/subsystem:efi_boot_service_driver",
 ]
+
+[registries]
+patina-fw = { index = "sparse+https://pkgs.dev.azure.com/patina-fw/artifacts/_packaging/temporary/Cargo/index/" }
```
makubacki: Please drop this before check-in and use crates.io.

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/bench_fn.rs
makubacki, line 1: To organize this file a little better, you might want to consider making `bench` a module and then have submodule for different categories. Like:

src/
  bench/
    controller.rs
    event.rs
    image.rs
    memory.rs
    protocol.rs
    ...
  bench.rs
  ...

makubacki, line 1: The normal cargo benchmarks include warm iterations and some statistics like min and max. Those might be worth adding.

.cargo/config.toml
```diff
@@ -1,7 +1,7 @@
 [target.x86_64-unknown-uefi]
 rustflags = [
     "-C", "link-arg=/base:0x0",
-    "-C", "link-arg=/subsystem:efi_boot_service_driver",
+    "-C", "link-arg=/subsystem:efi_application",
```
Javagedes, line 4: I don't think you can switch this to to efi_application. That will make the UefiHidDxe drivers build as an efi application instead of a dxe driver.

.cargo/config.toml
```diff
@@ -1,7 +1,7 @@
 [target.x86_64-unknown-uefi]
 rustflags = [
     "-C", "link-arg=/base:0x0",
-    "-C", "link-arg=/subsystem:efi_boot_service_driver",
+    "-C", "link-arg=/subsystem:efi_application",
 ]
 
 [target.i686-unknown-uefi]
```
Javagedes, line 7: I don't see a aarch64 target in this file, which means you won't be able to compile for aarch64 systems.

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/bench/controller.rs
```diff
@@ -0,0 +1,109 @@
...
+        _controller_handle: efi::Handle,
+        _num_children: usize,
+        _child_handle_buffer: *mut efi::Handle,
+    ) -> efi::Status {
+        efi::Status::SUCCESS
+    }
+
+    let controller_handle = unsafe {
+        BOOT_SERVICES
+            .install_protocol_interface_unchecked(None, &TEST_GUID1, 0x1111 as *mut core::ffi::c_void)
```
Javagedes, line 40: I think this handle needs to be the handle from your entry_point, right? Typically, only single instance protocols are installed with a null_ptr as their handle. But I'm not 100% familiar with this. @makubacki can probably confirm this or tell me I'm wrong :)

Same with the others below here.

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/bench/controller.rs
```diff
@@ -0,0 +1,109 @@
+use mu_rust_helpers::perf_timer::{Arch, ArchFunctionality as _};
```
Javagedes, line 1: These modules need documentation in general, but it is also especially good for these benchmarks because it is a place to describe exactly what benchmark here is benchmarking, and how it is doing the benchmark.

I'll only leave one comment regarding this here, but it will apply to all the modules :)

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/bench/event.rs
```diff
@@ -0,0 +1,108 @@
+use core::{ffi::c_void, ptr};
+
+use mu_rust_helpers::perf_timer::{Arch, ArchFunctionality as _};
+use patina::boot_services::{BootServices, event::EventType, tpl::Tpl};
+use r_efi::efi;
+use rolling_stats::Stats;
+
+use crate::{BOOT_SERVICES, error::BenchError};
+
+pub(crate) fn bench_check_event(_handle: efi::Handle, num_calls: usize) -> Result<Stats<f64>, BenchError> {
```
Javagedes, line 10: This benchmark only benchmarks half of what `check_event` does; It only checks the performance when the benchmark is actually signaled.

The more complex route is when when you check and event that has not been signaled. That scenario also needs a test I would say.

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/bench/event.rs
```diff
@@ -0,0 +1,108 @@
...
+        .map_err(|e| BenchError::BenchSetupFailure("Failed to create event", e))?;
+        let start = Arch::cpu_count();
+        BOOT_SERVICES.close_event(event_handle).map_err(|e| BenchError::BenchFailure("Failed to close event", e))?;
+        let end = Arch::cpu_count();
+        stats.update((end - start) as f64);
+    }
+    Ok(stats)
+}
+
+pub(crate) fn bench_signal_event(_handle: efi::Handle, num_calls: usize) -> Result<Stats<f64>, BenchError> {
```
Javagedes, line 84: You also need to write a benchmark for event groups, which results in multiple callbacks when a single event is fired. That gets slower as we have more listeners.

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/bench/image.rs
```diff
@@ -0,0 +1,43 @@
+use mu_rust_helpers::perf_timer::{Arch, ArchFunctionality as _};
```
Javagedes, line 1: was there any need to test `unload_image`?

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/bench/mod.rs
```diff
@@ -0,0 +1,15 @@
+use r_efi::efi;
```
Javagedes, line 1: The `mod.rs` way of doing things is old. We use the newer way, which is like this:

```
|- bench.rs // Your bench/mod.rs
|- bench/*
|   |- misc.rs
|   |- etc.
```

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/bench/tpl.rs
```diff
@@ -0,0 +1,34 @@
...
+use r_efi::efi::{self};
+use rolling_stats::Stats;
+
+use crate::{BOOT_SERVICES, error::BenchError};
+
+pub(crate) fn bench_raise_tpl(_handle: efi::Handle, num_calls: usize) -> Result<Stats<f64>, BenchError> {
+    let mut stats: Stats<f64> = Stats::new();
+    for _ in 0..num_calls {
+        let start = Arch::cpu_count();
+        let old_tpl = BOOT_SERVICES.raise_tpl(Tpl::NOTIFY);
```
Javagedes, line 12: I would probably raise to high to include the cost of disabling interrupts.

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/bench/tpl.rs
```diff
@@ -0,0 +1,34 @@
...
+        BOOT_SERVICES.restore_tpl(old_tpl);
+    }
+
+    Ok(stats)
+}
+
+pub(crate) fn bench_restore_tpl(_handle: efi::Handle, num_calls: usize) -> Result<Stats<f64>, BenchError> {
+    let mut stats: Stats<f64> = Stats::new();
+    for _ in 0..num_calls {
+        let old_tpl = BOOT_SERVICES.raise_tpl(Tpl::NOTIFY);
```
Javagedes, line 25: Probably need to test restoring from all the different TPLs that you can because restoring TPLs causes event notifies to trigger.

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+pub static BOOT_SERVICES: StandardBootServices = StandardBootServices::new_uninit();
+
+pub fn bench_start(handle: efi::Handle, st: *const system::SystemTable) -> Result<(), BenchError> {
+    debugln!(DEBUG_INFO, "Starting Services Benchmark Test...");
+
+    let mut output_buf = String::new();
+
+    // Writes fixed-width markdown table.
+    // Column headers.
+    writeln!(
```
Javagedes, line 47: Id turn these inidivual writes (or some grouped writes) into separate functions just to make this more udnerstandable. e.g.

```rust
self.write_header(&mut output_buf);
let results = self.run_tests()?;
self.write_results(results, &mut output_buf);
```

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    .map_err(|e| BenchError::WriteFailure("Write table header failed", e))?;
+    // Column seperators.
+    writeln!(
+        &mut output_buf,
+        "| {:-<32} | {:-<14} | {:-<12} | {:-<15} | {:-<15} | {:-<12} | {:-<12} | {:-<12} |",
+        "-", "-", "-", "-", "-", "-", "-", "-"
+    )
+    .map_err(|e| BenchError::WriteFailure("Write table header failed", e))?;
+
+    for (bf, num_calls) in BENCH_FNS {
```
Javagedes, line 68: I'd split this up into a separate function and gather the results into a vector of result structs or something, then report the results

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    unsafe { print_to_console(st, &output_buf.as_str()) };
+
+    Ok(())
+}
+
+/// Print a message to the UEFI console output.
+/// SAFETY: Caller must ensure that `system_table` is a valid pointer to a SystemTable.
+pub unsafe fn print_to_console(system_table: *const system::SystemTable, message: &str) {
+    // SAFETY: `system_table` is validated by the caller.
+    let con_out = (unsafe { &*system_table }).con_out;
```
Javagedes, line 124: You need to check that con_out is not a null_ptr.

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    ) -> efi::Status {
+        // Safety: This block is unsafe because it assumes that system_table and (*system_table).boot_services are correct,
+        // and because it mutates/accesses the global BOOT_SERVICES static.
+        unsafe {
+            BOOT_SERVICES.init(&*((*system_table).boot_services));
+            GLOBAL_ALLOCATOR.init((*system_table).boot_services);
+            init_debug((*system_table).boot_services);
+        }
+
+        bench_start(image_handle, system_table);
```
Javagedes, line 37: This function returns a result that you are ignoring.

UefiTestingPkg/PerfTests/TempTest/src/main.rs
```diff
@@ -0,0 +1,62 @@
+//! Temporary test application for benchmarking UEFI boot services.
```
Javagedes, line 1: Do you need this anymore?

UefiTestingPkg/PerfTests/TempTest/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+#
```
Javagedes, line 1: or this?

UefiTestingPkg/PerfTests/TempTest/TempTest.inf
```diff
@@ -0,0 +1,21 @@
+### @file
```
Javagedes, line 1: Or this?

Cargo.toml
```diff
@@ -9,6 +9,8 @@ members = [
   "HidPkg/UefiHidDxeV2",
   "MsCorePkg/HelloWorldRustDxe",
   "MsWheaPkg/Crates/MuTelemetryHelperLib",
+  "UefiTestingPkg/PerfTests/ServicesBenchmarkTest",
+  "UefiTestingPkg/PerfTests/TempTest",
```
Javagedes, line 13: or this?

UefiTestingPkg/PerfTests/TempTest/TempTest.inf
Javagedes, line 1: This needs to be added to the UefiTestpingPkg DSC so that it is compiled during CI.

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/ServicesBenchmarkTest.inf
```diff
@@ -0,0 +1,21 @@
+### @file
```
Javagedes, line 1: This needs to be added to the UefiTestpingPkg DSC so that it is compiled during CI.

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/resources/TempTest.efi
spbrogan, line 1: Need a readme that describes what these are.  We generally don't want binaries checked into git so maybe we could build these and/or just acknowledge that for this test tool we are not following normal practices.

spbrogan (changes requested):
Overall significant documentation is needed to make it really clear what we are benchmarking.  This tools target audience is not a seasoned firmware developer but more likely a test engineer.  We also want to make the output setup for automatic collection, compare, and fully support easy automation.    

I also would suggest that before each section of tests we gather relative metrics that provide context for the test.  For example, open protocol on a system with 10 protocols installed vs a system with 10000 will be different.  Same with number of handles, number of memory map entries, etc.

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/bench/controller.rs
spbrogan: doesn't really matter.  if null core will create new handle.  otherwise will install on handle provided.

berlin-with0ut-return:
This is being moved to https://github.com/OpenDevicePartnership/patina-apps

if your comment has not been addressed I will address it there

UefiTestingPkg/PerfTests/ServicesBenchmarkTest/src/bench/image.rs
berlin-with0ut-return: there are currently no uses in the core so i skipped it for now, but it can be pretty easily added if we want

berlin-with0ut-return: based on https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/background/memory_safety_strategy.md

UefiTestingPkg/PerfTests/TempTest/src/main.rs
berlin-with0ut-return: it is actually used by start_image so it is needed. perhaps TempTest is not the best name

UefiTestingPkg/PerfTests/TempTest/TempTest.inf
berlin-with0ut-return: the Cargo and inf files were for building the test image and are no longer needed

PR #787, opened 2025-11-14
HidPkg/UefiHidDxeV2: Prevent TPL escalation when sending output reports

### LINKED ISSUE #788: [Bug]: HidPkg/UefiHidDxeV2: ERROR - TPL 16 is too high! when update_leds is called

### Current Behavior

The UEFI log reports:
`ERROR - TPL 16 is too high!`

This occurs when `update_leds` is invoked by `simple_text_in_ex_set_state`.
The root cause is that `update_leds` calls `send_led_reports`, which eventually calls the `set_output_report` function of a HidIo instance.
The HidIo implementation may require the TPL to be at or below TPL_CALLBACK, but `simple_text_in_ex_set_state` raises the TPL to TPL_NOTIFY.
This mismatch leads to the error and potential HID I/O conflicts.

### Expected Behavior

`simple_text_in_ex_set_state` should execute without triggering TPL errors, and HID output reports should be sent at the correct TPL level.

### Steps To Reproduce

Invoke simple_text_in_ex_set_state (e.g., toggling LED state) while UEFI booting.
Observe UEFI log for:
`ERROR - TPL 16 is too high!`

### Build Environment

```markdown
- OS(s): Windows 11
- Tool Chain(s): Vs2022
- Targets Impacted: ALL
```

### Version Information

```text
95ea2bdfcd54b81ea50dfe0d4215812257c3104e
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

wenbhou:
## Description

Fixes https://github.com/microsoft/mu_plus/issues/788

Separate `send_output_reports` from `send_led_reports` to avoid issues caused by an elevated TPL when invoked from `simple_text_in_ex_set_state`, `simple_text_in_ex_reset` or `simple_text_in_reset`. These functions raise the TPL to TPL_NOTIFY, but the HID I/O instance may require the TPL to be at or below TPL_CALLBACK.
By decoupling these operations, we ensure proper TPL handling and prevent conflicts during HID output report transmission.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on laptop and no TPL conflicts observed.

## Integration Instructions

N/A

### DIFF
--- a/HidPkg/UefiHidDxeV2/src/keyboard.rs
+++ b/HidPkg/UefiHidDxeV2/src/keyboard.rs
@@ -437,11 +437,10 @@ impl KeyboardHidHandler {
     /// Called to send LED state to the device.
     pub fn send_led_reports(&mut self, hid_io: &dyn HidIo) -> Result<(), efi::Status> {
         let output_reports = self.generate_led_output_reports();
-        if let Err(result) = self.send_output_reports(hid_io, output_reports) {
-            debugln!(DEBUG_ERROR, "unexpected error sending output report: {:?}", result);
-            return Err(result);
-        }
-        Ok(())
+        self.send_output_reports(hid_io, output_reports).map_err(|err| {
+            debugln!(DEBUG_ERROR, "unexpected error sending output report: {:?}", err);
+            err
+        })
     }
 
     /// Returns a clone of the keystroke at the front of the keystroke queue.
--- a/HidPkg/UefiHidDxeV2/src/keyboard.rs
+++ b/HidPkg/UefiHidDxeV2/src/keyboard.rs
@@ -393,15 +393,32 @@ impl KeyboardHidHandler {
         Ok(())
     }
 
+    /// Called to send HID reports to the device.
+    fn send_output_reports(
+        &mut self,
+        hid_io: &dyn HidIo,
+        output_reports: Vec<(Option<ReportId>, Vec<u8>)>,
+    ) -> Result<(), efi::Status> {
+        for (id, output_report) in output_reports {
+            let result = hid_io.set_output_report(id.map(|x| u32::from(x) as u8), &output_report);
+            if let Err(result) = result {
+                debugln!(
+                    DEBUG_ERROR,
+                    "KeyboardHidHandler:set_output_report: unexpected error sending output report: {:?}",
+                    result
+                );
+                return Err(result);
+            }
+        }
+        Ok(())
+    }
+
     /// Resets the keyboard driver state. Clears any pending key state. `extended verification` will also reset toggle
     /// state.
-    pub fn reset(&mut self, hid_io: &dyn HidIo, extended_verification: bool) -> Result<(), efi::Status> {
+    pub fn reset(&mut self, _hid_io: &dyn HidIo, extended_verification: bool) -> Result<(), efi::Status> {
         self.last_keys.clear();
         self.current_keys.clear();
         self.key_queue.reset(extended_verification);
-        if extended_verification {
-            self.send_led_reports(hid_io)?;
-        }
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


HidPkg/UefiHidDxeV2/src/keyboard.rs
```diff
@@ -393,14 +393,34 @@ impl KeyboardHidHandler {
...
+                return Err(result);
+            }
+        }
+        Ok(())
+    }
+
     /// Resets the keyboard driver state. Clears any pending key state. `extended verification` will also reset toggle
     /// state.
-    pub fn reset(&mut self, hid_io: &dyn HidIo, extended_verification: bool) -> Result<(), efi::Status> {
+    pub fn reset(&mut self, _hid_io: &dyn HidIo, extended_verification: bool) -> Result<(), efi::Status> {
```
joschock: consider just removing `_hid_io` as I don't think this is an external API, so that wouldn't be a breaking change.

HidPkg/UefiHidDxeV2/src/keyboard/simple_text_in_ex.rs
```diff
@@ -279,9 +301,18 @@ impl SimpleTextInExFfi {
...
             };
             keyboard_handler.set_key_toggle_state(unsafe { key_toggle_state.read() });
-            keyboard_handler.update_leds(hid_io.as_ref()).err().unwrap_or(efi::Status::SUCCESS)
+            output_reports = keyboard_handler.generate_led_output_reports();
+            keyboard_handler_ref = Some(keyboard_handler);
+            hid_io_ref = Some(hid_io);
+            efi::Status::SUCCESS
         };
         context.boot_services.restore_tpl(old_tpl);
+        if let (Some(keyboard_handler), Some(hid_io)) = (keyboard_handler_ref, hid_io_ref) {
```
joschock, line 312: consider a small comment like `// Avoid sending output reports at the higher TPL`

joschock (approved):

HidPkg/UefiHidDxeV2/src/keyboard.rs
wenbhou: Removed.

HidPkg/UefiHidDxeV2/src/keyboard/simple_text_in_ex.rs
wenbhou: Added. Thanks.

HidPkg/UefiHidDxeV2/src/keyboard.rs
```diff
@@ -416,12 +436,10 @@ impl KeyboardHidHandler {
...
-            if let Err(result) = result {
-                debugln!(DEBUG_ERROR, "unexpected error sending output report: {:?}", result);
-                return Err(result);
-            }
+        let output_reports = self.generate_led_output_reports();
+        if let Err(result) = self.send_output_reports(hid_io, output_reports) {
+            debugln!(DEBUG_ERROR, "unexpected error sending output report: {:?}", result);
+            return Err(result);
         }
         Ok(())
```
makubacki: I think the error handling in the block can be simplified a bit.


```suggestion
        let output_reports = self.generate_led_output_reports();
        self.send_output_reports(hid_io, output_reports).map_err(|err| {
            debugln!(DEBUG_ERROR, "unexpected error sending output report: {:?}", err);
            err
        })
```

makubacki: Even simpler.


```suggestion
        let output_reports = self.generate_led_output_reports();
        self.send_output_reports(hid_io, output_reports).inspect_err(|err| {
            debugln!(DEBUG_ERROR, "unexpected error sending output report: {:?}", err);
        })
```

HidPkg/UefiHidDxeV2/src/keyboard.rs
```diff
@@ -530,7 +547,7 @@ impl HidReportReceiver for KeyboardHidHandler {
     fn initialize(&mut self, controller: efi::Handle, hid_io: &dyn HidIo) -> Result<(), efi::Status> {
```
makubacki, line 546: My understanding is that previously, the `self.reset(hid_io, true)?` call would send a led report:

```rust
        if extended_verification {
            self.send_led_reports(hid_io)?;
        }
```

Now, the led state is just cleared:

```rust
        if extended_verification {
            self.led_state.clear();
        }
```

Is that intentional or should a report be sent after the reset? For example:

```rust
       let descriptor = hid_io.get_report_descriptor()?;
        self.process_descriptor(descriptor)?;
        self.reset(true)?;
        self.send_led_reports(hid_io)?;
        self.install_protocol_interfaces(controller)?;
        self.initialize_keyboard_layout()?;
        Ok(())
    }
```

I thought that still might be useful to keep things consistent after reset.

makubacki (approved):

HidPkg/UefiHidDxeV2/src/keyboard.rs
wenbhou: The report to update keyboard leds is actually sent out in `simple_text_in_ex_reset` and `simple_text_in_reset` now, where the `keyboard_handler.reset()` is called.

makubacki: Ah, okay. I was looking too closely at this particular function.

os-d (approved):

PR #790, opened 2025-11-14
Fixing use after free issue for extra data in WHEA record

kuqin12:
## Description

The current implementation will free the extra data immediately after the entry is logged. This is fine if the log is directly materialized into `HwErrRec` UEFI variables.

However, if this logic occurs in the early DXE stage, it will cause the collected linked list contain dangling pointer.

This change fixed the issue by duplicating the buffer when it is added to the linked list and freed after linked list is processed.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested by adding a telemetry entry before variable service is available and verified that the WHEA driver collected and then later processed the entry properly.

## Integration Instructions

N/A

### DIFF
--- a/MsWheaPkg/MsWheaReport/Dxe/MsWheaReportList.c
+++ b/MsWheaPkg/MsWheaReport/Dxe/MsWheaReportList.c
@@ -28,10 +28,10 @@ CreateNewEntry (
   IN MS_WHEA_ERROR_ENTRY_MD  *MsWheaEntryMD
   )
 {
-  MS_WHEA_LIST_ENTRY  *MsWheaListEntry = NULL;
-  MS_WHEA_ERROR_ENTRY_MD  *TempMsWheaEntryMD = NULL;
-  MS_WHEA_ERROR_EXTRA_SECTION_DATA *ExtraSectionPtr = NULL;
-  UINT32              Index            = 0;
+  MS_WHEA_LIST_ENTRY                *MsWheaListEntry   = NULL;
+  MS_WHEA_ERROR_ENTRY_MD            *TempMsWheaEntryMD = NULL;
+  MS_WHEA_ERROR_EXTRA_SECTION_DATA  *ExtraSectionPtr   = NULL;
+  UINT32                            Index              = 0;
 
   // Input argument sanity check
   if (MsWheaEntryMD == NULL) {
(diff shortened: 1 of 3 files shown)


os-d (approved):

makubacki (approved):

MsWheaPkg/MsWheaReport/Dxe/MsWheaReportList.c
```diff
@@ -54,6 +56,15 @@ CreateNewEntry (
   Index = 0;
   CopyMem (&((UINT8 *)MsWheaListEntry->PayloadPtr)[Index], MsWheaEntryMD, sizeof (MS_WHEA_ERROR_ENTRY_MD));
 
+  TempMsWheaEntryMD = (MS_WHEA_ERROR_ENTRY_MD *)MsWheaListEntry->PayloadPtr;
+  ExtraSectionPtr = (MS_WHEA_ERROR_EXTRA_SECTION_DATA *)(UINTN)(MsWheaEntryMD->ExtraSection);
+  if (ExtraSectionPtr != NULL) {
+    TempMsWheaEntryMD->ExtraSection = (EFI_PHYSICAL_ADDRESS)(UINTN)AllocateCopyPool (
```
makubacki, line 62: I didn't walk the logic all the way through, but does everything work out if `AllocateCopyPool()` fails here (returns `NULL`)?

kuqin12: It would. This allocation contains its own data size and does not intrude into the normal packet reported, which is the normal case when other code reports telemetry without extra data.

kuqin12: And that is checked here: https://github.com/microsoft/mu_plus/blob/4cbdbab1ef42919743c21d61106f775a55850d33/MsWheaPkg/MsWheaReport/MsWheaReportHER.c#L292

It will log other information but the extra data. I think that is probably the best we can do when it comes to out of resources error.

liqiqiii (approved):

cfernald (approved):

PR #799, opened 2025-11-26
AdvLoggerPkg: Fix memory bucket stability

### LINKED ISSUE #778: [Bug]: Advanced Logger Does Not Reallocate Buffer In DXE

### Current Behavior

For platforms that initialize the advanced logger log before DXE, the log buffer does not sit in the runtime services data bucket. This breaks the hibernate expectation that all runtime memory is in the proper buckets. Because platforms don't often change the log size, it may usually go by unnoticed.

Note, the DXE core created memory log is not allocated in runtime services data, it should be.

### Expected Behavior

Advanced logger's log should live in the runtime services data bucket.

### Steps To Reproduce

n/a.

### Build Environment

```markdown
- OS(s):
- Tool Chain(s):
- Targets Impacted:
```

### Version Information

```text
top of tree
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Right now, the PEI Core instance of AdvancedLoggerLib will always allocate the logger buffer as `EfiRuntimeServicesData`. This means it is not allocated into the DXE Core managed RT Services Data memory bucket. The DXE Core instance of AdvancedLoggerLib either continues to use this buffer it is provided or allocates a new `EfiReservedMemoryType` buffer.

This change always allocates a `EfiReservedMemoryType` buffer in the DXE Core instance to ensure it is allocated from the reserved memory type buckets setup by the DXE Core. The PEI Core logger buffer is updated to be `EfiBootServicesData` so it does not affect runtime allocations.

The logger buffer will not attempt to be migrated if `PcdAdvancedLoggerFixedInRAM` is `FALSE`.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- CI
- Boot on a virtual and physical Intel platform and confirm the logger is functional
  - Note: Has not been tested on an ARM64 platform yet

## Integration Instructions

- N/A

### DIFF
--- a/AdvLoggerPkg/Include/AdvancedLoggerInternal.h
+++ b/AdvLoggerPkg/Include/AdvancedLoggerInternal.h
@@ -28,7 +28,7 @@
 // Version 2: New backward incompatible V2 structure (ADVANCED_LOGGER_MESSAGE_ENTRY_V2
 //
 #define ADVANCED_LOGGER_MSG_MAJ_VER  2
-#define ADVANCED_LOGGER_MSG_MIN_VER  0
+#define ADVANCED_LOGGER_MSG_MIN_VER  1
 
 #define ADVANCED_LOGGER_VERSION  ADVANCED_LOGGER_HW_LVL_VER
 
@@ -46,6 +46,11 @@
 #define ADVANCED_LOGGER_PHASE_CNT          11
 
 //
+// The maximum depth to follow when traversing chains of advanced logger info structures
+//
+#define ADVANCED_LOGGER_MAX_LOGGER_CHAIN_DEPTH  3
+
+//
 // These Pcds are used to carve out a PEI memory buffer from the temporary RAM.
 //
 //  PcdAdvancedLoggerBase -        NULL = UEFI starts with PEI, and SEC provides no memory log buffer
@@ -60,25 +65,26 @@
 #pragma pack (push, 1)
 
 typedef volatile struct {
-  UINT32      Signature;                          // Signature 'ALOG'
-  UINT16      Version;                            // Current Version
-  UINT16      Reserved[3];                        // Reserved for future
-  UINT32      LogBufferOffset;                    // Offset from LoggerInfo to start of log, expected to be the size of this structure 8 byte aligned
-  UINT32      Reserved4;
-  UINT32      LogCurrentOffset;                   // Offset from LoggerInfo to where to store next log entry.
-  UINT32      DiscardedSize;                      // Number of bytes of messages missed
-  UINT32      LogBufferSize;                      // Size of allocated buffer
-  BOOLEAN     InPermanentRAM;                     // Log in permanent RAM
-  BOOLEAN     AtRuntime;                          // After ExitBootServices
-  BOOLEAN     GoneVirtual;                        // After VirtualAddressChange
-  BOOLEAN     HdwPortInitialized;                 // HdwPort initialized
-  BOOLEAN     HdwPortDisabled;                    // HdwPort is Disabled
-  BOOLEAN     Reserved2[3];                       //
-  UINT64      TimerFrequency;                     // Ticks per second for log timing
-  UINT64      TicksAtTime;                        // Ticks when Time Acquired
-  EFI_TIME    Time;                               // Uefi Time Field
-  UINT32      HwPrintLevel;                       // Logging level to be printed at hw port
-  UINT32      Reserved3;                          //
+  UINT32                  Signature;              // Signature 'ALOG'
+  UINT16                  Version;                // Current Version
+  UINT16                  Reserved[3];            // Reserved for future
+  UINT32                  LogBufferOffset;        // Offset from LoggerInfo to start of log, expected to be the size of this structure 8 byte aligned
+  UINT32                  Reserved4;
+  UINT32                  LogCurrentOffset;       // Offset from LoggerInfo to where to store next log entry.
+  UINT32                  DiscardedSize;          // Number of bytes of messages missed
+  UINT32                  LogBufferSize;          // Size of allocated buffer
+  BOOLEAN                 InPermanentRAM;         // Log in permanent RAM
+  BOOLEAN                 AtRuntime;              // After ExitBootServices
+  BOOLEAN                 GoneVirtual;            // After VirtualAddressChange
+  BOOLEAN                 HdwPortInitialized;     // HdwPort initialized
+  BOOLEAN                 HdwPortDisabled;        // HdwPort is Disabled
+  BOOLEAN                 Reserved2[3];           //
+  UINT64                  TimerFrequency;         // Ticks per second for log timing
+  UINT64                  TicksAtTime;            // Ticks when Time Acquired
+  EFI_TIME                Time;                   // Uefi Time Field
+  UINT32                  HwPrintLevel;           // Logging level to be printed at hw port
+  UINT32                  Reserved3;              //
+  EFI_PHYSICAL_ADDRESS    NewLoggerInfoAddress;   // If non-zero, this field holds the address of new logger info that should be used
 } ADVANCED_LOGGER_INFO;
 
 typedef struct {
--- a/AdvLoggerPkg/Include/AdvancedLoggerInternal.h
+++ b/AdvLoggerPkg/Include/AdvancedLoggerInternal.h
@@ -11,8 +11,22 @@
 #ifndef __ADVANCED_LOGGER_INTERNAL_H__
 #define __ADVANCED_LOGGER_INTERNAL_H__
 
-#define ADVANCED_LOGGER_SIGNATURE    SIGNATURE_32('A','L','O','G')
-#define ADVANCED_LOGGER_HW_LVL_VER   5
+#define ADVANCED_LOGGER_SIGNATURE  SIGNATURE_32('A','L','O','G')
+
+//
+// Advanced Logger Hardware Port Level Support Version
+//
+// Writing to hardware port based on message DebugLevel is supported
+// (AdvancedLoggerHdwPortWrite()).
+//
+#define ADVANCED_LOGGER_HW_LVL_VER  5
+
+//
+// Advanced Logger Message structure versions
+//
+// Version 1: Original structure (ADVANCED_LOGGER_MESSAGE_ENTRY)
+// Version 2: New backward incompatible V2 structure (ADVANCED_LOGGER_MESSAGE_ENTRY_V2
+//
 #define ADVANCED_LOGGER_MSG_MAJ_VER  2
 #define ADVANCED_LOGGER_MSG_MIN_VER  0
 
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/BaseArm/AdvancedLoggerLib.c
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/BaseArm/AdvancedLoggerLib.c
@@ -111,6 +111,8 @@ AdvancedLoggerGetLoggerInfo (
   if (((mLoggerInfo) != NULL) && !ValidateInfoBlock ()) {
     mLoggerInfo = NULL;
     DEBUG ((DEBUG_ERROR, "%a: LoggerInfo marked invalid\n", __FUNCTION__));
+  } else if ((mLoggerInfo != NULL) && !FeaturePcdGet (PcdAdvancedLoggerFixedInRAM)) {
+    AdvancedLoggerCheckForNewerLogger (&mLoggerInfo, &mMaxAddress, &mBufferSize);
   }
 
   return mLoggerInfo;
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/BaseArm/AdvancedLoggerLib.c
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/BaseArm/AdvancedLoggerLib.c
@@ -110,7 +110,7 @@ AdvancedLoggerGetLoggerInfo (
 
   if (((mLoggerInfo) != NULL) && !ValidateInfoBlock ()) {
     mLoggerInfo = NULL;
-    DEBUG ((DEBUG_ERROR, "%a: LoggerInfo marked invalid\n", __FUNCTION__));
+    DEBUG ((DEBUG_ERROR, "%a: LoggerInfo marked invalid\n", __func__));
   } else if ((mLoggerInfo != NULL) && !FeaturePcdGet (PcdAdvancedLoggerFixedInRAM)) {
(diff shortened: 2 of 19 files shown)


makubacki:
My understanding is that there are at least two additional changes likely needed.

1. Allow an option for platforms to use a static memory region throughout all of boot. This region would be platform-defined (fixed) and not dynamically allocated. This will split the incoming buffer to DXE into either a "static" and "dynamic" case. DXE will only migrate the buffer if it falls under the "dynamic" case. The case might be able to be simply detected by checked with: `PcdAdvancedLoggerBase != 0`.

2. In the "dynamic" buffer case, where it is migrated to a DXE allocated buffer, the MM Core AdvancedLoggerLib instance will initially start working with the Boot Services Data buffer as described in `gAdvancedLoggerHobGuid`, but we'd need to send the updated shared buffer to the MM Core when it is allocated by the DXE Core instance.

For others that have worked more with adv logger, is that true? Also, any quick sanity checks of the current changes (while those updates are being made) is appreciated.

os-d (commented):
As noted offline, I think this works, but we should only do this if we aren't doing the fixed buffer region

makubacki:
@cfernald, @kuqin12, @os-d, this is now ready for review.

We do plan to move to a fixed runtime region allocated in PEI. This is being checked into release/202502 which is needed by current platforms. The PEI fixed buffer may or may not be something we want to risk merging into 202502 (leaning more toward not). This can simply be adjusted at that time to not perform the buffer migration. That already is controlled with `PcdAdvancedLoggerFixedInRAM`. Perhaps then it would be `!PcdAdvancedLoggerFixedInRAM || gEfiMemoryTypeInformationGuid Resource Descriptor HOB present` or similar logic. The main point is it can relatively co-exist with PEI carved out RT memory

It could also be dropped entirely at that time if all platforms based on future release branches are forced to switch to a fixed PEI RT memory region.

AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
```diff
@@ -357,65 +362,238 @@ ProcessProtocolRegistration (
...
+  //
+  NewLoggerInfo = (ADVANCED_LOGGER_INFO *)AllocateReservedPages (FixedPcdGet32 (PcdAdvancedLoggerPages));
+  if (NewLoggerInfo == NULL) {
+    ASSERT (NewLoggerInfo != NULL);
+    return;
+  }
+
+  //
+  // Unblock the new buffer for MM access
+  // If this is not needed on a platform, tne null instance of MmUnblockMemoryLib can be used.
```
kuqin12:
```suggestion
  // If this is not needed on a platform, the null instance of MmUnblockMemoryLib can be used.
```

AdvLoggerPkg/Include/AdvancedLoggerInternal.h
```diff
@@ -46,25 +46,26 @@
...
+  BOOLEAN                 GoneVirtual;            // After VirtualAddressChange
+  BOOLEAN                 HdwPortInitialized;     // HdwPort initialized
+  BOOLEAN                 HdwPortDisabled;        // HdwPort is Disabled
+  BOOLEAN                 Reserved2[3];           //
+  UINT64                  TimerFrequency;         // Ticks per second for log timing
+  UINT64                  TicksAtTime;            // Ticks when Time Acquired
+  EFI_TIME                Time;                   // Uefi Time Field
+  UINT32                  HwPrintLevel;           // Logging level to be printed at hw port
+  UINT32                  Reserved3;              //
+  EFI_PHYSICAL_ADDRESS    NewLoggerInfoAddress;   // If non-zero, this field holds the address of new logger info that should be used
```
os-d, line 87: When we update this structure, we also need to bump the version number (and then the parsing logic to take the next version and any update required).

AdvLoggerPkg/Library/AdvancedLoggerLib/BaseArm/AdvancedLoggerLib.c
```diff
@@ -110,7 +110,9 @@ AdvancedLoggerGetLoggerInfo (
 
   if (((mLoggerInfo) != NULL) && !ValidateInfoBlock ()) {
     mLoggerInfo = NULL;
-    DEBUG ((DEBUG_ERROR, "%a: LoggerInfo marked invalid\n", __FUNCTION__));
+    DEBUG ((DEBUG_ERROR, "%a: LoggerInfo marked invalid\n", __func__));
+  } else if (mLoggerInfo != NULL) {
+    AdvancedLoggerCheckForNewerLogger (&mLoggerInfo, &mMaxAddress, &mBufferSize);
```
os-d, line 115: This instance I don't think we want to do this, because this instance is expecting the fixed at build PCD that defines the log region, e.g. this is a static case. We share this log region with secure world, so it can't move (without complex memory sharing that may or may not be possible depending on the platform).

In general, if the static option is used, I think we shouldn't relocate the buffer, or even check for relocating.

AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
```diff
@@ -357,65 +362,238 @@ ProcessProtocolRegistration (
...
+  //
+  Status = MmUnblockMemoryRequest (
```
os-d, line 479: This is fine for now, though two downsides:
- One Kun mentioned in discussion, that there is no API to re-block the old region, which is now freed, so SMM has access to some region the OS can use (if this non-runtime memory region will be unblocked, I assume it must so that StMM can log to it? Before end of of DXE? Maybe I am misunderstanding this flow and that isn't a concern).
- With Patina StMM we are currently proposing that memory unblocking in DXE won't be allowed. Now, that doesn't really affect this and that presupposes that we have memory buckets in PEI which obviates the need for the relocation at all, but just mentioning for completeness.

AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
```diff
@@ -357,65 +362,238 @@ ProcessProtocolRegistration (
...
+        );
+      NewLoggerInfo->LogCurrentOffset = NewLoggerInfo->LogBufferOffset + USED_LOG_SIZE (ExistingLoggerInfo);
+    }
+
+    NewLoggerInfo->DiscardedSize = ExistingLoggerInfo->DiscardedSize;
+
+    //
+    // Set the old logger info's NewLoggerInfoAddress to redirect to the new buffer
+    //
+    ExistingLoggerInfo->NewLoggerInfoAddress = PA_FROM_PTR (NewLoggerInfo);
```
os-d, line 523: In this window it is possible that we lose logs from an interrupt, because we copy over, then set the NewLoggerInfoAddress, so we could copy, interrupt, log to old buffer, come back and set NewLoggerInfoAddress and never get the interrupt logs. Perhaps we can:
- Write LogCurrentOffset to NewBufferInfo
- Set NewLoggerInfoAddress (and probably should do that with a compare exchange to ensure something doesn't see a partial write again in the interrupt case, like we do for the LogCurrentOffset)
- Do the copy from LogBufferOffset -> OldInfo->LogCurrentOffset

Probably simpler to disable interrupts while doing this.

AdvLoggerPkg/Include/AdvancedLoggerInternal.h
makubacki: Added a commit that explains what versioning means and bumped the minor version in this update.

AdvLoggerPkg/Library/AdvancedLoggerLib/BaseArm/AdvancedLoggerLib.c
makubacki: I was waiting to hear what we want to do on Arm platforms. Do they set `PcdAdvancedLoggerFixedInRAM` to true?

os-d: Yeah, we can rely on that PCD being set, so this instance can also check that PCD and only check for new logger if it is false (although I think this whole instance would be broken if it is false).

AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
makubacki: StMm launches in PEI. My understanding is that it can and will start writing to the buffer based on its launch schedule irrespective of PEI and DXE, so the original Boot Services Data region is unblocked to allow that. There is no API to reblock it, but I didn't consider that a problem worth adding an API for. It's another gap in the original design of RT DXE buckets with PEI MM IPL launch and communication buffers.

Yes, this is just for the C DXE Core.

AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
makubacki: I raised to `TPL_NOTIFY` during the migration code. This prevents all practical notifies from interrupting during the full migration section and complies with UEFI spec restrictions around TPL for protocol services invocation.

os-d: You might still lose logs from an interrupt, though. Say a timer interrupt (or whatever) fires after the copy, before the new address set, so it will log to the old log, but here we've already done the copy and so miss out on those logs. I think it would be safer to entirely disable interrupts (could do TPL_HIGH, but also messing with TPLs can also bring headaches), so could use the CpuArch protocol (definitely available by end of DXE) to disable/enable interrupts during this time period.

If we just do that for the copy and setting the new logger address, should be fine, but I guess the driver lib is relying on the protocol being produced again, right? In which case that gets a little hairier. We can say that we might miss logs in this window and that's okay, but doesn't feel very "advanced" :)

AdvLoggerPkg/Library/AdvancedLoggerLib/BaseArm/AdvancedLoggerLib.c
makubacki: Added the PCD check before calling `AdvancedLoggerCheckForNewerLogger()` in this instance.

AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
makubacki: @kuqin12, let me know if I missed something or you'd like additional changes.

AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
makubacki: - Timer notifications are registered to `TPL_NOTIFY` or lower.
- Timer events are queued at their trigger time and signaled when the TPL is lowered beyond their registered level. For example, a `TPL_NOTIFY` timer notification will not interrupt code executing at `TPL_NOTIFY`.
- Execution of services from another interrupt source would generally be a violation of calling restrictions if calling into UEFI sources managed by TPL.

- `OnEndOfDxe()` is executed at `TPL_CALLBACK`.
  - So, a timer notification at `TPL_NOTIFY` could interrupt it
- This change surrounds the actual migration section with `TPL_NOTIFY` so normal timer notifications are not an issue.

Going to `TPL_HIGH` requires reducing the guarded section (because of calling restrictions on functions called which try to raise to `TPL_NOTIFY` and cause TPL inversion). I'd like to keep the set of migration code in the TPL area together right now which includes some of those calling restrictions.

The more likely case may be that an asynchronous (hardware SMI) occurs during the copy, but SMIs are not disabled with normal CPU interrupts anyway.

I guess that it's not an issue to proactively disable interrupts entirely, but in that case, I will register `OnEndOfDxe()` at `TPL_NOTIFY` itself and then disable interrupts internally in the function.

makubacki: Thinking about that a bit more, it might make the most sense to unblock the new region, trigger a SW MMI, and then copy in MM which would have isolation against all cases.

makubacki: @os-d & @kuqin12, any concern with that?

os-d: The timer interrupt is a bad example, because you are right that will get dispatched at TPL_NOTIFY (though any logs in the core handling of the timer interrupt and queueing we would miss). For something like the ARM64 hardware interrupt protocol, an interrupt will get generated and the exception handler will directly call the handler. There's an equivalent on the x86 side if different interrupts are registered for.

> Thinking about that a bit more, it might make the most sense to unblock the new region, trigger a SW MMI, and then copy in MM which would have isolation against all cases.

That makes sense to me and as you say gives greater isolation.

AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
kuqin12: No, within the scope of this PR, we are good. But API to reblock a region. :) That sounds fun. If we actually want to do this, we can add it and use the variable driver (as always) as the first testing vehicle.

AdvLoggerPkg/Library/AdvancedLoggerLib/DxeCore/AdvancedLoggerLib.c
makubacki: After prototyping the copy-in-MM approach today, I think it's going to get too complex trying to support supervised and non-supervised MM and I'm going to go back to just copying in DXE. Like some past cases, the fact that the Ring3Broker is a normal `MM_STANDALONE` module type while also serving as the "user mode core" makes hooking in flows unique to it more complicated.

makubacki: Surrounded most of the code at TPL_HIGH_LEVEL and all of the migration code at least at TPL_NOTIFY.

kuqin12 (approved):

os-d (approved):

cfernald (approved):

os-d (approved):

PR #820, opened 2026-01-29
Configure watchdog interrupt for ARM platform

rohansenUSA:
## Description

When using `MpManagement` to suspend all cores to a C-state that powers down the core the non-secure watchdog is being used to wake up a core after a certain amount of time.

On ARM reference platforms the non-secure watchdog interrupt is a level interrupt and active high. and this PR configures the interrupt trigger accordingly. Also, when all cores are powered down interrupt `IRM (Interrupt Routing Mode)` can't be used and the interrupt have to be routed to a specific core using the affinity fields. This PR programs the interrupt to go to the `BSP`.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Tested on custom Microsoft hardware.

Following use case was run:
1. Program watchdog timer.
2. Power down cores.
3. Watchdog interrupt to wake up BSP.

## Integration Instructions

N/A

### DIFF
--- a/UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/AARCH64/SuspendHandling.c
+++ b/UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/AARCH64/SuspendHandling.c
@@ -720,10 +720,10 @@ CpuArchBspSleepPrep (
 
   // Watchdog interrupt is level and active high
   Status = mInterruptProtocol->SetTriggerType (
-                  mInterruptProtocol,
-                  PcdGet32 (PcdGenericWatchdogEl2IntrNum),
-                  EFI_HARDWARE_INTERRUPT2_TRIGGER_LEVEL_HIGH
-                  );
+                                 mInterruptProtocol,
+                                 PcdGet32 (PcdGenericWatchdogEl2IntrNum),
+                                 EFI_HARDWARE_INTERRUPT2_TRIGGER_LEVEL_HIGH
+                                 );
   if (EFI_ERROR (Status)) {
     DEBUG ((DEBUG_ERROR, "%a setting trigger type failed - %r.\n", __FUNCTION__, Status));
     goto Done;
--- a/UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/AARCH64/SuspendHandling.c
+++ b/UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/AARCH64/SuspendHandling.c
@@ -31,6 +31,8 @@
 #include <Protocol/WatchdogTimer.h>
 #include <Guid/ArmMpCoreInfo.h>
 #include <Library/MuArmGicExLib.h>
+#include <Protocol/HardwareInterrupt2.h>
+
 #include "MpManagementInternal.h"
 
 /* Features flags for CPU SUSPEND power state parameter format. Bits [1:1] */
@@ -52,6 +54,10 @@
 #define PSTATE_TYPE_POWERDOWN   0x1
 
 #define AP_TEMP_STACK_SIZE  EFI_PAGE_SIZE
+
+/* GIC specific information */
+#define FIRST_SPI_INTID                  32
+#define GIC_IROUTER_REGISTER_SIZE_BYTES  8
 
 /*
   Architectural metadata structure for ARM context losing resume routines.
@@ -81,6 +87,7 @@ UINT64                            *gApStacksBase      = NULL;
 CONST UINT64                      gApStackSize        = AP_TEMP_STACK_SIZE;
 UINT64                            mWatchDogTimer      = 0;
 EFI_WATCHDOG_TIMER_ARCH_PROTOCOL  *mWatchDogProtocol  = NULL;
+EFI_HARDWARE_INTERRUPT2_PROTOCOL  *mInterruptProtocol = NULL;
 
 /**
   EFI_CPU_INTERRUPT_HANDLER that is called when a processor interrupt occurs.
@@ -667,6 +674,8 @@ CpuArchBspSleepPrep (
   IN UINTN  TimeoutInMicrosecond
   )
 {
+  UINT64      MpId;
+  UINT64      CpuTarget;
   EFI_STATUS  Status;
 
   if (mWatchDogProtocol == NULL) {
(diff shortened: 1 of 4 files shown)


UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/AARCH64/SuspendHandling.c
```diff
@@ -689,13 +698,50 @@ CpuArchBspSleepPrep (
...
+      DEBUG ((DEBUG_ERROR, "%a locating hardware interrupt protocol failed - %r.\n", __FUNCTION__, Status));
+      goto Done;
+    }
+  }
+
+  // Watchdog interrupt is level and active high
+  Status = mInterruptProtocol->SetTriggerType (
+                                 mInterruptProtocol,
+                                 PcdGet32 (PcdGenericWatchdogEl2IntrNum),
+                                 EFI_HARDWARE_INTERRUPT2_TRIGGER_LEVEL_HIGH
```
kuqin12, line 725: Should this be `EFI_HARDWARE_INTERRUPT2_TRIGGER_EDGE_RISING`? According to https://github.com/microsoft/mu_silicon_arm_tiano/blob/bd7850321a762c1eb2f0a3b11bcac0265c863741/ArmPkg/Drivers/GenericWatchdogDxe/GenericWatchdogDxe.c#L388

UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/AARCH64/SuspendHandling.c
```diff
@@ -689,13 +698,50 @@ CpuArchBspSleepPrep (
...
+
+  // All cores will power off so we need affinity routing instead of IRM.
+  // The receiving core for the interrupt will be the bsp.
+  MpId      = ArmReadMpidr ();
+  CpuTarget = MpId &
+              (ARM_CORE_AFF0 | ARM_CORE_AFF1 | ARM_CORE_AFF2 | ARM_CORE_AFF3);
+  MmioWrite64 (
+    PcdGet64 (PcdGicDistributorBase) +
+    ARM_GICD_IROUTER +
+    ((PcdGet32 (PcdGenericWatchdogEl2IntrNum) - FIRST_SPI_INTID) * GIC_IROUTER_REGISTER_SIZE_BYTES),
```
kuqin12, line 740: There are routines that are supposed to perform this during boot: https://github.com/microsoft/mu_silicon_arm_tiano/blob/bd7850321a762c1eb2f0a3b11bcac0265c863741/ArmPkg/Drivers/ArmGicDxe/GicV3/ArmGicV3Dxe.c#L681. But I guess the assumption in the code path is preventing the GIC initialization from programming these registers.

I think this change here is good. But would you think it will be worth changing the condition check there so that GIC can be programmed to this state?

UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/AARCH64/SuspendHandling.c
rohansenUSA: The two latest ARM server reference platforms: Genesis-128c and Voyager-68c both have the AP_NSWDOG_WS0 interrupt as level trigger and polarity high.

What I think we really need is for ARM to make that interrupt completely configurable in the wdog driver.

UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/AARCH64/SuspendHandling.c
rohansenUSA: Yeah, I actually looked into why this wasn't being done. The Initialization of all the interrupts to go to the BSP is only done if `GICD_CTLR.DS == 1`. This bit disables GIC security and allows NS-EL2 to manipulate interrupts for all exception levels and security states which we can't allow.

If `GICD_CTLR.DS == 0` the driver assumes the interrupts have been setup by another FW component probably TFA. The following comment in `ArmGicV3Dxe.c` seems to suggest this:
```
    // If the Disable Security (DS) control bit is set, we are dealing with a
    // GIC that has only one security state. In this case, let's assume we are
    // executing in non-secure state (which is appropriate for DXE modules)
    // and that no other firmware has performed any configuration on the GIC.
    // This means we need to reconfigure all interrupts to non-secure Group 1
    // first.
```
In TFA we don't initialize any interrupts that isn't specifically owned in EL3. I tend to be more in favor of setting up the interrupts that you specifically want to support and nothing more.

In most cases this is not an issue since all interrupts by default have `IRM (Interrupt Routing Mode) == 1` basically meaning interrupt will go to any processor that can handle the specific interrupt group but since we are powering down all the processors we can't use `IRM` but need to program affinity.

FYI. I believe the choice with not waking up a processor to handle an interrupt where `IRM == 1` is IMPLEMENTATION DEFINED.

Shrirammc55 (approved):

UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/AARCH64/SuspendHandling.c
kuqin12: Yeah, I did not see any ARM document that defines the trigger type. Can you please confirm whether this is imp-def? If so, I wonder if the watchdog timer driver in general is broken.

On the same topic, would you think it will be better to program the edge in the `GenericWatchdogDxe` driver? Then we would not need to do it here?

UefiTestingPkg/FunctionalSystemTests/MpManagement/Driver/AARCH64/SuspendHandling.c
kuqin12: Thanks for the explanation. It makes sense for the platform to enable the DS bit and not assume the routing from TFA. But the fact that we are running on the assumption of some other entity (TFA) programmed the GIC, but it actually was not programmed is a bit bothering.

Would you think it might be better to program this when an interrupt ID receives an interrupt handler registration?

kuqin12:
@rohansenUSA I do not mind getting this change in if this is blocking. But I am wondering if there is something else GIC related going wrong in ArmPkg.

os-d (approved):

kuqin12 (approved):

PR #821, opened 2026-01-29
Added PcdAlwaysPrintAssertMsgToSerialPort

PaddyDengAmi:
## Description

With the PCD enabled, assert lib will print the ASSERT message to serial port in addition of AdvancedLogger

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on platform and disable normal debug print via AdvancedLogger, the ASSERT message still shows up in the serial port console.

## Integration Instructions

N/A

### DIFF
--- a/AdvLoggerPkg/AdvLoggerPkg.dec
+++ b/AdvLoggerPkg/AdvLoggerPkg.dec
@@ -131,6 +131,12 @@
   #
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerHdwPortDebugPrintErrorLevel|0xFFFFFFFF|UINT32|0x00010180
 
+  ## Always Print Assert Message to Serial Port
+  #
+  # Enable AssertLib to print ASSERT message to serial port in addition of AdvLogger
+  #
+  gAdvLoggerPkgTokenSpaceGuid.PcdAlwaysPrintAssertMsgToSerialPort|FALSE|BOOLEAN|0x00010199
+
 
 [UserExtensions.TianoCore."ExtraFiles"]
   AdvLoggerPkgExtra.uni
--- a/AdvLoggerPkg/Library/AssertLib/AssertLib.c
+++ b/AdvLoggerPkg/Library/AssertLib/AssertLib.c
@@ -18,6 +18,7 @@
 #include <Library/DebugLib.h>
 #include <Library/PcdLib.h>
 #include <Library/PrintLib.h>
+#include <Library/SerialPortLib.h>
 
 //
 // Define the maximum debug and assert message length that this library supports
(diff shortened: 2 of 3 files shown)


os-d:
I wouldn't add the PCD and just always print the assert message to serial. If we are hitting assertlib, the system is dead, may as well just always print it.

makubacki (changes requested):
This needs clarification as to what you are trying to do and why existing mechanisms are not sufficient. Then we can determine how to handle this.

Another PCD at the `DebugLib` level in the abstraction stack to bypass calling `AdvancedLoggerWrite()` which calls `AdvancedLoggerHdwPortWrite()` which calls `SerialPortWrite()` (in the non-null instance of `AdvancedLoggerHdwPortLib`) is strange.

kuqin12:
> This needs clarification as to what you are trying to do and why existing mechanisms are not sufficient. Then we can determine how to handle this.
> 
> Another PCD at the `DebugLib` level in the abstraction stack to bypass calling `AdvancedLoggerWrite()` which calls `AdvancedLoggerHdwPortWrite()` which calls `SerialPortWrite()` (in the non-null instance of `AdvancedLoggerHdwPortLib`) is strange.

Agreed. This should have handled in the `AdvancedLoggerWrite` call above. You might want to figure out why the prints are not going through the `AdvancedLoggerHdwPortWrite` instead. Maybe `HdwPortDisabled` in the header was set to FALSE, or the error level was set improperly.

MarcChen46:
Thanks @makubacki @kuqin12 and @os-d for your comments.

**Background**:
In our custom `AdvancedLoggerHdwPortWrite` implementation, we have a runtime debug mode check that controls whether debug messages are printed to the serial console. By default, this is disabled.

However, we want to **always print ASSERT-related messages to the serial console**, regardless of the debug mode setting. This ensures that during normal validation or any formal use case, any `ASSERT` will be immediately visible to the validation team, allowing them to investigate and debug issues in a timely manner.

Since our BIOS is built as a release build, the system does not halt on `ASSERT`. This makes it even more critical that `ASSERT` messages are always output to the serial console so they are not silently missed.

**Alternative approach considered**:
We could implement the `ASSERT` detection logic directly in `AdvancedLoggerHdwPortWrite` by checking if the message matches the `ASSERT` message pattern (e.g., checking if the buffer starts with "ASSERT"). However, this approach has a maintenance concern: if the `ASSERT` message format changes in the future(it actually not changed for 10 years, not sure if it is a real concern), our pattern-matching logic would also need to be updated to remain in sync.

Please provide your comment and suggestion, thanks a lot.

makubacki:
> Please provide your comment and suggestion, thanks a lot.

The AssertLib.c file in AdvLoggerPkg is not that complex. Can you just create your own instance? Then you can have whatever logic you want in there.

MarcChen46:
Thanks for the comment, we will implement our own library.

@PaddyDengAmi, please close abandon this PR, thanks.

PaddyDengAmi:
Dropping this PR per discussion.

PR #823, opened 2026-02-03
Refactor HID Keyboard Reset support

joschock:
## Description
 
This PR changes how CTRL-ALT-DEL resets are handled by the HID stack. Previously, they were handled by the lower layer key processing code; but this caused them to be handled at an implementation-defined TPL based on what TPL the input reports are generated at by the HID I/O layer. 

This change moves the CTRL-ALT-DEL to use the existing SimpleTextInEx key registration infrastructure to handle the reset logic, effectively moving the reset handling up to a higher level of abstraction.

This also simplifies disabling this capability behind a feature flag should it be desirable in the future to separate reset logic from the HID stack entirely.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified CTRL-ALT-DEL resets system from UEFI shell and that reset occurs at TPL_CALLBACK.

## Integration Instructions

N/A

### DIFF
--- a/HidPkg/UefiHidDxeV2/src/keyboard.rs
+++ b/HidPkg/UefiHidDxeV2/src/keyboard.rs
@@ -579,7 +579,7 @@ extern "efiapi" fn reset_notification_function(key_data: *mut protocols::simple_
 
     //DEL scan code received with shift state indicating CTRL-ALT also pressed.
     debugln!(DEBUG_WARN, "Ctrl-Alt-Del pressed, resetting system.");
-    if let Some(runtime_services) = unsafe { RUNTIME_SERVICES.load(Ordering::SeqCst).as_mut() } {
+    if let Some(runtime_services) = unsafe { RUNTIME_SERVICES.load(Ordering::SeqCst).as_ref() } {
         (runtime_services.reset_system)(efi::RESET_COLD, efi::Status::SUCCESS, 0, core::ptr::null_mut());
     }
     panic!("Reset failed.");
--- a/HidPkg/UefiHidDxeV2/src/keyboard.rs
+++ b/HidPkg/UefiHidDxeV2/src/keyboard.rs
@@ -19,9 +19,18 @@ use alloc::{
     vec,
     vec::Vec,
 };
+use core::sync::atomic::Ordering;
 use core::{ffi::c_void, ptr};
 
-use r_efi::{efi, hii, protocols};
+use r_efi::{
+    efi, hii,
+    protocols::{
+        self,
+        simple_text_input_ex::{
+            LEFT_ALT_PRESSED, LEFT_CONTROL_PRESSED, RIGHT_ALT_PRESSED, RIGHT_CONTROL_PRESSED, SHIFT_STATE_VALID,
+        },
+    },
+};
 
 use hidparser::{
     ArrayField, ReportDescriptor, ReportField, VariableField,
@@ -31,6 +40,7 @@ use mu_rust_helpers::function;
 use rust_advanced_logger_dxe::{DEBUG_ERROR, DEBUG_VERBOSE, DEBUG_WARN, debugln};
 
 use crate::{
+    RUNTIME_SERVICES,
     boot_services::UefiBootServices,
     hid_io::{HidIo, HidReportReceiver},
     keyboard::key_queue::OrdKeyData,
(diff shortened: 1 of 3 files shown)


HidPkg/UefiHidDxeV2/src/keyboard.rs
```diff
@@ -542,13 +552,59 @@ impl KeyboardHidHandler {
...
+    }
+
+    if key_data.key_state.key_shift_state & (LEFT_ALT_PRESSED | RIGHT_ALT_PRESSED) == 0 {
+        return efi::Status::SUCCESS; // no ALT pressed.
+    }
+
+    //DEL scan code received with shift state indicating CTRL-ALT also pressed.
+    debugln!(DEBUG_WARN, "Ctrl-Alt-Del pressed, resetting system.");
+    if let Some(runtime_services) = unsafe { RUNTIME_SERVICES.load(Ordering::SeqCst).as_mut() } {
+        (runtime_services.reset_system)(efi::RESET_COLD, efi::Status::SUCCESS, 0, core::ptr::null_mut());
```
antklein, line 583: NIT: was this intentionally changed from a `RESET_WARM`?

HidPkg/UefiHidDxeV2/src/keyboard.rs
```diff
@@ -542,13 +552,59 @@ impl KeyboardHidHandler {
...
+
+    if key_data.key_state.key_shift_state & (LEFT_CONTROL_PRESSED | RIGHT_CONTROL_PRESSED) == 0 {
+        return efi::Status::SUCCESS; // no CTRL pressed.
+    }
+
+    if key_data.key_state.key_shift_state & (LEFT_ALT_PRESSED | RIGHT_ALT_PRESSED) == 0 {
+        return efi::Status::SUCCESS; // no ALT pressed.
+    }
+
+    //DEL scan code received with shift state indicating CTRL-ALT also pressed.
```
antklein, line 580: NIT: spacing

apop5:
I do have a concern.

Consider a more complex system that has multiple keyboards, and therefore multiple SimpleTextInEx protocols

Consplitter (SimpleTextInEx , with no device path)
   USB KB1 (SimpleTextInEx , valid device path)
   USB KB2 (SimpleTextInEx, valid device path)
   PS2 KB1
   
With this change, USB KB1, USB KB2 would each have a keypress handler installed to handle CTRL+ALT+DEL, but Consplitter and PS2 KB1 would not have anything registered.

Would it be better to move the registration into the console splitter? That way there could be a single place that would handle resets?

joschock:
> I do have a concern.
> 
> Consider a more complex system that has multiple keyboards, and therefore multiple SimpleTextInEx protocols
> 
> Consplitter (SimpleTextInEx , with no device path) USB KB1 (SimpleTextInEx , valid device path) USB KB2 (SimpleTextInEx, valid device path) PS2 KB1
> 
> With this change, USB KB1, USB KB2 would each have a keypress handler installed to handle CTRL+ALT+DEL, but Consplitter and PS2 KB1 would not have anything registered.

Your understanding is correct - there would be a reset notify linked to each SimpleTextInEx produced by the UefiHidV2. This is consistent (at least semantically) with what would happen today if you were using standard EDK2 drivers instead of this stack, though. The difference is where the hook happens - with existing implementation (in this driver, as well as in the other EDK2 keyboard drivers) the hook happens at the device layer during keystroke processing. This PR changes it to happen at the key registration layer - which is a better point in my view, and more consistent with what I think the right long-term direction is (see second part of my comment below). 

PS2 KB1 would have it's own reset trigger here: https://github.com/microsoft/mu_basecore/blob/afb0c811c17d4943fde9816e0f2ec9eff9ccd1c0/MdeModulePkg/Bus/Isa/Ps2KeyboardDxe/Ps2KbdCtrller.c#L1352.

>Would it be better to move the registration into the console splitter? That way there could be a single place that would handle resets?

It would be much better, IMHO, but would require a fairly coordinated and broad change across EDK2 since all the existing keyboard drivers implement this at the device layer. This PR preps for that by removing the reset logic to a key registration and out of the bowels of the keystroke handler. Eventually I envision this behind a cargo feature gate `legacy_reset_behavior` if something like a registration at the consplitter level materializes in EDK2, but I didn't see the need to implement that in this PR yet; happy to do so if there is value there. 

See also comments here: https://github.com/tianocore/edk2/pull/12038

HidPkg/UefiHidDxeV2/src/keyboard.rs
```diff
@@ -542,13 +552,59 @@ impl KeyboardHidHandler {
...
+        return efi::Status::INVALID_PARAMETER;
+    }
+
+    // SAFETY: null-checked above, using read_unaligned to avoid any alignment issues.
+    let key_data = unsafe { key_data.read_unaligned() };
+    if key_data.key.scan_code != key_queue::SCAN_DELETE {
+        return efi::Status::SUCCESS; // No delete
+    }
+
+    if key_data.key_state.key_shift_state & SHIFT_STATE_VALID == 0 {
```
apop5: Nit: I think the three if statements could all be combined into one of statement.

If left alt or right alt are pressed and if left control or right control and shift state valid.

This would end up capturing all combinations (left alt, right alt, left control + delete) but I think that is okay.

apop5 (approved):

apop5 (approved):

HidPkg/UefiHidDxeV2/src/keyboard.rs
```diff
@@ -542,13 +552,56 @@ impl KeyboardHidHandler {
...
+        || key_data.key_state.key_shift_state & (LEFT_CONTROL_PRESSED | RIGHT_CONTROL_PRESSED) == 0
+        || key_data.key_state.key_shift_state & (LEFT_ALT_PRESSED | RIGHT_ALT_PRESSED) == 0
+    {
+        // DEL pressed without valid CTRL-ALT qualification, so do not reset.
+        return efi::Status::SUCCESS;
+    }
+
+    //DEL scan code received with shift state indicating CTRL-ALT also pressed.
+    debugln!(DEBUG_WARN, "Ctrl-Alt-Del pressed, resetting system.");
+    if let Some(runtime_services) = unsafe { RUNTIME_SERVICES.load(Ordering::SeqCst).as_mut() } {
```
makubacki: nit: Does an immutable ref work?

```suggestion
    if let Some(runtime_services) = unsafe { RUNTIME_SERVICES.load(Ordering::SeqCst).as_ref() } {
```

makubacki (approved):

HidPkg/UefiHidDxeV2/src/keyboard.rs
joschock: sure - that works. updated in latest.

apop5 (approved):

HidPkg/UefiHidDxeV2/src/keyboard.rs
joschock: Yes; there really isn't any spec on this, and reset warm has special meaning on a number of architectures around debugging that is undesirable.  So I changed it to reset-cold.

HidPkg/UefiHidDxeV2/src/keyboard.rs
joschock: didn't see this till after the merge, so won't fix.

PR #824, opened 2026-02-06
AdvLoggerPkg: DecodeUefiLog handle frequency errors.

apop5:
## Description

When calling DecodeUefiLog, if the system was not able to provide a frequency in the log, it will be reported as 0. With no verification check of a zero frequency, DecodeUefiLog would attempt a divide by zero when converting time to nano seconds.

Add a check that prevents divide by zero. This will report times of 0 for all timestamps in the decoded log.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
On a physical platform where no timer frequency was reported, encountered a python exception.
After change, log was correctly retrieved and displays a timestamp of 0.

## Integration Instructions
No integration necessary.


kuqin12 (approved):

makubacki (approved):

makubacki:
@apop5, I went ahead and rebased with autocomplete so I can rebase my PR on the Python script change.

PR #825, opened 2026-02-06
[REBASE & FF] Fix Memory Tests to Run on Patina

os-d:
## Description

See individual commit descriptions for full details. This changeset updates the paging audit simple test, full audit, and memory protections test app to run on Patina.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Tested on physical Intel, Q35, and SBSA.

## Integration Instructions

Ensure the platform is producing the memory protection HOB with:

```c
DxeSettings.HeapGuardPolicy.Fields.UefiPageGuard = 0;
DxeSettings.HeapGuardPolicy.Fields.UefiPoolGuard = 0;
```
Until Patina enables those features, otherwise those tests will execute and will fail due to no guard pages enabled.

This is a breaking change because it depends on mu_basecore version >= v2025020003.0.2 as some header definitions are depended on in that release.

### DIFF
--- a/UefiTestingPkg/AuditTests/PagingAudit/UEFI/X64/PagingAuditProcessor.c
+++ b/UefiTestingPkg/AuditTests/PagingAudit/UEFI/X64/PagingAuditProcessor.c
@@ -577,6 +577,9 @@ GetFlatPageTableData (
   UINTN                           NumPage2MNotPresent = 0;
   UINTN                           NumPage1GNotPresent = 0;
   UINT64                          Address;
+  BOOLEAN                         SelfMapped;
+
+  SelfMapped = FALSE;
 
   //
   // First, fail fast if some of the parameters don't look right.
@@ -599,12 +602,27 @@ GetFlatPageTableData (
   //
   Pml4 = (PAGE_MAP_AND_DIRECTORY_POINTER *)AsmReadCr3 ();
 
+  //
+  // Check for the self-map entry as the last entry
+  //
+  if ((Pml4[0x1FF].Bits.PageTableBaseAddress << 12) == (UINTN)Pml4) {
+    SelfMapped = TRUE;
+    Pml4       = (PAGE_MAP_AND_DIRECTORY_POINTER *)0xFFFFFFFFFFFFF000;
+  }
+
   for (Index4 = 0x0; Index4 < 0x200; Index4++) {
-    if (!Pml4[Index4].Bits.Present) {
+    if (!Pml4[Index4].Bits.Present || (((Index4 == 0x1FF) || (Index4 == 0x1FE)) && SelfMapped)) {
       continue;
     }
 
-    Pte1G = (PAGE_TABLE_1G_ENTRY *)(UINTN)(Pml4[Index4].Bits.PageTableBaseAddress << 12);
+    if (SelfMapped) {
+      Pte1G = (PAGE_TABLE_1G_ENTRY *)(UINTN)((0xFFFFFFFFFFE00000 + SIZE_4KB * Index4));
+    } else {
+      //
+      // No self-map entry, so just use the physical address.
+      //
+      Pte1G = (PAGE_TABLE_1G_ENTRY *)(UINTN)(Pml4[Index4].Bits.PageTableBaseAddress << 12);
+    }
 
     for (Index3 = 0x0; Index3 < 0x200; Index3++ ) {
       if (!Pte1G[Index3].Bits.Present) {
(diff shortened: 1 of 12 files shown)


berlin-with0ut-return (approved):

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/DxeMemoryProtectionTestApp.c
```diff
@@ -1540,79 +1545,211 @@ ImageProtection (
...
+
+      RemoveEntryList (&ImageRangeDescriptor->Link);
+      FreePool (ImageRangeDescriptor);
+    }
+
+    FreePool (ImageRangeDescriptorHead);
+  } else {
+    // if the Mu protocol isn't produced, walk the debug image info table
+    // and verify section attributes directly
+    EFI_DEBUG_IMAGE_INFO_TABLE_HEADER    *DebugTableHeader;
```
kuqin12: Really?

UefiTestingPkg/AuditTests/PagingAudit/UEFI/X64/PagingAuditProcessor.c
```diff
@@ -599,12 +602,27 @@ GetFlatPageTableData (
   //
   Pml4 = (PAGE_MAP_AND_DIRECTORY_POINTER *)AsmReadCr3 ();
 
+  //
+  // Check for the self-map entry as the last entry
```
kuqin12, line 606: Does it work with a semi-self-map table? i.e. not all entries are self-mapped?

UefiTestingPkg/Library/FlatPageTableLib/X64/FlatPageTableX64.c
```diff
@@ -16,6 +17,555 @@
 #define X64_RW_BIT       BIT1
 #define X64_NX_BIT       BIT63
 
+#define IA32_PE_BASE_ADDRESS_MASK_40  0xFFFFFFFFFF000ull
```
apop5: I understand removing CpuPageTableLib from library classes, but these defines are all available in CpuPageTable.h, which can be consumed without PageTableLib library class. 

Do we really want to redefine a second copy of them here?

apop5 (approved):

makubacki (approved):

UefiTestingPkg/AuditTests/PagingAudit/UEFI/X64/PagingAuditProcessor.c
os-d: There is no such thing as a semi self-mapped table. You either have a recursive entry in the top level, which creates the self map VA range, or you don't.

What this doesn't do (which I did in UefiExt, e.g., to make it more robust) is check for a self map entry in any spot except the last one. That is a nice optimization because the math is all easier.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/DxeMemoryProtectionTestApp.c
os-d: This is an easy way to get all the loaded image protocols in one shot. It can be changed to search for all protocol instances, but I thought this was more performant and since it is a test, relying on the debug table is just fine, in my opinion, since it caches the loaded image protocols.

UefiTestingPkg/Library/FlatPageTableLib/X64/FlatPageTableX64.c
os-d: Yeah, that was me being lazy and not changing all the function names. I should change the function names :)

os-d: Actually, nvm, I already had the header here, the names are not an issue. Not sure why the redefines are not a problem. But I will remove them.

UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/DxeMemoryProtectionTestApp.c
os-d: Chatted offline, comment here was on defining variables mid function. Fixed to be at the top.

kuqin12 (approved):

UefiTestingPkg/Library/FlatPageTableLib/X64/FlatPageTableX64.c
os-d: Oh, ugh. These definitions are in CpuPageTable.h not CpuPageTableLib.h. Great...so that's why the redefinition didn't occur. Adding that.

os-d: Which I'm not, because it is an internal header: https://github.com/microsoft/mu_basecore/blob/release/202502/UefiCpuPkg/Library/CpuPageTableLib/CpuPageTable.h

These definitions shouldn't be used by any code, really (but we are special as a test), so I think the duplication is appropriate.

kuqin12 (approved):

PR #828, opened 2026-02-10
AdvLoggerPkg: Initialize structures in LineParserTestApp

makubacki:
## Description

Logger info structures are not zeroed and initialized right now, causing garbage data to be read in some fields and impact test results.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Run LineParserTestApp on QEMU Q35

## Integration Instructions

- N/A

### DIFF
--- a/AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
+++ b/AdvLoggerPkg/UnitTests/LineParser/LineParserTestApp.c
@@ -408,7 +408,9 @@ InitializeInMemoryLog (
     UT_ASSERT_TRUE (FALSE);
   }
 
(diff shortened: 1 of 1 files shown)


os-d (approved):

kuqin12 (approved):
The change looks good. But I still would like to get rid of the test and its underlying library at some point..

makubacki:
> The change looks good. But I still would like to get rid of the test and its underlying library at some point..

Sounds good. This is just keeping the wheels turning for now.

PR #849, opened 2026-03-06
UefiTestingPkg/MemoryAttributeProtocolFuncTestApp

apop5:
## Description
Reduce the amount of memory allocated. Originally test was allocating 4096 pages of memory (16 MB) and changing attributes only the first page.

Remove the reallocation test. With the upstream changes to leak memory if they do not have RW set when Free is called,
(https://github.com/tianocore/edk2/commit/2d69507a4dde02f1abf20c7eb3a43d1d3ef6b98f) this test is failing because it left pages RO.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
QemuQ35 202511 failing test before, passing after. 

## Integration Instructions
No Integration necessary.

### DIFF
--- a/UefiTestingPkg/FunctionalSystemTests/MemoryAttributeProtocolFuncTestApp/MemoryAttributeProtocolFuncTestApp.c
+++ b/UefiTestingPkg/FunctionalSystemTests/MemoryAttributeProtocolFuncTestApp/MemoryAttributeProtocolFuncTestApp.c
@@ -74,7 +74,7 @@ FreePagesWithProtectionAttributesTestCase (
   Length      = EFI_PAGE_SIZE;
 
   // Allocate any pages
-  Status = gBS->AllocatePages (AllocateAnyPages, EfiLoaderCode, Length, &BaseAddress);
+  Status = gBS->AllocatePages (AllocateAnyPages, EfiLoaderCode, EFI_SIZE_TO_PAGES (Length), &BaseAddress);
   UT_ASSERT_NOT_EFI_ERROR (Status);
   UT_ASSERT_NOT_NULL ((VOID *)((UINTN)BaseAddress));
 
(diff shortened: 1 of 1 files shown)


UefiTestingPkg/FunctionalSystemTests/MemoryAttributeProtocolFuncTestApp/MemoryAttributeProtocolFuncTestApp.c
```diff
@@ -170,15 +170,15 @@ AllocateFreeAllocateAtAddressTestCase (
 
   Attributes  = 0;
   BaseAddress = 0;
-  Length      = EFI_PAGE_SIZE;
+  Length      = 1;
 
   // Allocate any pages
   Status = gBS->AllocatePages (AllocateAnyPages, EfiLoaderCode, Length, &BaseAddress);
```
kuqin12: Would it be better to make this converted to number of pages and not change other places?

kuqin12 (approved):

UefiTestingPkg/FunctionalSystemTests/MemoryAttributeProtocolFuncTestApp/MemoryAttributeProtocolFuncTestApp.c
os-d: Or at least rename it NumPages.

os-d (approved):

UefiTestingPkg/FunctionalSystemTests/MemoryAttributeProtocolFuncTestApp/MemoryAttributeProtocolFuncTestApp.c
apop5: Done, switch everything back and use EFI_SIZE_TO_PAGES for the allocate/free

kuqin12 (approved):

PR #850, opened 2026-03-09
UefiTestingPkg\MemoryAttributeProtocolFuncTestApp restore attributes.

apop5:
## Description
When MemoryAttributeProtocolFuncTestApp executes, it leaves some regions of memory as RWX.

The ordering of test applications in mu_tiano_platforms is not fixed. Different environments can result in a different run order.

With the memory regions left as RWX, if DxePagingAudit runs after this application it will report a region was RWX.

After the individual tests are run, use memory attributes protocol to restore the attributes to their original value upon allocation.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Local Build and Test on Q35/Sbsa

## Integration Instructions
No integration necessary.

### DIFF
--- a/UefiTestingPkg/FunctionalSystemTests/MemoryAttributeProtocolFuncTestApp/MemoryAttributeProtocolFuncTestApp.c
+++ b/UefiTestingPkg/FunctionalSystemTests/MemoryAttributeProtocolFuncTestApp/MemoryAttributeProtocolFuncTestApp.c
@@ -64,6 +64,7 @@ FreePagesWithProtectionAttributesTestCase (
   EFI_PHYSICAL_ADDRESS           BaseAddress;
   UINTN                          Length;
   UINT64                         Attributes;
+  UINT64                         OriginalAttributes;
 
   Status = gBS->LocateProtocol (&gEfiMemoryAttributeProtocolGuid, NULL, (VOID **)&MemoryAttribute);
   UT_ASSERT_NOT_EFI_ERROR (Status);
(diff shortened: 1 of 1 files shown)


UefiTestingPkg/FunctionalSystemTests/MemoryAttributeProtocolFuncTestApp/MemoryAttributeProtocolFuncTestApp.c
```diff
@@ -87,6 +92,10 @@ FreePagesWithProtectionAttributesTestCase (
 
   DEBUG ((DEBUG_INFO, "%a - Attributes for memory at base: 0x%llx 0x%llx\n", __FUNCTION__, BaseAddress, Attributes));
 
+  // Restore the orignal attributes
+  Status = MemoryAttribute->SetMemoryAttributes (MemoryAttribute, BaseAddress, Length, OriginalAttributes);
+  UT_ASSERT_NOT_EFI_ERROR (Status);
```
os-d, line 97: Is this not susceptible to the same issue it had when you changed it before? i.e. the memory getting leaked and so we can't change attributes on it?

os-d: Oh, that problem was the free failed, not the set?

os-d (approved):

UefiTestingPkg/FunctionalSystemTests/MemoryAttributeProtocolFuncTestApp/MemoryAttributeProtocolFuncTestApp.c
apop5: Correct.

The previous implementation would change attributes, free, and then attempt to allocate an exact address. But because the attributes were set to nx/ro, the memory would leak and the allocate exact address would fail. As discussed, attempting to allocate an exact address after messing with attributes was testing something internal about the implementation of the core, so the secondary allocate exact address was removed. 

This is to address that when "freed" (leaked) that the attributes were RWX in the PTE2MB due to ->ClearAttributes call. 

The easiest solution was to restore all memory attributes back to the attributes the allocation provided. I view this as cleaning up the test cases correctly.

Javagedes (approved):

PR #851, opened 2026-03-10
Add CompilerIntrinsicsLib to MsUiThemePpi.inf.

jaykrell:
Error buiding:
```
INFO - /usr/lib/aarch64-linux-gnu/bin/ld: /tmp/ccDHolw2.ltrans0.ltrans.o:
 in function `MsUiThemePpiEntry.isra.0':
INFO - MU/MsGraphicsPkg/MsUiTheme/Pei/MsUiThemePpi.c:112:
(.text.MsUiThemePpiEntry.isra.0+0x1dc4):
 undefined reference to `memcpy'
INFO - collect2: error: ld returned 1 exit status
```
```
GuidHob->Name = gMsUiThemeHobGuid
```

Assigning a GUID, compiler generates call to `memcpy `, which is what CompilerIntrinsicsLib is for.

- [ ] Impacts functionality?
No.
- [ ] Impacts security?
No.
- [ ] Breaking change?
It can break builds.
- [ ] Includes tests?
Not necessary.
- [ ] Includes documentation?
Not necessary.

## How This Was Tested

Build.

## Integration Instructions

None required.

### DIFF
--- a/MsGraphicsPkg/MsGraphicsPkg.dsc
+++ b/MsGraphicsPkg/MsGraphicsPkg.dsc
@@ -26,6 +26,7 @@
 [LibraryClasses.common]
   BaseLib|MdePkg/Library/BaseLib/BaseLib.inf
   BaseMemoryLib|MdePkg/Library/BaseMemoryLib/BaseMemoryLib.inf
+  CompilerIntrinsicsLib|MdePkg/Library/CompilerIntrinsicsLib/CompilerIntrinsicsLib.inf
 
 [LibraryClasses.common]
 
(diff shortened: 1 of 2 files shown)


os-d:
This is intended to be resolved by either including MdeLibs.dsc.inc or directly adding `NULL|MdePkg/Library/CompilerIntrinsicsLib/CompilerIntrinsicsLib.inf` to the platform DSC. It should not be added to every module.

jaykrell:
Let's see if we can fix all the breaks instead?
I have a list of 3-4. I think 2 will not be controversial (struct assignment in Mu).
2 might be controversial (passing va_list by value in PrintLib -- I think it should be by address, and receiver CopyMem it. Probably wrapped in a struct, so that if it is an array, it can be portably address-taken),
Here is one:
https://github.com/microsoft/mu_plus/pull/852

PR #863, opened 2026-04-01
AdvancedFileLogger: Filter log file output by PcdDebugPrintErrorLevel

liqiqiii:
## Description

AdvancedFileLogger: Filter log file output by PcdDebugPrintErrorLevel

Rust UEFI drivers (e.g., UefiHidDxeV2) use the AdvancedLogger protocol directly via the [debugln!] macro, bypassing the C DebugLib layer. Unlike C drivers, where DebugPrint() checks PcdDebugPrintErrorLevel via BaseDebugPrintErrorLevelLib before any message enters the AdvancedLogger pipeline, Rust drivers write to the protocol with no level filtering. This causes DEBUG_VERBOSE messages (e.g., [VERB] usage 0x42 inserted) to appear in the file logger output, even though they are correctly filtered from serial output by PcdAdvancedLoggerHdwPortDebugPrintErrorLevel.

## How This Was Tested

Tested with AdvancedFileLogger, no more verbose level log from rust driver is seen, matches the behavior of c drivers.

### DIFF
--- a/AdvLoggerPkg/AdvancedFileLogger/AdvancedFileLogger.inf
+++ b/AdvLoggerPkg/AdvancedFileLogger/AdvancedFileLogger.inf
@@ -63,6 +63,7 @@
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerPages            ## CONSUMES
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedFileLoggerForceEnable  ## CONSUMES
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedFileLoggerFlush        ## CONSUMES
+  gEfiMdePkgTokenSpaceGuid.PcdDebugPrintErrorLevel              ## CONSUMES
 
 [Depex]
   TRUE
--- a/AdvLoggerPkg/AdvancedFileLogger/FileAccess.c
+++ b/AdvLoggerPkg/AdvancedFileLogger/FileAccess.c
@@ -531,6 +531,17 @@ WriteALogFile (
   Status   = AdvancedLoggerAccessLibGetNextFormattedLine (&LogDevice->AccessEntry);
 
   while (Status == EFI_SUCCESS) {
+    //
+    // Filter messages by PcdDebugPrintErrorLevel before writing to the log file.
+    // This ensures the file logger respects the same debug level mask as DebugPrint(),
+    // filtering out messages (e.g. DEBUG_VERBOSE from Rust drivers) that bypass the
+    // DebugLib level check.
+    //
+    if ((LogDevice->AccessEntry.DebugLevel & PcdGet32 (PcdDebugPrintErrorLevel)) == 0) {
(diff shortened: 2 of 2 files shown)


antklein:
Shouldn't PcdAdvancedLoggerHdwPortDebugPrintErrorLevel properly filter both serial and file logging by default?

https://github.com/microsoft/mu_plus/blob/31f2b85ddb606cf6e4b4019ec2db1349d97b162b/AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c#L280-L286

liqiqiii:
> Shouldn't PcdAdvancedLoggerHdwPortDebugPrintErrorLevel properly filter both serial and file logging by default?
> 
> https://github.com/microsoft/mu_plus/blob/31f2b85ddb606cf6e4b4019ec2db1349d97b162b/AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c#L280-L286

no, it will only be applied when the log is exported to the serial ports. It will not affect file logger.

antklein:
> > Shouldn't PcdAdvancedLoggerHdwPortDebugPrintErrorLevel properly filter both serial and file logging by default?
> > https://github.com/microsoft/mu_plus/blob/31f2b85ddb606cf6e4b4019ec2db1349d97b162b/AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c#L280-L286
> 
> no, it will only be applied when the log is exported to the serial ports. It will not affect file logger.

I see, I understand your point now. My only concern is mixing medicine here. But Core team can comment on that. :)

kuqin12 (approved):

AdvLoggerPkg/AdvancedFileLogger/AdvancedFileLogger.inf
```diff
@@ -63,6 +63,7 @@
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerPages            ## CONSUMES
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedFileLoggerForceEnable  ## CONSUMES
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedFileLoggerFlush        ## CONSUMES
+  gEfiMdePkgTokenSpaceGuid.PcdDebugPrintErrorLevel              ## CONSUMES
```
antklein, line 66: Would a better future design better align Rust drivers to the current C design with better log filter?
Or maybe AdvLoggerPkg in general should be improved to have better "single-source" log filtering?

os-d: To Antaeus's point, future investment is in Patina components, not standalone Rust drivers. Even the component referenced here has a PR up in Patina to add it as a component. I think it would be better to change the existing Rust standalone components to align to the current design or to ignore this problem and transition to the Patina components.

liqiqiii:
Per offline discussion with Oliver, close this PR as we are moving towards patina components for rust drivers.

PR #864, opened 2026-04-01
[REBASE & FF] Copy FltUsedLib to MsCorePkg

os-d:
## Description

FltUsedLib is a library that has a single line, adding a reference to `_fltused`, which is needed by MSVC and clang when building floating point code.

FltUsedLib is only used in mu_plus, so in order to reduce deltas from edk2, it is moved to mu_plus. This commit then updates references in mu_plus to this new library.

A PR will be put up to mu_basecore to drop the library after this merges.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Build testing.

## Integration Instructions

Platforms should update to `FltUsedLib|MsCorePkg/Library/FltUsedLib/FltUsedLib.inf` in their DSCs. This change is not a breaking change, but the removal in mu_basecore will be.

### DIFF
--- a/MsCorePkg/Library/FltUsedLib/FltUsedLib.c
+++ b/MsCorePkg/Library/FltUsedLib/FltUsedLib.c
@@ -1,11 +1,13 @@
 /** @file
-  Lib to include if using floats
+  This library provides the _fltused symbol which is required to by MSVC and clang
+  when floating point operations are used. This library should be linked in for
+  any module that uses floating point operations.
 
-Copyright (C) Microsoft Corporation.
-SPDX-License-Identifier: BSD-2-Clause-Patent
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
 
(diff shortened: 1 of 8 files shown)


vineelko (approved):

MsCorePkg/Library/FltUsedLib/FltUsedLib.c
```diff
@@ -0,0 +1,11 @@
+/** @file
+  Lib to include if using floats
```
makubacki: Can you clean this up to be a little more descriptive while here?

makubacki: Same for the inf comment

makubacki (approved):

MsCorePkg/Library/FltUsedLib/FltUsedLib.c
os-d: Updated both.

vineelko (approved):

PR #868, opened 2026-04-08
AdvLoggerPkg: Drop buffer migration

makubacki:
## Description

Effectively reverts commits af0f64c and 9fc8521 preserving structure compatibility and general cleanup.

The advanced logger buffer starts accumulating messages in the first phase advanced logger is active, and the buffer continues to be used throughout remaining phases. Each phase has to account for either allocating or using an existing logger buffer. Due to the persistent nature of the buffer throughout boot, it is allocated as a runtime memory type early in boot to support its preservation and access at OS runtime.

However, S4 resume relies upon a stable memory map across suspend and resume. Since the advanced logger buffer is allocated as a runtime memory type prior to DXE, it is in a buffer outside the DXE memory bins. This leads to an increased likelihood of the buffer being at a different address across suspend and resume affecting hibernate resume. Advanced logger buffer migration was a mechanism to address this but doing so complicates MM treatment of the buffer.

Because a single advanced logger buffer is used across all phases, including a shared buffer between PEI and MM and DXE and MM, the buffer must be unlocked for MM access to accommodate a PEI MM IPL. This creates a problem where both requirements cannot simultaneously be met without additional complexity in MM access code. In addition, MM environments are inconsistent in their support for memory unlocking.

In the end, hibernation has always been a best effort scenario with this flow and other MM solutions already require runtime allocations outside of DXE memory bins, so this change removes the functional logic for migration.

The memory bin scenario will be addressed in the future when the bins are allocated in a pre-DXE phase (e.g. PEI) and picked up by DXE.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- AdvLoggerPkg CI
- Boot to OS on QEMU Q35 and SBSA
- Boot to EFI shell on OVMF X64 w/ edk2 Standalone MM

## Integration Instructions

- N/A

### DIFF
--- a/AdvLoggerPkg/Include/AdvancedLoggerInternal.h
+++ b/AdvLoggerPkg/Include/AdvancedLoggerInternal.h
@@ -46,11 +46,6 @@
 #define ADVANCED_LOGGER_PHASE_CNT          11
 
 //
-// The maximum depth to follow when traversing chains of advanced logger info structures
-//
-#define ADVANCED_LOGGER_MAX_LOGGER_CHAIN_DEPTH  3
-
-//
 // These Pcds are used to carve out a PEI memory buffer from the temporary RAM.
 //
 //  PcdAdvancedLoggerBase -        NULL = UEFI starts with PEI, and SEC provides no memory log buffer
@@ -84,7 +79,9 @@ typedef volatile struct {
   EFI_TIME                Time;                   // Uefi Time Field
   UINT32                  HwPrintLevel;           // Logging level to be printed at hw port
   UINT32                  Reserved3;              //
-  EFI_PHYSICAL_ADDRESS    NewLoggerInfoAddress;   // If non-zero, this field holds the address of new logger info that should be used
+  EFI_PHYSICAL_ADDRESS    Deprecated;             // Deprecated field that should be preserved for binary compatibility
+                                                  // and not reused. Previous description: If non-zero, this field
+                                                  // holds the address of new logger info that should be used
 } ADVANCED_LOGGER_INFO;
 
 typedef struct {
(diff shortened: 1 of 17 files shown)


AdvLoggerPkg/Include/AdvancedLoggerInternal.h
cfernald: Should we mark `NewLoggerInfoAddress` reserved/deprecated?

cfernald (approved):

AdvLoggerPkg/Include/AdvancedLoggerInternal.h
os-d: I think we should, as long as we say that Reserved fields cannot be used in the future.

os-d (approved):

cfernald (approved):
re-approving since github seems to have glitched and didn't count it before.

kuqin12 (approved):

AdvLoggerPkg/Include/AdvancedLoggerInternal.h
makubacki: Renamed the field to `Deprecated`. Clarified old and new usage in the field comment.

makubacki:
Note: Waiting for final server platform testing to conclude before merging

apop5:
> Note: Waiting for final server platform testing to conclude before merging

This is good to go, it was a switch of the MmUnblockMemoryLib from an actual instance to the NULL instance that was causing the test hang. Once it switched back, the system booted.

apop5 (approved):

PR #877, opened 2026-04-24
AdvLoggerPkg: Update functionality for PeilessSec

apop5:
## Description

In PeilessSec platforms using a PcdAdvancedLoggerBase at a fixed location, there are other firmware components which may run prior to Uefi and create log messages.

Perform a check at the fixed location for a valid log signature. If the valid log signature is found, treat the log buffer as already initialized and then reinitialize the hwdport.

Found in a platform where pre UEFI firmware is initializing the log buffer, but using a different serial port from what the Uefi code uses.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Physical Platform

## Integration Instructions
No integration necessary (changes are backward compatible with existing implementation)


apop5:
closing due to not being needed by platform.

PR #886, opened 2026-05-07
AdvancedLoggerPkg/MmCoreArm: Initialize LoggerInfo if not already set by earlier firmware phase

sureshkumarpMSFT:
## Description

In the MmCoreArm instance of AdvancedLoggerLib, AdvancedLoggerGetLoggerInfo() previously assumed that an earlier firmware phase (e.g. BL31 prior to StandaloneMM) had already initialized the ADVANCED_LOGGER_INFO header at PcdAdvancedLoggerBase. On ARM boot flows where StandaloneMM is the first phase to touch this buffer, the signature check in subsequent logger calls fails and no log records are written.

This change adds a one-time, in-place initialization of the ADVANCED_LOGGER_INFO header inside AdvancedLoggerGetLoggerInfo() when the signature is not already valid:

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified on an ARM platform where no pre-StandaloneMM entity initializes the AdvancedLogger buffer

## Integration Instructions

N/A

### DIFF
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
@@ -24,6 +24,43 @@
 //
 
 /**
+  Initialize the AdvancedLogger header at PcdAdvancedLoggerBase if a previous
+  firmware phase has not already done so.
+
+  On boot flows where StandaloneMM runs before any phase that would normally
+  lay down the ADVANCED_LOGGER_INFO header (e.g. BL31 -> StMM -> BL33/UEFI),
+  the buffer signature will be invalid on first entry. This routine performs
+  a one-time initialization of the header so that subsequent logging calls
+  have a valid buffer to write into.
+
+  @param  LoggerInfo  Caller-supplied non-NULL pointer to the fixed buffer.
+**/
+STATIC
+VOID
+InitializeLoggerHeaderIfNeeded (
+  IN ADVANCED_LOGGER_INFO  *LoggerInfo
+  )
+{
+  UINTN  LogBufferSize;
+
+  if (LoggerInfo->Signature == ADVANCED_LOGGER_SIGNATURE) {
+    return;
+  }
+
+  LogBufferSize = EFI_PAGES_TO_SIZE (FixedPcdGet32 (PcdAdvancedLoggerPages));
+
+  ZeroMem ((VOID *)LoggerInfo, sizeof (ADVANCED_LOGGER_INFO));
+  LoggerInfo->Signature          = ADVANCED_LOGGER_SIGNATURE;
+  LoggerInfo->Version            = ADVANCED_LOGGER_INFO_VER;
+  LoggerInfo->LogBufferSize      = (UINT32)(LogBufferSize - sizeof (ADVANCED_LOGGER_INFO));
+  LoggerInfo->LogBufferOffset    = EXPECTED_LOG_BUFFER_OFFSET (LoggerInfo);
+  LoggerInfo->LogCurrentOffset   = LoggerInfo->LogBufferOffset;
+  LoggerInfo->HwPrintLevel       = FixedPcdGet32 (PcdAdvancedLoggerHdwPortDebugPrintErrorLevel);
+  LoggerInfo->HdwPortInitialized = TRUE;
+  LoggerInfo->InPermanentRAM     = TRUE;
+}
+
+/**
   The logger Information Block is carved from the Trust Zone at a specific fixed address.
 
   This address is obtained from the PcdAdvancedLoggerBase.  The size of the Advanced Logger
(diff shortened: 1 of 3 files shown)


AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -23,6 +23,43 @@
...
+**/
+STATIC
+VOID
+InitializeLoggerHeaderIfNeeded (
+  IN ADVANCED_LOGGER_INFO  *LoggerInfo
+  )
+{
+  UINTN  LogBufferSize;
+
+  if (LoggerInfo->Signature == ADVANCED_LOGGER_SIGNATURE) {
```
os-d, line 50: nit: This function is static, but the implementation may change over time, I would still do a null check here

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -23,6 +23,43 @@
...
+    return;
+  }
+
+  LogBufferSize = EFI_PAGES_TO_SIZE (FixedPcdGet32 (PcdAdvancedLoggerPages));
```
os-d, line 54: We probably aren't doing this in other implementations, but we should also ensure LogBufferSize > sizeof(ADVANCED_LOGGER_INFO).

That could be a separate commit that applies it to all the libs that initialize if you want.

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -23,6 +23,43 @@
...
+
+  ZeroMem ((VOID *)LoggerInfo, sizeof (ADVANCED_LOGGER_INFO));
+  LoggerInfo->Signature          = ADVANCED_LOGGER_SIGNATURE;
+  LoggerInfo->Version            = ADVANCED_LOGGER_INFO_VER;
+  LoggerInfo->LogBufferSize      = (UINT32)(LogBufferSize - sizeof (ADVANCED_LOGGER_INFO));
+  LoggerInfo->LogBufferOffset    = EXPECTED_LOG_BUFFER_OFFSET (LoggerInfo);
+  LoggerInfo->LogCurrentOffset   = LoggerInfo->LogBufferOffset;
+  LoggerInfo->HwPrintLevel       = FixedPcdGet32 (PcdAdvancedLoggerHdwPortDebugPrintErrorLevel);
+  LoggerInfo->HdwPortInitialized = TRUE;
```
os-d, line 71:
```suggestion
      AdvancedLoggerHdwPortInitialize ();
      LoggerInfo->HdwPortInitialized = TRUE;
```
I think we need to actually make sure the HW port is initialized

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -64,6 +101,12 @@ AdvancedLoggerGetLoggerInfo (
     return NULL;
   }
 
+  //
+  // First StMM caller in a boot lays down the header; subsequent callers
+  // (re-entries, DxeCore, etc.) find a valid signature and skip init.
```
os-d:
```suggestion
  // Check if the log has been initialized by a previous entity (e.g. TF-A)
  // or a previous log in StMM. If not, initialize.
```

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
```diff
@@ -23,6 +23,43 @@
...
+  Initialize the AdvancedLogger header at PcdAdvancedLoggerBase if a previous
+  firmware phase has not already done so.
+
+  On boot flows where StandaloneMM runs before any phase that would normally
+  lay down the ADVANCED_LOGGER_INFO header (e.g. BL31 -> StMM -> BL33/UEFI),
+  the buffer signature will be invalid on first entry. This routine performs
+  a one-time initialization of the header so that subsequent logging calls
+  have a valid buffer to write into.
+
+  @param  LoggerInfo  Caller-supplied non-NULL pointer to the fixed buffer.
```
makubacki:
```suggestion
  @param[in]  LoggerInfo  Caller-supplied non-NULL pointer to the fixed buffer.
```

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
sureshkumarpMSFT: Please check now

AdvLoggerPkg/Library/AdvancedLoggerLib/MmCoreArm/AdvancedLoggerLib.c
sureshkumarpMSFT: Please check the latest changes

os-d (approved):

apop5 (approved):

makubacki (approved):

apop5 (approved):

PR #888, opened 2026-05-11
Fix SWM init retry when SRE is not ready

mikitl:
## Description

GopRegisteredCallback sets mGop when it finds GOP but exits early if SRE is not available yet. Because mGop remains set, the callback never retries on subsequent GOP notifications, so SWM never initializes.

Reset mGop to NULL when SRE is not found so the callback can retry.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on:

 - ARM64 platform with GPU driver that installs GOP during early display enumeration, before GopOverrideDxe and SRE have processed it
 - Verified SWM callback retries and completes Stage2 initialization after SRE becomes available
 - Confirmed Front Page and on-screen keyboard render correctly
 - Verified no regression when GOP and SRE appear simultaneously (standard boot path)

## Integration Instructions

N/A

### DIFF
--- a/MsGraphicsPkg/SimpleWindowManagerDxe/WindowManager.c
+++ b/MsGraphicsPkg/SimpleWindowManagerDxe/WindowManager.c
@@ -1531,6 +1531,7 @@ GopRegisteredCallback (
                   );
 
   if (EFI_ERROR (Status)) {
+    mGop = NULL;
     goto Exit;
   }
 
--- a/MsGraphicsPkg/SimpleWindowManagerDxe/WindowManager.c
+++ b/MsGraphicsPkg/SimpleWindowManagerDxe/WindowManager.c
@@ -1518,9 +1518,6 @@ GopRegisteredCallback (
     goto Exit;
   }
 
-  mAbsPointerMode.AbsoluteMaxX = mGop->Mode->Info->HorizontalResolution;
-  mAbsPointerMode.AbsoluteMaxY = mGop->Mode->Info->VerticalResolution;
-
   // Determine if the Simple Rendering Engine Protocol is available on the same Console Out handle.  The
   // Rendering Engine driver provides both Graphics Output and Rendering Engine protocols.
   //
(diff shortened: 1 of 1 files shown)


joschock (approved):

MsGraphicsPkg/SimpleWindowManagerDxe/WindowManager.c
apop5, line 1550: Should the same mGop = NULL be applied to this error condition as well?

mikitl: I don't think so, at this stage `DriverCleanUp()` already called `Status = gBS->CloseEvent (mGopRegisterEvent);` making this callback unavailable if `gEfiGraphicsOutputProtocolGuid` get installed again.

We can move the `DriverCleanUp()` outside of the if-condition and add the `mGop = NULL` there if we want to retry in a subsequent `gEfiGraphicsOutputProtocolGuid` installation after `DriverInitStage2()` failed.

But I am not sure if this is the right design decision

MsGraphicsPkg/SimpleWindowManagerDxe/WindowManager.c
```diff
@@ -1531,6 +1531,7 @@ GopRegisteredCallback (
                   );
 
   if (EFI_ERROR (Status)) {
+    mGop = NULL;
```
makubacki, line 1531: If the GOP handle validation criteria requires that `gMsSREProtocolGuid` be on the same handle, it seems we should not set global state (`mAbsPointerMode.AbsoluteMaxX` & `mAbsPointerMode.AbsoluteMaxY`) based on it until after that is confirmed.

So, move Lines 1521 & 1522 below this check.

mikitl: Completed 👍

makubacki (approved):

apop5 (approved):

PR #896, opened 2026-06-15
[REBASE&FF] .azurepipelines: Add CLANGPDB CI

### LINKED ISSUE #895: [Bug]: Unremoved deprecated code in MemoryProtectionExceptionHandlerLib

### Current Behavior

MemoryProtectionExceptionHandlerLib can not be compiled(aarch64) due to the deprecated and removed file "MemoryProtectionNonstopMode.h".
https://github.com/microsoft/mu_plus/blob/f7c042a9d1428cdf6be773e2b8ca6cf51f66cb6f/MsCorePkg/Library/MemoryProtectionExceptionHandlerLib/MemoryProtectionExceptionHandlerLibAArch64.c#L18

Log:
```
INFO - Building ... MU_BASECORE/MdeModulePkg/Library/BaseExceptionPersistenceLibNull/BaseExceptionPersistenceLibNull.inf [AARCH64]
INFO - Common/MU/MsCorePkg/Library/MemoryProtectionExceptionHandlerLib/MemoryProtectionExceptionHandlerLibAArch64.c:18:10: fatal error: 'Protocol/MemoryProtectionNonstopMode.h' file not found
INFO -    18 | #include <Protocol/MemoryProtectionNonstopMode.h>
INFO -       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### Expected Behavior

This include line should be removed and a functionality check is required.

### Steps To Reproduce

Compile with this lib when building AARCH64 target => lack the header => BOOM.

### Build Environment

```markdown
- OS(s): Ubuntu 24.04
- Tool Chain(s): CLANGPDB
- Targets Impacted: ALL
- Target Arch: AARCH64
```

### Version Information

```text
commit: f7c042a9d1428cdf6be773e2b8ca6cf51f66cb6f
```

### Urgency

Medium

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

Javagedes:
## Description

Add CI that specifically builds each package in DEBUG and RELEASE with CLANGPDB on both Windows and Linux. Applies necessary changes to compile all packages with CLANGPDB.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Windows: https://dev.azure.com/projectmu/mu/_build/results?buildId=111589&view=results
Linux: https://dev.azure.com/projectmu/mu/_build/results?buildId=111590&view=results

## Integration Instructions

N/A

### DIFF
--- a/.azurepipelines/Windows-CLANGPDB.yml
+++ b/.azurepipelines/Windows-CLANGPDB.yml
@@ -1,0 +1,68 @@
+## @file
+# Azure Pipeline build file for a build using Windows and the CLANG PDB toolchain.
+#
+# IMPORTANT: This file requires the local repo to contain a .azurepipelines/Matrix-Build-Job.yml file that
+#            performs the actual build steps for the configurations and packages supported by the repo.
+#
+# NOTE: This file is automatically synchronized from Mu DevOps. Update the original file there
+#       instead of the file in this repo.
+#
+# - Mu DevOps Repo: https://github.com/microsoft/mu_devops
+# - File Sync Settings: https://github.com/microsoft/mu_devops/blob/main/.sync/Files.yml
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+variables:
+- group: architectures-x86-64
+- group: tool-chain-windows-clangpdb
+- group: coverage
+
+extends:
+  template: MuDevOpsWrapper.yml
+  parameters:
+    do_ci_build: true
+    do_ci_setup: true
+    do_non_ci_build: false
+    do_non_ci_setup: true
+    do_pr_eval: true
+    rust_build: true
+    os_type: Windows_NT
+    build_matrix:
+      TARGET_MFCI_XML:
+        Pkgs: 'XmlSupportPkg,MfciPkg'
+        Targets: 'DEBUG,RELEASE'
+        ArchList: $(arch_list)
+      TARGET_GRAPHICS_WHEA_HID:
+        Pkgs: 'MsGraphicsPkg,MsWheaPkg,HidPkg'
+        Targets: 'DEBUG,RELEASE'
+        ArchList: $(arch_list)
+      TARGET_CORE:
+        Pkgs: 'MsCorePkg'
+        Targets: 'DEBUG,RELEASE'
+        ArchList: $(arch_list)
+      TARGET_TEST_PCBDS_LOG:
+        Pkgs: 'TpmTestingPkg,UefiTestingPkg,PcBdsPkg,AdvLoggerPkg'
+        Targets: 'DEBUG,RELEASE'
+        ArchList: $(arch_list)
+      TARGET_MFCI_XML_ARM:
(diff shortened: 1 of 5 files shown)


.azurepipelines/Windows-CLANGPDB.yml
```diff
@@ -0,0 +1,68 @@
...
+#
+# - Mu DevOps Repo: https://github.com/microsoft/mu_devops
+# - File Sync Settings: https://github.com/microsoft/mu_devops/blob/main/.sync/Files.yml
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+variables:
+- group: architectures-x86-64
```
antklein, line 18: Any reason to not build Windows arm64 as well?
I agree that it's most likely redundant, but wanted to make sure this was done on purpose.

Javagedes: I don't believe any MU repository currently builds on a Windows ARM runner. This repository definitely does not, just from looking at the jobs that run as a PR gate.

I'm not personally against adding that support, but I'm not sure if there has already been discussion regarding this within the team. The main thing I can think of is if that support was added, it would increase our already congested pipeline usage.

@makubacki or @kuqin12 may have more context.

apop5 (approved):

makubacki:
@Javagedes, is the PR ready for review now?

.azurepipelines/Windows-CLANGPDB.yml
makubacki: I'm not opposed to it either, but a large part of why we built on different hosts in the past was toolchain/host compatibility - VS on Windows hosts and GCC on Linux hosts. And then there was the problem of Arm assembly not fully being supported by MASM, etc. For this and other reasons (like Arm platform dependencies sometimes only working on Linux hosts & GCC), we haven't broadly tested Windows on Arm upstream.

If there is substantial reason to verify the build on Windows hosts, like unit tests or something else, we can take that as a separate feature request in mu_devops and figure out how to best apply it keeping agent load in mind.

Javagedes:
> @Javagedes, is the PR ready for review now?

Yes

os-d (approved):
This doesn’t actually fix the reported error; did that not repro or are we not building the lib in CI?

makubacki:
@Javagedes, [this commit](https://github.com/microsoft/mu_plus/pull/896/changes/6766b9bf560fec5dac059b6b09a1dbf54af4d06f) message has a typo that makes it difficult to read:

> Any if the error handling conditions that cause a
goto `Done` will call `if (Blt != NULL)` even
though Blt may not have been initialized. This
commit sets Blt to null at the beggining of the
function to resolve this.

If you could, please fix that before merging.

makubacki (approved):

makubacki:
> If you could, please fix that before merging.

Since @Javagedes is unavailable at the moment, I made that update. Given the pipeline files look reasonable and it would be beneficial to start getting coverage, I am merging the PR now.

PR #876, opened 2026-04-17
AdvLoggerPkg/AdvancedLoggerLib: Print to HW if logger info is not available

makubacki:
## Description

In the course of testing advanced logger changes with edk2 Standalone MM on OVMF, there was a change needed in MmPlatformHobProducerLibOvmf in OvmfPkg to account for gAdvancedLoggerHobGuid in the HOBs that are passed to the MM environment.

Prior to this change, the system appeared to hang entering that environment:

```
INFO - Loading PEIM at 0x00005B95000 EntryPoint=0x00005B95340 StandaloneMmIplPei.efi
INFO - StandaloneMM IPL loading MM Core at MMRAM address 7FF0000
INFO - MmCoreImageBase  - 0x0000000007FF0000
INFO - MmCoreImageSize  - 0x0000000000010000
INFO - StandaloneMM IPL calling Standalone MM Core at MMRAM address - 0x0000000007FF1000
```

After the change, the issue is clearly printed to serial output: ("MmCore AdvancedLoggerGetLoggerInfo: Advanced Logger Hob not found")

```
INFO - Loading PEIM at 0x00005B94000 EntryPoint=0x00005B94340 StandaloneMmIplPei.efi
INFO - StandaloneMM IPL loading MM Core at MMRAM address 7FF0000
INFO - MmCoreImageBase  - 0x0000000007FF0000
INFO - MmCoreImageSize  - 0x0000000000010000
INFO - StandaloneMM IPL calling Standalone MM Core at MMRAM address - 0x0000000007FF1000
INFO - MmCore AdvancedLoggerGetLoggerInfo: Advanced Logger Hob not found
INFO - MmMain - 0x5B11000
INFO - MmramRangeCount - 0x4
INFO - MmramRanges[0]: 0x0000000006000000 - 0x1000
INFO - MmramRanges[1]: 0x0000000006001000 - 0xE000
INFO - MmramRanges[2]: 0x000000000600F000 - 0x1FE1000
INFO - MmramRanges[3]: 0x0000000007FF0000 - 0x10000
INFO - MmInitializeMemoryServices
INFO - MmAddMemoryRegion 2 : 0x000000000600F000 - 0x0000000001FE1000
INFO - HobSize - 0x538
INFO - MmHobStart - 0x7FEE810
INFO - MmInstallConfigurationTable For HobList
INFO - ASSERT [StandaloneMmCore] AdvancedLoggerLib.c(164): mLoggerInfo != ((void *) 0)
```

[...]

PR #902, opened 2026-06-18
AdvLoggerPkg: Gate hardware port writes at OS runtime

chaokai-ching:
## Description

`AdvancedLoggerWrite` unconditionally wrote debug output to the hardware serial
port at both boot time and OS runtime (after ExitBootServices). Some platforms'
serial implementations cannot be safely called at OS runtime.

This change adds an **opt-in** FeatureFlag PCD
`PcdAdvancedLoggerHdwPortRuntimeDisable` (default **FALSE**):

- **Default (FALSE):** behavior is unchanged — hardware port writes remain
  enabled at runtime, so existing platforms are unaffected.
- **TRUE:** a platform whose serial cannot be safely called after
  ExitBootServices restricts hardware port writes to boot time only, while
  early-boot and boot-time serial output is preserved.

Runtime is detected via the logger info block's `AtRuntime` field when
available. The DXE runtime `AdvancedLoggerLib` instance clears its logger info
pointer at ExitBootServices, so it consults a module-scoped
`ADVANCED_LOGGER_RUNTIME` build flag plus a `gAdvancedLoggerAtRuntime` flag to
distinguish post-EBS runtime from early boot when the logger info block is NULL.

- [x] Impacts functionality? Only when the new PCD is set TRUE (default FALSE = no change)
- [ ] Impacts security?
- [ ] Breaking change? No — default preserves existing behavior
- [ ] Includes tests? Not included; hardware-port gating path has no existing host-test coverage. Can add a GoogleTest if desired.
- [ ] Includes documentation? PCD documented in AdvLoggerPkg.dec

## How This Was Tested

Built an ARM AARCH64 platform (DEBUG); all `AdvancedLoggerLib` instances
compile. Verified on ARM AARCH64 hardware with a temporary driver that forces a
serial write on the first post-SetVirtualAddressMap runtime call:

- PCD default (FALSE): the firmware faults on the runtime serial write
  (reproduces the unsafe-at-runtime behavior).
- PCD TRUE: the runtime write is suppressed and the system boots cleanly,
  while boot-time serial output is unaffected.

## Integration Instructions

None required by default. Platforms whose serial implementation cannot be
called safely at OS runtime should set
`gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerHdwPortRuntimeDisable` to TRUE.

### DIFF
--- a/AdvLoggerPkg/AdvLoggerPkg.dec
+++ b/AdvLoggerPkg/AdvLoggerPkg.dec
@@ -99,14 +99,14 @@
   #
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerAutoWrapEnable|FALSE|BOOLEAN|0x00010188
 
-  ## PcdAdvancedLoggerHdwPortRuntimeDisable - Suppress hardware (serial) port writes at OS
+  ## PcdAdvancedLoggerHdwPortOsRuntimeDisable - Suppress hardware port writes at OS
   # runtime (after ExitBootServices). By default (FALSE), behavior is unchanged: hardware port
   # writes remain enabled at runtime, matching the historical default. Set to TRUE only on a
-  # platform whose underlying serial implementation cannot be safely called after
+  # platform whose underlying hardware port implementation cannot be safely called after
   # ExitBootServices; this restricts hardware port writes to boot time while preserving
-  # early-boot and boot-time serial output.
+  # early-boot and boot-time hardware port output.
   #
-  gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerHdwPortRuntimeDisable|FALSE|BOOLEAN|0x0001018A
+  gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerHdwPortOsRuntimeDisable|FALSE|BOOLEAN|0x0001018A
 
 
 [PcdsFixedAtBuild]
--- a/AdvLoggerPkg/AdvLoggerPkg.dec
+++ b/AdvLoggerPkg/AdvLoggerPkg.dec
@@ -100,11 +100,9 @@
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerAutoWrapEnable|FALSE|BOOLEAN|0x00010188
 
   ## PcdAdvancedLoggerHdwPortOsRuntimeDisable - Suppress hardware port writes at OS
-  # runtime (after ExitBootServices). By default (FALSE), behavior is unchanged: hardware port
-  # writes remain enabled at runtime, matching the historical default. Set to TRUE only on a
-  # platform whose underlying hardware port implementation cannot be safely called after
-  # ExitBootServices; this restricts hardware port writes to boot time while preserving
-  # early-boot and boot-time hardware port output.
+  # runtime (after ExitBootServices). By default (FALSE), hardware port writes remain
+  # enabled at runtime. Set to TRUE on a platform whose underlying hardware port
+  # implementation cannot be safely called after ExitBootServices.
   #
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerHdwPortOsRuntimeDisable|FALSE|BOOLEAN|0x0001018A
 
--- a/AdvLoggerPkg/AdvLoggerPkg.dec
+++ b/AdvLoggerPkg/AdvLoggerPkg.dec
@@ -99,6 +99,15 @@
   #
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerAutoWrapEnable|FALSE|BOOLEAN|0x00010188
 
+  ## PcdAdvancedLoggerHdwPortRuntimeDisable - Suppress hardware (serial) port writes at OS
+  # runtime (after ExitBootServices). By default (FALSE), behavior is unchanged: hardware port
+  # writes remain enabled at runtime, matching the historical default. Set to TRUE only on a
+  # platform whose underlying serial implementation cannot be safely called after
+  # ExitBootServices; this restricts hardware port writes to boot time while preserving
+  # early-boot and boot-time serial output.
+  #
+  gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerHdwPortRuntimeDisable|FALSE|BOOLEAN|0x0001018A
+
 
 [PcdsFixedAtBuild]
   ## Advanced Logger Base - NULL = UEFI starts with PEI or DXE, and there is no SEC, or SEC
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
@@ -136,6 +136,38 @@ AdvancedLoggerMemoryLoggerWrite (
 }
 
 /**
+  Returns whether the given message's debug level is enabled for the hardware port.
+
+  This is the generic hardware port debug level policy shared by the Advanced Logger instances
+  that honor the logger info block's dynamic level. Instances that must not consult the dynamic
+  level (such as SEC instances) provide their own AdvancedLoggerPrintToHwPort instead of using
+  this helper.
+
+  @param  LoggerInfo  The logger info block, or NULL if it is not available.
+  @param  DebugLevel  The debug level of the message being logged.
+
+  @retval TRUE   The message's debug level is enabled for the hardware port.
+  @retval FALSE  The message's debug level is not enabled for the hardware port.
+**/
+BOOLEAN
+EFIAPI
+AdvancedLoggerHwPortLevelEnabled (
+  IN ADVANCED_LOGGER_INFO  *LoggerInfo,
+  IN UINTN                 DebugLevel
+  )
+{
+  UINT32  HwPortDebugLevel;
+
+  HwPortDebugLevel = PcdGet32 (PcdAdvancedLoggerHdwPortDebugPrintErrorLevel);
+
+  if ((LoggerInfo != NULL) && (LoggerInfo->Version >= ADVANCED_LOGGER_INFO_HW_LVL_SUPPORTED_VER)) {
+    HwPortDebugLevel = LoggerInfo->HwPrintLevel;
+  }
+
+  return (BOOLEAN)((DebugLevel & HwPortDebugLevel) != 0);
+}
+
+/**
   Write data from buffer to possible debugging devices.
 
   This is the interface from PeiCore
@@ -160,25 +192,12 @@ AdvancedLoggerWrite (
   )
 {
   ADVANCED_LOGGER_INFO  *LoggerInfo;
-  UINT32                HwPortDebugLevel;
 
   // All messages go to the in memory log.
   LoggerInfo = AdvancedLoggerMemoryLoggerWrite (DebugLevel, Buffer, NumberOfBytes);
 
-  HwPortDebugLevel = PcdGet32 (PcdAdvancedLoggerHdwPortDebugPrintErrorLevel);
-
-  // Only selected messages go to the hdw port.
-
-  if ((LoggerInfo == NULL) || (!LoggerInfo->HdwPortDisabled)) {
- #ifndef ADVANCED_LOGGER_SEC
-    if ((LoggerInfo != NULL) && (LoggerInfo->Version >= ADVANCED_LOGGER_INFO_HW_LVL_SUPPORTED_VER)) {
-      HwPortDebugLevel = LoggerInfo->HwPrintLevel;
-    }
-
- #endif
-
-    if ((DebugLevel & HwPortDebugLevel) && AdvancedLoggerPrintToHwPort (LoggerInfo)) {
-      AdvancedLoggerHdwPortWrite (DebugLevel, (UINT8 *)Buffer, NumberOfBytes);
-    }
+  // Only selected messages go to the hardware port.
+  if (AdvancedLoggerPrintToHwPort (LoggerInfo, DebugLevel)) {
+    AdvancedLoggerHdwPortWrite (DebugLevel, (UINT8 *)Buffer, NumberOfBytes);
   }
 }
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
@@ -19,6 +19,13 @@
 
 #include "../AdvancedLoggerCommon.h"
 
+#ifdef ADVANCED_LOGGER_RUNTIME
+//
+// Defined by the DXE runtime AdvancedLoggerLib instance; TRUE after ExitBootServices.
+//
+extern BOOLEAN  gAdvancedLoggerAtRuntime;
+#endif
+
 /**
   Write data from buffer into the in memory logging buffer.
 
@@ -161,6 +168,8 @@ AdvancedLoggerWrite (
 {
   ADVANCED_LOGGER_INFO  *LoggerInfo;
   UINT32                HwPortDebugLevel;
+  BOOLEAN               HwPortWriteAllowed;
+  BOOLEAN               AtRuntime;
 
   // All messages go to the in memory log.
   LoggerInfo = AdvancedLoggerMemoryLoggerWrite (DebugLevel, Buffer, NumberOfBytes);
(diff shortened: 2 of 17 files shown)


chaokai-ching:
@microsoft-github-policy-service agree company="Microsoft"

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
```diff
@@ -177,7 +186,50 @@ AdvancedLoggerWrite (
...
+      //
+      if (LoggerInfo != NULL) {
+        AtRuntime = LoggerInfo->AtRuntime;
+      } else {
+ #ifdef ADVANCED_LOGGER_RUNTIME
+        //
+        // The DXE runtime instance clears its logger info pointer at ExitBootServices, so a
+        // NULL block is ambiguous between early boot and runtime; consult the runtime flag.
+        //
+        AtRuntime = gAdvancedLoggerAtRuntime;
```
kuqin12: Do we really need 3 knobs to make this happen? The runtime instance already has the signal by setting the gBS and mLoggerInfo to null at EBS, I assume that is enough to be treated as an indicator.

Note that this behavior was newly introduced in https://github.com/microsoft/mu_plus/pull/876/changes. I think we can apply the new PCD to that old code and it would work?

apop5: I agree with Kun. I think we can accomplish the same behavior by checking mBS for just the runtime, and avoid additional variables.

AdvLoggerPkg/AdvLoggerPkg.dec
```diff
@@ -99,6 +99,15 @@
...
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerAutoWrapEnable|FALSE|BOOLEAN|0x00010188
 
+  ## PcdAdvancedLoggerHdwPortRuntimeDisable - Suppress hardware (serial) port writes at OS
+  # runtime (after ExitBootServices). By default (FALSE), behavior is unchanged: hardware port
+  # writes remain enabled at runtime, matching the historical default. Set to TRUE only on a
+  # platform whose underlying serial implementation cannot be safely called after
+  # ExitBootServices; this restricts hardware port writes to boot time while preserving
+  # early-boot and boot-time serial output.
+  #
+  gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerHdwPortRuntimeDisable|FALSE|BOOLEAN|0x0001018A
```
makubacki: "Runtime" is overloaded. Please call this "OsRuntime".

```suggestion
  gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerHdwPortOsRuntimeDisable|FALSE|BOOLEAN|0x0001018A
```

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
```diff
@@ -177,7 +186,50 @@ AdvancedLoggerWrite (
...
+    // By default hardware port writes are always allowed, preserving the historical
+    // behavior of writing debug output to the serial port at both boot time and OS runtime.
+    //
+    // The underlying hardware serial port implementation on some platforms cannot be safely
+    // called at OS runtime (after ExitBootServices). Such a platform sets the FeatureFlag PCD
+    // PcdAdvancedLoggerHdwPortRuntimeDisable to TRUE to restrict hardware port writes to boot
+    // time only. The runtime detection below is therefore only needed for those opt-in
+    // platforms; for everyone else the PCD is a compile-time FALSE and this whole block is
+    // optimized away, leaving behavior and code size unchanged.
+    //
```
makubacki: These large comments blocks are redundant with the code, affect code readability, and should be dropped imo.

```suggestion
```

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
makubacki: - I do not think setting `mLoggerInfo` to `NULL` is a good way to indicate "at runtime" (especially when `ADVANCED_LOGGER_INFO` itself has an `AtRuntime` field).
  - I know the field is used for cases outside this condition.
  - It is fine to set it to `NULL` at EBS (and fine to check for `mLoggerInfo == NULL` when needed of course), but it being `NULL` does not exclusively mean "at runtime", it means "logger info not available" (which includes the `AtRuntime` field).
  - So, while it functionally could work it's unnecessarily confusing and mixes responsibilities.
- `AdvancedLoggerWrite()` should have a clear "At Runtime" dependency if that is a dependency, not a subtle coupling with logger info being available (which might be the case for a number of reasons).

1. Please keep the code in 876 to address that issue.
2. Make "at runtime" status clear and obvious (however that is done) as a dependency in `AdvancedLoggerWrite()`.
   - A new variable, etc.
3. Simply compare the PCD value against that status and then the PCD and status will be obviously reconciled against each other.

kuqin12: Agree that checking logger info being null truly means the code running at runtime or not. But this function has coupled excessive phase specific checks. We should not compound yet another one into the same function just to keep them the same function.

We might be better served if we have a `bool AdvancedLoggerPrintToHwPort (loggerinfo, level)` provided by each instance. Then this function can just check this function and print if this new function return true.

makubacki: That also works. My main ask is we untangle the subtle coupling for the runtime case with logger info being `NULL` and make it clear what's happening.

I think that something like a `AdvancedLoggerPrintToHwPort()` function to abstract instance-specific logic would be nice and prevents the runtime details from leaking into the common code.

chaokai-ching: Each instance can now override the HW-port decision via `AdvancedLoggerPrintToHwPort()`. 
It either uses the common default or its own override, so `AdvancedLoggerWrite()` holds no phase-specific logic. At this time, only the Runtime instance implements the override, and the build flag and extern global are removed. 
It also has simplified gating logic in the `AdvancedLoggerCommon.c`

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
chaokai-ching: Done. removed

AdvLoggerPkg/Library/AdvancedLoggerLib/Sec/AdvancedLoggerLib.inf
```diff
@@ -49,6 +50,7 @@
 
 [FeaturePcd]
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerAutoWrapEnable
+  gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerHdwPortOsRuntimeDisable
```
kuqin12: With the new function added, we probably do not need this PCD sprinkled in all the inf files?

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
```diff
@@ -177,7 +177,7 @@ AdvancedLoggerWrite (
 
  #endif
 
-    if (DebugLevel & HwPortDebugLevel) {
+    if ((DebugLevel & HwPortDebugLevel) && AdvancedLoggerPrintToHwPort (LoggerInfo)) {
       AdvancedLoggerHdwPortWrite (DebugLevel, (UINT8 *)Buffer, NumberOfBytes);
     }
```
kuqin12: I was thinking about this:

```
if (AdvancedLoggerPrintToHwPort (LoggerInfo, HwPortDebugLevel )) {
  AdvancedLoggerHdwPortWrite (DebugLevel, (UINT8 *)Buffer, NumberOfBytes);
}
```

And `AdvancedLoggerPrintToHwPort` will host the logic covered by this comment. Then there will be no `ADVANCED_LOGGER_SEC`.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerHwPort.c
```diff
@@ -0,0 +1,50 @@
+/** @file
+  Default AdvancedLoggerPrintToHwPort implementation, shared by the Advanced Logger library
+  instances that do not override it for special hardware port handling at OS runtime.
+
+  Permits hardware port writes at boot time, and at OS runtime unless a platform
+  opts in to PcdAdvancedLoggerHdwPortOsRuntimeDisable. The DXE runtime instance provides its
+  own implementation (it clears its logger info block at ExitBootServices) instead of using
+  this file.
+
+  Copyright (C) Microsoft Corporation. All rights reserved.
```
kuqin12:
```suggestion
  Copyright (c) Microsoft Corporation.
```

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerHwPort.c
```diff
@@ -0,0 +1,50 @@
...
+  @retval TRUE   Hardware port writes are permitted.
+  @retval FALSE  Hardware port writes are currently suppressed (OS runtime, opt-in platform).
+**/
+BOOLEAN
+EFIAPI
+AdvancedLoggerPrintToHwPort (
+  IN ADVANCED_LOGGER_INFO  *LoggerInfo
+  )
+{
+  if (FeaturePcdGet (PcdAdvancedLoggerHdwPortOsRuntimeDisable) &&
```
kuqin12: As mentioned, this PCD is not relevant in the "common" implementation. We would only want to use that in the runtime instance, correct?

chaokai-ching: Line 264 in AdvLoggerSmmAccessLib.c sets AtRuntime as well. If we remove the PcdAdvancedLoggerHdwPortOsRuntimeDisable check, we will not be able to capture the runtime status.

kuqin12: Good observation. But with this refactor in place, do we need to check `LoggerInfo->AtRuntime` here at all? Seems to me, the runtime check is only relevant to the runtime version.

AdvLoggerPkg/Library/AdvancedLoggerLib/Runtime/AdvancedLoggerLib.c
```diff
@@ -130,6 +138,38 @@ AdvancedLoggerGetPhase (
...
+{
+  BOOLEAN  AtOsRuntime;
+
+  if (!FeaturePcdGet (PcdAdvancedLoggerHdwPortOsRuntimeDisable)) {
+    return TRUE;
+  }
+
+  AtOsRuntime = (LoggerInfo != NULL) ? LoggerInfo->AtRuntime : mAdvancedLoggerAtRuntime;
+
+  return (BOOLEAN)(!AtOsRuntime);
```
makubacki: You shouldn't need to the `BOOLEAN` cast here since `AtOsRuntime` is `BOOLEAN` and `AdvancedLoggerPrintToHwPort()` returns a `BOOLEAN` type.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerHwPort.c
```diff
@@ -0,0 +1,50 @@
...
+#include "AdvancedLoggerCommon.h"
+
+/**
+  Returns whether this Advanced Logger instance permits writing debug output to the
+  hardware port.
+
+  Hardware port writes are always permitted unless the platform sets
+  PcdAdvancedLoggerHdwPortOsRuntimeDisable, in which case they are suppressed once at OS
+  runtime (after ExitBootServices), as reported by the logger info block's AtRuntime field.
+
```
makubacki: We normally avoid putting implementation-level detail in interface descriptions. This can be described in one place (like the PCD description).

```suggestion
```

AdvLoggerPkg/AdvLoggerPkg.dec
```diff
@@ -99,6 +99,15 @@
   #
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerAutoWrapEnable|FALSE|BOOLEAN|0x00010188
 
+  ## PcdAdvancedLoggerHdwPortOsRuntimeDisable - Suppress hardware port writes at OS
+  # runtime (after ExitBootServices). By default (FALSE), behavior is unchanged: hardware port
+  # writes remain enabled at runtime, matching the historical default. Set to TRUE only on a
+  # platform whose underlying hardware port implementation cannot be safely called after
+  # ExitBootServices; this restricts hardware port writes to boot time while preserving
+  # early-boot and boot-time hardware port output.
```
makubacki: This is pretty verbose and has some info that isn't very relevant long term.


```suggestion
  ## PcdAdvancedLoggerHdwPortOsRuntimeDisable - Suppress hardware port writes at OS
  # runtime (after ExitBootServices). By default (FALSE), hardware port writes remain
  # enabled at runtime. Set to TRUE on a platform whose underlying hardware port
  # implementation cannot be safely called after ExitBootServices.
```

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerHwPort.c
makubacki: I agree. This version should be generic. With instance-specific logic abstracted to instances that need specific considerations. The other place that might need this are MM instances (which, to my understanding, retain access to `mLoggerInfo->AtRuntime` after Exit Boot Services).

chaokai-ching: I think I get your point now - we don't need to account for MM usage at this time. If a module requires OS‑runtime hardware‑port support, it can override AdvancedLoggerPrintToHwPort  (the same way the Runtime instance does). I'll drop the PCD and AtRuntime from the common implementation and scope it to the Runtime instance for now.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
chaokai-ching: `ADVANCED_LOGGER_SEC` is now removed. `AdvancedLoggerPrintToHwPort` now hosts the full decision of whether to print to the HW port; each phase module either uses the common implementation or provides its own.

AdvLoggerPkg/Library/AdvancedLoggerLib/Sec/AdvancedLoggerLib.inf
chaokai-ching: Agree. Removed

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerHwPort.c
chaokai-ching: Change applied.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerHwPort.c
chaokai-ching: Removed

AdvLoggerPkg/AdvLoggerPkg.dec
chaokai-ching: Good suggestion. The comment block is now updated

AdvLoggerPkg/Library/AdvancedLoggerLib/Runtime/AdvancedLoggerLib.c
chaokai-ching: Done - cast is now removed.

chaokai-ching:
The **Mu Plus PR EDK2 Linux CLANGPDB (Build_TARGET_CORE)** check failed, but the job log looks unrelated to this change. The first failure I see is in the *Download Cargo Binstall (Linux)* setup step, while downloading the `Binaries` artifact: `requests.exceptions.ChunkedEncodingError: Response ended prematurely`. After that, several steps were skipped, the job reported `stuart_pr_eval: command not found`, and no `Build` folder was produced (`Not found SourceFolder: /__w/1/s/Build`). I don't see any compiler or EDK2 build invocation in this job log, so it doesn't look like the sources were compiled in this run. The checks matrix also shows the same CORE target passing on the GCC and Windows CLANGPDB pipelines, and *Analyze (AdvLoggerPkg)* passing — which makes me think this is a transient infrastructure / artifact-download issue rather than the code.

A re-run of the failed job should clear it.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
```diff
@@ -135,6 +135,38 @@ AdvancedLoggerMemoryLoggerWrite (
...
+  UINT32  HwPortDebugLevel;
+
+  HwPortDebugLevel = PcdGet32 (PcdAdvancedLoggerHdwPortDebugPrintErrorLevel);
+
+  if ((LoggerInfo != NULL) && (LoggerInfo->Version >= ADVANCED_LOGGER_INFO_HW_LVL_SUPPORTED_VER)) {
+    HwPortDebugLevel = LoggerInfo->HwPrintLevel;
+  }
+
+  return (BOOLEAN)((DebugLevel & HwPortDebugLevel) != 0);
+}
```
kuqin12, line 168: Maybe I missed something, but why do we need to double abstract this logic into another function?

kuqin12: Okay, I guess it is to reuse the logic in runtime instance.

kuqin12 (approved):

kuqin12:
> A re-run of the failed job should clear it.

Tried to rerun. Will keep an eye on this.

AdvLoggerPkg/Library/AdvancedLoggerLib/AdvancedLoggerCommon.c
chaokai-ching: yup, the helper function is for the default and Runtime to share the debug-level logic

kuqin12:
> > A re-run of the failed job should clear it.
> 
> Tried to rerun. Will keep an eye on this.

All good now.

makubacki (approved):

PR #909, opened 2026-06-25
[REBASE & FF] AdvLoggerPkg: PeiCore Instance: Use Local PeiMain.h

os-d:
## Description

The PEI Core AdvLoggerLib instance relies on PeiMain.h which defines the PEI_CORE_INSTANCE structure. A special PlatformBlob is added there to use stack space for the adv logger info pointer. This avoids a HOB lookup for every log.

However, edk2's PeiMain.h has been updated to use a private header in PeiMain.h. This is pulled into mu_basecore, which then breaks building this library because it doesn't have access to the private header.

Using PEI_CORE_INSTANCE to host the advanced logger pointer has always been a hack; it will be separately addressed and removed, but in the meantime, this library needs to continue to work.

This copies PeiMain.h, removes the use of the private header, and consumes it in the PeiCore instance of AdvLoggerLib.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booting patina-qemu to shell with the PEI memory bin changes cherry-picked.

## Integration Instructions

N/A.

### DIFF
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.inf
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.inf
@@ -5,6 +5,7 @@
 #
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
 #
+#Override : 00000002 | MdeModulePkg/Core/Pei/PeiMain.h | 5f8415080619b4d12d294f54e3fec16d | 2026-06-25T18-37-10 | 3fcfa38b0117f38468e2106feef56feb9fd47e28
 ##
 
 [Defines]
@@ -25,6 +26,7 @@
   ../AdvancedLoggerCommon.h
   ../AdvancedLoggerCommon.c
   ../AdvancedLoggerHwPort.c
+  PeiMain.h
 
 [Packages]
   MdePkg/MdePkg.dec
--- a/AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.c
+++ b/AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.c
@@ -15,13 +15,13 @@
 #include <Library/AdvancedLoggerAccessLib.h>
 
 /**
-  Including the PeiMain.h from PeiCore in order to access the Platform Blob data member.
+  Include a copy of PeiMain.h from PeiCore in order to access the Platform Blob data member.
 
   This is breaking the rules, but PeiCore on a system using ROM for PeiPreMem has no place
   to store long term data besides the Hob or Ppi list.  Accessing these list for high
   frequency operations is a performance issue.
 **/
-#include <Core/Pei/PeiMain.h>
+#include "PeiMain.h"
 
 #include <Ppi/AdvancedLogger.h>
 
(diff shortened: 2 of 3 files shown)


makubacki:
@os-d, since this hinges on following up with dropping the hack, can you go ahead and create a tracking issue for it?

AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.inf
```diff
@@ -63,3 +62,9 @@
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerPreMemPages                  ## CONSUMES
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerPages                        ## CONSUMES
   gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerHdwPortDebugPrintErrorLevel  ## CONSUMES
+
+[BuildOptions]
+  # This is a hack. It allows this library to use PeiMain.h which includes a private header.
+  # This is temporary and the use of PeiMain.h will be removed.
+  GCC:*_*_*_CC_FLAGS  = -I$(WORKSPACE)/MU_BASECORE/MdeModulePkg/Core/PrivateInclude
+  MSFT:*_*_*_CC_FLAGS = /I$(WORKSPACE)/MU_BASECORE/MdeModulePkg/Core/PrivateInclude
```
apop5: This will only work for repos that have mu_basecore under the root of the workspace

For cases where mu_basecore exists under Common/ this is going to break as well.

Not sure that there is a way around this

@Javagedes might have a different suggestion

os-d:
> @os-d, since this hinges on following up with dropping the hack, can you go ahead and create a tracking issue for it?

Created https://github.com/microsoft/mu_plus/issues/910

AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.inf
os-d: I was not aware of any platforms that didn't have basecore under the root. But this is true. If we believe such platforms exist (and it isn't just under Common as well, that can also be handled directly), I can reverse this workaround and make the private header non-private, though I was hoping to avoid changing the core code.

I could also create a local definition of PEI_CORE_INSTANCE and simply have it refer to that.

Problems with all solutions, which is why I want to get rid of the hack :)

makubacki: Could you just add a temporary, local, custom header that duplicates the types? You could even add an override tag.

os-d: Yeah, that's what I mean. I'll update to do that, I agree it is less brittle.

os-d: Added. I'll update the override tag when the basecore change goes in.

makubacki (approved):
Approving, with the understanding that this is a necessary temporary workaround while the GitHub issue is resolved.

apop5 (approved):

PR #912, opened 2026-06-26
[REBASE & FF] AdvLoggerPkg: PeiCore Instance: Update Override Tag And Apply SecDebugAgent Fix

os-d:
## Description

Update the override tag in PeiMain.h as it has been updated in basecore v2025110002.0.8.

Also, SecDebugAgent was missed when using the local header, update it.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

This is marked as a breaking change because the build will break if mu_basecore v2025110002.0.8 is not updated to (or later) at the same time by consumers.


apop5 (approved):

kuqin12 (approved):

apop5 (approved):

makubacki (approved):
Approving to amend the hack with GitHub issue tracking long term solution.

PR #913, opened 2026-06-26
AdvLoggerPkg: Use Scratchpad Region instead of PEI_CORE_INSTANCE

### LINKED ISSUE #910: [Task]: Create Scratchpad Space for PeiCore AdvancedLoggerLib and Drop PEI_CORE_INSTANCE Hack
### Feature Overview

Currently, the PEI Core instance of AdvancedLoggerLib relies on hacking the PEI_CORE_INSTANCE to have a stack based pointer to the advanced logger buffer. This avoids a HOB lookup on every log message. The lib instance uses PeiMain.h to have the definition of the structure.

However, mu_basecore has been updated to include an edk2 change that added a private header to PeiMain.h (which is private but not enforced so by the build system), which breaks the builds. In response, a temporary workaround  was added to force include this private header.

### Solution Overview

The hack of using PEI_CORE_INSTANCE can be dropped entirely and a separate scratchpad space can be used to avoid the HOB lookup.

### Alternatives Considered

_No response_

### Urgency

Medium

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

os-d:
## Description

Currently, The SecDebugAgent and PeiCore instances of advanced logger use the internal PEI_CORE_INSTANCE to store a pointer to the advanced logger buffer. This is a performance optimization to avoid a HOB lookup on every debug message.

However, this is also a hack and requires a basecore override to reach into the core internals.

This commit changes that behavior by introducing a PCD to define a scratchpad page where the pointer can be stored. This must exist outside of the PHIT free memory region so that it is not handed out as another allocation in PEI, because the allocator does not respect memory allocation HOBs for that region.

If a platform does not specify this PCD, a HOB lookup will be performed on every debug message.

Closes #910 .

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booting to shell on Q35, seeing PEI logs. Added a spin loop when the PCD is used to confirm it wasn't falling back to the HOB lookup. Physical Intel testing is going on in parallel with this PR.

## Integration Instructions

This is marked as a breaking change because there is a significant performance impact if a platform does not take action here. However, the build and run will still work if not.

In order to keep the same performance guarantees as the current code, platforms must provide a page address that is outside of the PHIT free memory region in `gAdvLoggerPkgTokenSpaceGuid.PcdAdvancedLoggerScratchpadBase`. This must be available to write to in SEC and PEI_CORE.


cfernald (approved):

AdvLoggerPkg/Library/AdvancedLoggerLib/PeiCore/AdvancedLoggerLib.c
```diff
@@ -482,14 +569,21 @@ AdvancedLoggerGetLoggerInfo (
...
+      if (LoggerInfo != NULL) {
+        SetCachedLoggerInfo (LoggerInfo);
+        LogPtr->LogBuffer = PA_FROM_PTR (LoggerInfo);
+
+        LoggerInfo->LogCurrentOffset = EXPECTED_LOG_BUFFER_OFFSET (LoggerInfo) + USED_LOG_SIZE (LoggerInfo);
+        LoggerInfo->LogBufferOffset  = EXPECTED_LOG_BUFFER_OFFSET (LoggerInfo);
+      }
 
       // return the pointer
       return LoggerInfo;
```
vineelko: Should the `return LoggerInfo;` be inside the above non-null check?

os-d: Actually, I think it should be outside of !LoggerInfo->InPermanentRAM. I think if we find the HOB, we should always be returning.

os-d: Updated and offline @vineelko tested and it fixed an issue he had encountered in a certain configuration.

apop5 (approved):

PR #919, opened 2026-07-14
MsGraphicsPkg: use GOP from ConsoleOut

cmruffin:
## Description
When multiple GOP protocol instances are installed in the system, different components of MsGraphicsPkg can connect to different GOP instances leading to inconsistent graphics rendering or failures. Always use the one that is connected to ConsoleOut.

## How This Was Tested
Boot test with multiple GOP configuration.

## Integration Instructions
N/A

### DIFF
--- a/MsGraphicsPkg/Library/SimpleUIToolKit/SimpleUIToolKit.c
+++ b/MsGraphicsPkg/Library/SimpleUIToolKit/SimpleUIToolKit.c
@@ -42,9 +42,9 @@ InitializeUIToolKit (
 
   // Determine if the GOP is available.
   //
-  Status = gBS->LocateProtocol (
+  Status = gBS->HandleProtocol (
+                  gST->ConsoleOutHandle,
                   &gEfiGraphicsOutputProtocolGuid,
-                  NULL,
                   (VOID **)&mUITGop
                   );
 
--- a/MsGraphicsPkg/SimpleWindowManagerDxe/WindowManager.c
+++ b/MsGraphicsPkg/SimpleWindowManagerDxe/WindowManager.c
@@ -1507,9 +1507,9 @@ GopRegisteredCallback (
 
   // Determine if the Graphics Output Protocol is available on the Console Out handle.
   //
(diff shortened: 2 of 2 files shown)


cmruffin:
@spbrogan can you please help review it?

cmruffin:
@apop5 @makubacki can you please remove the stale tag?

MsGraphicsPkg/Library/SimpleUIToolKit/SimpleUIToolKit.c
```diff
@@ -42,9 +42,9 @@ InitializeUIToolKit (
 
   // Determine if the GOP is available.
   //
-  Status = gBS->LocateProtocol (
+  Status = gBS->HandleProtocol (
+                  gST->ConsoleOutHandle,
                   &gEfiGraphicsOutputProtocolGuid,
-                  NULL,
```
kuqin12, line 47: Do you only need these 2 instances changed? How about OSK and SwmDialog?

MsGraphicsPkg/Library/SimpleUIToolKit/SimpleUIToolKit.c
kuqin12, line 55: Do you need a fallback if failed to find GOP on con out?

cmruffin: In the case of WindowManagerDxe.c, a fallback will cause an issue because it can fully initialize on any GOP instead of only the one that is connected to ConOut. In SimpleUIToolkit.c, I think it is not an issue but also not necessary. In other code locations there is a fallback, but I think these mask problems that are best identified and resolved.

MsGraphicsPkg/Library/SimpleUIToolKit/SimpleUIToolKit.c
cmruffin: That may be a further improvement but I am not currently able to re-test in a multiple GOP environment with that change so I would prefer to leave it for a future activity.

PR #923, opened 2026-07-23
AdvLoggerPkg/AdvancedFileLogger: Flush logs only once per reset

liqiqiii:
## Description

A capsule-armed reset converts a warm reset into a cold reset by issuing a second, nested `ResetSystem()` call. `NvmExpressDxe` shuts its controller down on the first reset notification (via the reset notification protocol, which fires after this platform-specific reset filter) and does not re-initialize it, so re-writing the logs on the nested reset issues I/O to a powered-down NVMe controller. This stalls for the full NVMe command timeout (~5 seconds) and logs a fatal NVMe WHEA telemetry record `(EFI_IO_BUS_SCSI | EFI_IOB_EC_INTERFACE_ERROR, 0x02070005)`, and the log write ultimately fails with a timeout.

Add a module-static latch so the in-memory log is flushed to media only once per reset sequence. Nested reset notifications early-return instead of re-writing to the already-shut-down controller. The normal ReadyToBoot / ExitBootServices / sync flush paths are unaffected.

For details on how to complete these options and their meaning refer to [CONTRIBUTING.md](https://github.com/microsoft/mu/blob/HEAD/CONTRIBUTING.md).

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with hardware, work as expected.

### DIFF
--- a/AdvLoggerPkg/AdvancedFileLogger/AdvancedFileLogger.c
+++ b/AdvLoggerPkg/AdvancedFileLogger/AdvancedFileLogger.c
@@ -19,6 +19,14 @@ UINT32      mWritingSemaphore        = 0;
 EFI_EVENT  mReadyToBootEvent      = NULL;
 EFI_EVENT  mExitBootServicesEvent = NULL;
 
+//
+// Set once the in-memory log has been flushed to media during a reset
+// notification. Prevents re-flushing on a nested reset (for example, a
+// capsule-armed warm-to-cold reset conversion) after storage controllers have
+// already been shut down.
+//
+STATIC BOOLEAN  mLogsFlushedOnReset = FALSE;
+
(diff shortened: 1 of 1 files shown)


apop5:
Has the unregister path been tested? What if, after the log was written, the notification function was unregistered?

liqiqiii:
> Has the unregister path been tested? What if, after the log was written, the notification function was unregistered?

Good catch Aaron, I will get some Intel secured core/non secured core devices tested as well. The secured core could be highly influence by this.

apop5:
@liqiqiii 

Is this still needed?

PR #938, opened 2026-09-11
UefiTestingPkg: Correctly handle cached DMA audit ACPI tables

### LINKED ISSUE #937: [Bug]: DMAVTdProtectionUnitTestApp Failure

### Current Behavior

After commit https://github.com/microsoft/mu_plus/commit/484e502d12f0a86b7a479aab61b6d9af71feafac, the DMAVTdProtectionUnitTestApp fails on Intel platforms with:

```
Test: Verify excluded ranges are marked reserved in Run All shell based unit and functional tests
File: DMAProtectionTestArch.c:55
Error: Status is EFI_ERROR (Already started)
```

Tests call `GetDmarAcpiTable()`, but the Intel VT-d implementation caches the DMAR table and returns `EFI_ALREADY_STARTED` on the second call without a reboot.

### Expected Behavior

All tests including "Verify excluded ranges are marked reserved" pass successfully.

### Steps To Reproduce

Run DMAVTdProtectionUnitTestApp on an Intel platform with the commit included in the app build.

### Build Environment

```markdown
N/A
```

### Version Information

```text
As of v2025110003.0.3
```

### Urgency

Low

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

eeshanl:
## Description

Correctly handle cached DMA audit ACPI tables. Fixes https://github.com/microsoft/mu_plus/issues/937

Tests call GetDmarAcpiTable(), but the Intel VT-d and IVRS implementation caches the ACPI table and returns EFI_ALREADY_STARTED on the second call without a reboot.

GetAcpiTable() will no longer return already started and will always refresh the pointer with the data. On the caller side, the caching of the global variable is handled so we don't call GetAcpiTable() redundantly. 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Intel platform and all tests now pass.

## Integration Instructions

N/A


UefiTestingPkg/AuditTests/DMAProtectionAudit/UEFI/Acpi.c
```diff
@@ -142,8 +142,9 @@ FindAcpiPtr (
 /**
   Get the ACPI table.
 
-  @retval EFI_SUCCESS           The ACPI table is got.
-  @retval EFI_NOT_FOUND         The ACPI table is not found.
+  @retval EFI_SUCCESS            The ACPI table is got.
```
makubacki: :) Do you mind changing this while here?

```suggestion
  @retval EFI_SUCCESS            The ACPI table was found.
```

UefiTestingPkg/AuditTests/DMAProtectionAudit/UEFI/Acpi.c
```diff
@@ -157,7 +158,7 @@ GetAcpiTable (
   if (AcpiTable == NULL) {
     return EFI_INVALID_PARAMETER;
   } else if (*AcpiTable != NULL) {
-    return EFI_ALREADY_STARTED;
+    return EFI_SUCCESS;
```
apop5: Can you add a comment to why  being passed a valid AcpiTable is considered a success? it seems counter insulative that an OUT parameter is being checked and short-circuiting the execution

eeshanl: Sure, I can add a comment for that and change to IN OUT

Another possibility is that we can keep the EFI_ALREADY_STARTED, and handle it on the caller side to not assert and fail.
What would you prefer?

makubacki: I think this is overly complicated.

Imo, the `GetAcpiTable()` function should only do exactly that and return an ACPI table as an output. The intermediate callers to `GetAcpiTable()` like `GetDmarAcpiTable()` should have knowledge of their global storage variable and not request a table if it's already located.

eeshanl: Updated callers of GetAcpiTable() to check if the global is not NULL and just return success, else return GetAcpiTable().

makubacki: Part of the complexity that remains is `GetAcpiTable()` doesn't really need to have logic and error codes for the incoming state of a non-null `*AcpiTable` pointer. If someone called the function and the outer pointer (`AcpiTable`) is non-null, then they're asking for it to be refreshed to the latest tables. The caller should have determined if that call was the right one to make by deciding beforehand if they need the pointer updated.

As of your most recent push, the original issue is mitigated by the caller checks, but this is overall still a little extra complexity.

makubacki: Basically, I think you can just drop this now:

```c
  else if (*AcpiTable != NULL) {
    return EFI_SUCCESS;
  }
```

And ideally return instead of just assert if `AcpiConfigurationTable` is `NULL` later in the function.

eeshanl: Done

UefiTestingPkg/AuditTests/DMAProtectionAudit/UEFI/Acpi.c
```diff
@@ -172,12 +171,12 @@ GetAcpiTable (
                );
   }
 
-  if (EFI_ERROR (Status)) {
+  ASSERT (AcpiConfigurationTable != NULL);
```
apop5:
```suggestion
  if(AcpiConfigurationTable ==NULL) {
  DEBUG((DEBUG_ERROR, "System Table Acpi10 was invalid!!!\n"));
  return EFI_NOT_FOUND:
```


Asserts are not useful like this. They can be compiled out. If you are going to use an assert, it means execution needs to stop, so wrap it in a return statement.

Asserts in functional system test code are also not recommended. We run these test on our virtual platforms, and those platforms, if they hit an assert, will not continue to the next unit test until the pipeline times out.

makubacki (approved):
Overall, looks fine now with the assert open remaining. Please wait for Intel system verification before merging.

UefiTestingPkg/AuditTests/DMAProtectionAudit/UEFI/Acpi.c
eeshanl: Removed the ASSERT.

apop5 (approved):

PR #939, opened 2026-09-13
GopOverrideDxe: Add driver binding protocol

cmruffin:
## Description

Adds UEFI driver model to GopOverrideDxe so that the GOP can be disconnected and a new GOP reconnected in BDS.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit tests with disconnect/connect sequence in BDS.

## Integration Instructions

N/A


MsGraphicsPkg/GopOverrideDxe/GopOverrideDxe.c
```diff
@@ -11,31 +16,210 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
+                  );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "ERROR [GOP]: Unable to open binding marker - code=%r\n", Status));
+    goto UninstallMarker;
+  }
 
+  //
+  // Uninstall the original GraphicsOutputProtocol on this handle.
+  //
+  Status = gBS->UninstallMultipleProtocolInterfaces (
```
apop5, line 168: I am not sure that this will work.

In the past, I've experienced Gop Protocols that consume the passed gop instance to do a CR to get back to a private data structure. 
Uninstalling the existing GOP will break this behavior.

cmruffin: The interface is still there and published on the same handle with mMsGopOverrideProtocolGuid which is consumed by RenderingEngineDxe. RenderingEngineDxe will call the original GOP using its original interface pointer so structure offset referemces should still work. Note also this is incumbent behavior so we have some empirical evidence that it works ok.

